PR #14, opened 2021-06-10
Require master token

webrouse:
Zistil som, ze je potrebny `master` token, tak to musime validovat:
https://keboola.slack.com/archives/CQBEHAXDL/p1623341702074000


ondrajodas (approved):
👍

PR #72, opened 2021-07-22
Cleanup: naming

webrouse:
All `naming` methods moved to `naming.go` in `model` package (because model structs are inputs for naming methods).
This change prevents circular dependencies problems in next PRs.


kacurez (approved):
lgtm

PR #73, opened 2021-07-22
Cleanup: Split `model.go` to multiple files.

webrouse:


src/model/base.go
```diff
@@ -1,9 +1,7 @@
 package model
 
 import (
-	"fmt"
```
webrouse, line 4: Subor `model.go` som rozdelil na `keys.go`, `base.go`, `rename.go` ... kedze tam bolo toho uz vela a bolo to neprehladne.

kacurez (approved):
lgtm

PR #124, opened 2021-09-21
Embedded model.PathsState to model.State

webrouse:
Part of: https://keboola.atlassian.net/browse/COM-1104

### DIFF
--- a/internal/pkg/state/state.go
+++ b/internal/pkg/state/state.go
@@ -20,7 +20,6 @@ import (
 type State struct {
 	*Options
 	*model.State
-	*model.PathsState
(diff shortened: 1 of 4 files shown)


internal/pkg/state/state.go
```diff
@@ -20,7 +20,6 @@ import (
 type State struct {
 	*Options
 	*model.State
-	*model.PathsState
```
webrouse, line 23: Povodne bola struktura `PathsState` sucastou `state.State`, ... ale teraz je sucastou `model.State`, tj. o uroven nizsie.

kacurez (approved):
lgtm

PR #160, opened 2021-10-04
Fix keys embedding and interfaces

webrouse:
Part of: https://keboola.atlassian.net/browse/COM-1059

### DIFF
--- a/internal/pkg/model/keys.go
+++ b/internal/pkg/model/keys.go
@@ -16,9 +16,11 @@ const (
 )
 
 type Key interface {
+	Level() int     // hierarchical level, "1" for branch, "2" for config, ...
 	Kind() Kind     // kind of the object: branch, config, ...
 	Desc() string   // human-readable description of the object
 	String() string // unique string representation of the key
+	ObjectId() string
 }
 
 type WithKey interface {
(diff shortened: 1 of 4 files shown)


internal/pkg/model/keys.go
```diff
@@ -16,9 +16,11 @@ const (
 )
 
 type Key interface {
+	Level() int     // hierarchical level, "1" for branch, "2" for config, ...
```
webrouse, line 19: Fixol som `interface`, obsahovali vela duplicitnych metod, preto som pouzil `embedding`.

Hierarchia je takato: `Key -> Object -> ObjectState` (vzdy sa pridavaju dalsie metody).

internal/pkg/model/object.go
```diff
@@ -21,11 +21,8 @@ const (
 )
 
 type Object interface {
-	Level() int // hierarchical level, "1" for branch, "2" for config, ...
-	Kind() Kind
+	Key
```
webrouse, line 24: Interface `Key` je vnorene do interface `Object` ... tj. `Object` ma tie metody co `Key` + dalsie.

kacurez (approved):
lgtm

PR #301, opened 2021-11-08
Variables: rename id -> configId

webrouse:
Part of: https://keboola.atlassian.net/browse/COM-1137

### DIFF
--- a/internal/pkg/manifest/manifest_test.go
+++ b/internal/pkg/manifest/manifest_test.go
@@ -377,7 +377,7 @@ func fullJson() string {
       "relations": [
         {
           "componentId": "keboola.wr-db-mysql",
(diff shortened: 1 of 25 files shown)


internal/pkg/manifest/manifest_test.go
```diff
@@ -377,7 +377,7 @@ func fullJson() string {
       "relations": [
         {
           "componentId": "keboola.wr-db-mysql",
-          "id": "12",
+          "configId": "12",
```
webrouse, line 380: Drobnost: premenoval som `id` -> `configId` v relation, v `manifest.json` ... aby bolo jasne, ze je to ID cieloveho configu, a nie nejake ID relacie.

PR #321, opened 2021-11-12
Add info about diff --details flag

webrouse:


test/diff/expected-stdout
```diff
@@ -6,3 +6,5 @@ Diff:
 * B bar | changed: description
 * C bar/extractor/ex-generic-v2/empty | changed: name
 - C main/extractor/ex-generic-v2/empty
+
+Use --details flag to list the changed fields.
```
webrouse, line 10: Info ako vypisat podrobnosti pri `diff`.

JakubMatejka (approved):

PR #335, opened 2021-11-17
Orchestrator: use pointers for Phase/Task

webrouse:
Part of: https://keboola.atlassian.net/browse/COM-1215

### DIFF
--- a/internal/pkg/model/orchestrator.go
+++ b/internal/pkg/model/orchestrator.go
@@ -5,14 +5,14 @@ import (
 )
 
 type Orchestration struct {
-	Phases []Phase
+	Phases []*Phase
 }
 
 type Phase struct {
 	PhaseKey
 	PathInProject
 	DependsOn []PhaseKey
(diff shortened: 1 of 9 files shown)


internal/pkg/model/orchestrator.go
```diff
@@ -5,14 +5,14 @@ import (
 )
 
 type Orchestration struct {
-	Phases []Phase
+	Phases []*Phase
```
webrouse, line 8: Pre `Phase/Task` budeme pouzivat pointer.
Kedze bez toho sa s tym blbo pracuje, napr. pri premenovani `Phase` (dalsie PRs).

Podobne je to aj v transformation `Block/Code`:
https://github.com/keboola/keboola-as-code/blob/9a2ef792834b6baa520c699dcfafd00198432609/internal/pkg/model/object.go#L160-L173

JakubMatejka (approved):

PR #356, opened 2021-11-29
Add FirstUpper method

webrouse:
Part of: https://keboola.atlassian.net/browse/COM-1232


internal/pkg/strhelper/strhelper.go
```diff
@@ -68,6 +68,10 @@ func FirstLower(str string) string {
 	return strings.ToLower(string(str[0])) + str[1:]
 }
 
+func FirstUpper(str string) string {
```
webrouse, line 71: Nova metoda.

JakubMatejka (approved):

PR #439, opened 2022-01-17
Cleanup the hierarchy of packages in pkg/lib/operation

webrouse:
V `pkg/lib/operation` sa nam miesali operacie nad projektom a template, tak som to fixol.

Pred:
image

Po:
image


JakubMatejka (approved):
👍

PR #463, opened 2022-01-20
Use naming with IDs instead of names, in template

webrouse:
Part of: https://keboola.atlassian.net/browse/COM-1275

### DIFF
--- a/internal/pkg/naming/template.go
+++ b/internal/pkg/naming/template.go
@@ -63,9 +63,15 @@ func (p PathTemplate) placeholderToRegexp(placeholder string) string {
 }
 
 func ForTemplate() Template {
-	t := TemplateWithoutIds()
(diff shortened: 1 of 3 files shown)


internal/pkg/naming/template.go
```diff
@@ -63,9 +63,15 @@ func (p PathTemplate) placeholderToRegexp(placeholder string) string {
 }
 
 func ForTemplate() Template {
-	t := TemplateWithoutIds()
-	t.Branch = ``
-	return t
+	return Template{
+		Config:              "{component_type}/{component_id}/{config_id}",
```
webrouse, line 67: Template pouziva human readable IDs, tak ich pouzijeme aj v nazvoch adresarov.

JakubMatejka (approved):

PR #586, opened 2022-03-03
Templates: mark configs from template metadata

webrouse:
Part of: https://keboola.atlassian.net/browse/KAC-5

### DIFF
--- a/pkg/lib/operation/project/local/template/use/operation.go
+++ b/pkg/lib/operation/project/local/template/use/operation.go
@@ -5,6 +5,8 @@ import (
 	"fmt"
 	"sort"
 
+	gonanoid "github.com/matoous/go-nanoid/v2"
+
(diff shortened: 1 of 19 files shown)


pkg/lib/operation/project/local/template/use/operation.go
```diff
@@ -79,10 +81,13 @@ func Run(tmpl *template.Template, o Options, d dependencies) error {
 	// Create tickets provider, to generate new IDS
 	tickets := storageApi.NewTicketProvider()
 
+	// Generate ID for the template instance
+	instanceId := gonanoid.Must()
```
webrouse, line 85: Pouzil som `nanoid` ako instance ID, je kratsie ako `UUID` a nemusi sa v URL escapovat:
https://github.com/matoous/go-nanoid

JakubMatejka (approved):

PR #645, opened 2022-04-26
HTTP pool: request can be send only once

webrouse:
Related to: https://github.com/keboola/keboola-as-code/pull/642#discussion_r858336134


JakubMatejka (approved):
super 👍

PR #654, opened 2022-04-28
Use template, add instance name to the payload def

webrouse:
Slack: https://keboola.slack.com/archives/C02JAQXLY92/p1651064424172229

### DIFF
--- a/api/templates/design.go
+++ b/api/templates/design.go
@@ -242,6 +242,7 @@ var _ = Service("templates", func() {
 			templateAttr()
 			versionAttr()
 			branchAttr()
+			instanceNameAttr()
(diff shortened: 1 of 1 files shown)


api/templates/design.go
```diff
@@ -242,6 +242,7 @@ var _ = Service("templates", func() {
 			templateAttr()
 			versionAttr()
 			branchAttr()
+			instanceNameAttr()
```
webrouse, line 245: Pridane `name` do payload v `/use` endpointe, malo to byt az v dalsej verzii, no nakoniec je to vo `v1`.

JakubMatejka (approved):

PR #713, opened 2022-06-09
Fix bug in inputs usage metadata

webrouse:
Jira: https://keboola.atlassian.net/browse/KAC-146


internal/pkg/template/use/context_test.go
```diff
@@ -112,6 +131,7 @@ func TestContext(t *testing.T) {
   "Input2": 789,
   "Input3": 3.5,
   "Input4": false,
+  "Input5": "",
```
webrouse, line 134: V `jsonnet` sa pouzije prazdna hodnota.

JakubMatejka (approved):

PR #852, opened 2022-09-30
Templates API: keep components order

webrouse:
Jira: https://keboola.atlassian.net/browse/KAC-253

### DIFF
--- a/internal/pkg/api/server/templates/service/mapper.go
+++ b/internal/pkg/api/server/templates/service/mapper.go
@@ -3,7 +3,6 @@ package service
 import (
 	"context"
 	"fmt"
-	"sort"
 	"time"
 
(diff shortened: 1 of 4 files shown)


internal/pkg/api/server/templates/service/mapper.go
```diff
@@ -183,7 +184,6 @@ func ComponentsResponse(d dependencies.ForProjectRequest, in []string) (out []st
 		}
 		out = append(out, componentId)
 	}
-	sort.Strings(out)
```
webrouse, line 186: Podla poradia `components` sa v UI vykresluju ikonky, takze to nemozeme zoradrovat.

JakubMatejka (approved):

PR #871, opened 2022-10-10
Remove unused code from utils

michaljurecko:
Part of: https://keboola.atlassian.net/browse/KAC-115


internal/pkg/utils/path.go
```diff
@@ -1,51 +0,0 @@
...
-
-import (
-	"fmt"
-	"regexp"
-	"strings"
-
-	"github.com/umisama/go-regexpcache"
-)
-
-type PathTemplate string
```
michaljurecko, line 11: Nikde sa nepouziva, nieco podobne je v `naming` pkg, asi sa to zabudlo zmazat.

jprochazk (approved):

PR #902, opened 2022-10-18
Hide keboola.sandboxes configurations from the kbc project folder structure

jprochazk:
https://keboola.atlassian.net/browse/KAC-240

### DIFF
--- a/internal/pkg/model/filter.go
+++ b/internal/pkg/model/filter.go
@@ -3,6 +3,7 @@ package model
 import (
 	"strings"
 
+	"github.com/keboola/go-client/pkg/sandboxesapi"
 	"github.com/keboola/go-client/pkg/storageapi"
 	"github.com/spf13/cast"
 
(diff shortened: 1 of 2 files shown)


internal/pkg/model/filter.go
```diff
@@ -28,6 +29,10 @@ type ObjectsFilter struct {
 	ignoredComponents ComponentIDs
 }
 
+var alwaysIgnoredComponents = map[string]bool{
```
jprochazk, line 32: Pokud to dám do `ignoredComponents` tak to bude vidět v manifestu. Stačí to takhle?

jprochazk:
Ten test failure vypadá unrelated

internal/pkg/model/filter.go
michaljurecko, line 32: Test je unrelated, ale musis fixnut linter, `// nolint: gochecknoglobals` (private global je ok, neda sa prepisat)

michaljurecko (approved):

PR #1117, opened 2023-01-10
store: Add clock field

michaljurecko:
Part of: https://keboola.atlassian.net/browse/BA-40

Changes:
- Added `clock` field to `Store` struct.

### DIFF
--- a/internal/pkg/service/buffer/store/store.go
+++ b/internal/pkg/service/buffer/store/store.go
@@ -3,6 +3,7 @@
 package store
 
 import (
+	"github.com/benbjohnson/clock"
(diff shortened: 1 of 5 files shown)


internal/pkg/service/buffer/store/store.go
```diff
@@ -11,17 +12,31 @@ import (
 )
 
 type Store struct {
+	clock  clock.Clock
```
michaljurecko, line 15: Store vznikol ako prvy a nepouziva `clock`, potrebujem to pre jednoduchsie otestovanie.

JakubMatejka (approved):

PR #1204, opened 2023-02-22
Include buffer/templates generated code in the git repo

michaljurecko:
Part of: https://keboola.atlassian.net/browse/PSGO-81

Changes:
- Commited generated code from Buffer/Templates services.
- It is better to have the generated code versioned and see the changes, e.g. when we update a library.


michaljurecko:
(mac-os CI ma nejaky problem s cache)

JakubMatejka (approved):

PR #1255, opened 2023-03-23
Single, exteral configuration for each service

michaljurecko:
Part of:
- https://keboola.atlassian.net/browse/PSGO-144
  - For a test.
- https://keboola.atlassian.net/browse/PSGO-108

Changes::
- Each service (`buffer-api`, `buffer-worker`, `templates-api`) has a new `config` pkg.
- It is single source of all configurable parameters.
- **Thanks to this, we will be able to test E2E use-cases that we did not until now.**

-----------------------------

image

image

image


JakubMatejka (approved):
👍

PR #1280, opened 2023-04-05
TA: Async templates use/upgrade

JakubMatejka:
(Nad https://github.com/keboola/keboola-as-code/pull/1279)


michaljurecko (approved):
:+1: super

PR #1300, opened 2023-04-18
BA: Fix benchmarks to work with MiniKube

michaljurecko:
Changes:
- Fixed K6 benchmarks to work with MiniKube deployment.

### DIFF
--- a/scripts/k6/buffer-api/common.js
+++ b/scripts/k6/buffer-api/common.js
@@ -1,4 +1,5 @@
 import http from "k6/http";
+import { URL } from 'https://jslib.k6.io/url/1.0.0/index.js';
 import {sleep, check} from 'k6';
 import { Counter } from 'k6/metrics';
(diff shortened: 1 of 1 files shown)


scripts/k6/buffer-api/common.js
```diff
@@ -71,7 +73,7 @@ export function setupReceiver(exports) {
         throw new Error("failed to get receiver");
     }
 
-    const receiverUrl = res.json().url
+    const receiverUrl = stripUrlHost(res.json().url)
```
michaljurecko, line 76: V MiniKube `deploy_local.sh` je nastavena public URL na `buffer.<host prefix>` (ako v produkcii), a host prefix je by default `keboola.com`.

Tj. URLs vygenerovane BA v MiniKube zacinaju ako `https://buffer.keboola.com`, ... v benchmarku musime teda host ignorovat.

JakubMatejka (approved):

PR #1362, opened 2023-06-23
fix(provisioning/etcd): add missing volumeMounts field

dakraus:
In case etcd is configured to encrypt client-to-server communication using TLS certificates (with `auth.client.secureTransport: true`), the rendering of the etcd-defragger CronJob fails. Therefore the missing `volumeMounts` field was added to the CronJob.


michaljurecko:
Thanks @dakraus for the PR!
I created a copy of the branch in our repo, to make CI working:
https://github.com/keboola/keboola-as-code/pull/1365

PR #1451, opened 2023-12-04
Use jsonEncoder for log file

jachym-tousek-keboola:
https://keboola.atlassian.net/browse/PSGO-298

Solution is slightly different than the original task but I do think it serves the intended purpose.

I decided to only change the format of the logfile for now. If someone wants to work with the logs, I think they should use this file and don't see a flag to modify stdout format as necessary. The logfile also has all logs in one place rather than having some in stdout and some in stderr which is more convenient for machine processing.

### DIFF
--- a/internal/pkg/log/cli_test.go
+++ b/internal/pkg/log/cli_test.go
@@ -6,6 +6,7 @@ import (
 	"path/filepath"
 	"testing"
 
+	"github.com/keboola/go-utils/pkg/wildcards"
 	"github.com/stretchr/testify/assert"
 
 	"github.com/keboola/keboola-as-code/internal/pkg/utils/ioutil"
@@ -37,10 +38,16 @@ func TestCliLogger_File(t *testing.T) {
 	assert.NoError(t, file.File().Close())
 
 	// Assert, all levels logged with the level prefix
-	expected := "DEBUG\tDebug msg\nINFO\tInfo msg\nWARN\tWarn msg\nERROR\tError msg\n"
+	expected := `
+{"level":"debug","time":"%s","message":"Debug msg"}
+{"level":"info","time":"%s","message":"Info msg"}
+{"level":"warn","time":"%s","message":"Warn msg"}
+{"level":"error","time":"%s","message":"Error msg"}
+`
+
 	content, err := os.ReadFile(filePath)
 	assert.NoError(t, err)
-	assert.Equal(t, expected, string(content))
+	wildcards.Assert(t, expected, string(content))
 }
 
 func TestCliLogger_VerboseFalse(t *testing.T) {
--- a/internal/pkg/log/tofile.go
+++ b/internal/pkg/log/tofile.go
@@ -2,22 +2,22 @@
 package log
 
 import (
-	"go.uber.org/zap"
 	"go.uber.org/zap/zapcore"
 )
 
 // fileCore writes to a logFile.
 func fileCore(logFile *File) zapcore.Core {
 	// Log all
-	fileLevels := zap.LevelEnablerFunc(func(l zapcore.Level) bool { return true })
+	fileLevels := zapcore.DebugLevel
 
 	// Log time, level, msg
-	encoder := zapcore.NewConsoleEncoder(zapcore.EncoderConfig{
-		TimeKey:          "ts",
-		LevelKey:         "level",
-		MessageKey:       "msg",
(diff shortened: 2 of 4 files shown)


internal/pkg/log/cli_test.go
```diff
@@ -37,10 +38,15 @@ func TestCliLogger_File(t *testing.T) {
...
+	expected := ""
+	expected += `{"level":"debug","time":"%s","message":"Debug msg"}` + "\n"
+	expected += `{"level":"info","time":"%s","message":"Info msg"}` + "\n"
+	expected += `{"level":"warn","time":"%s","message":"Warn msg"}` + "\n"
+	expected += `{"level":"error","time":"%s","message":"Error msg"}` + "\n"
+
 	content, err := os.ReadFile(filePath)
 	assert.NoError(t, err)
-	assert.Equal(t, expected, string(content))
+	wildcards.Assert(t, expected, string(content))
```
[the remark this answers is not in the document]
michaljurecko: Yes, it is universal package.

internal/pkg/log/tofile.go
```diff
@@ -2,22 +2,22 @@
...
-		TimeKey:          "ts",
-		LevelKey:         "level",
-		MessageKey:       "msg",
-		EncodeLevel:      zapcore.CapitalLevelEncoder,
-		ConsoleSeparator: "\t",
+	encoder := zapcore.NewJSONEncoder(zapcore.EncoderConfig{
+		TimeKey:     "time",
+		LevelKey:    "level",
+		MessageKey:  "message",
+		EncodeLevel: zapcore.LowercaseLevelEncoder,
```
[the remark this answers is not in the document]
michaljurecko: Yes, it is OK.

`LowercaseLevelEncoder serializes a Level to a lowercase string. For example, InfoLevel is serialized to "info".`

We prefer `camelCaseJsonFieldNaming`.

internal/pkg/log/cli_test.go
```diff
@@ -37,10 +38,15 @@ func TestCliLogger_File(t *testing.T) {
 	assert.NoError(t, file.File().Close())
 
 	// Assert, all levels logged with the level prefix
-	expected := "DEBUG\tDebug msg\nINFO\tInfo msg\nWARN\tWarn msg\nERROR\tError msg\n"
+	expected := ""
+	expected += `{"level":"debug","time":"%s","message":"Debug msg"}` + "\n"
+	expected += `{"level":"info","time":"%s","message":"Info msg"}` + "\n"
+	expected += `{"level":"warn","time":"%s","message":"Warn msg"}` + "\n"
+	expected += `{"level":"error","time":"%s","message":"Error msg"}` + "\n"
+
```
michaljurecko, line 46: https://github.com/keboola/keboola-as-code/blob/5cfcf3c6e908b2c8910f77a501229145f43c5517/internal/pkg/service/common/etcdop/session_test.go#L54-L64

Could we keep this way of writing multi-line texts in tests ^^^? Or do you think it's a bad way?

The `wildcard` package automatically ignores leading and traling `\n`.

With the `assert.Equal` you can use `strings.TrimSpace`.

----------------

My point of view:

Pros:
- It is easy for editing.
- No escaping is needed.
- It provides easy refactoring with regexp, your format would complicate it (we have a lot of such assertions in the code).

Cons:
- It is at the beginning of the line, without indentation.

internal/pkg/log/tofile.go
```diff
@@ -2,22 +2,22 @@
...
 import (
-	"go.uber.org/zap"
 	"go.uber.org/zap/zapcore"
 )
 
 // fileCore writes to a logFile.
 func fileCore(logFile *File) zapcore.Core {
 	// Log all
-	fileLevels := zap.LevelEnablerFunc(func(l zapcore.Level) bool { return true })
+	fileLevels := zapcore.DebugLevel
```
michaljurecko, line 11: :+1:

michaljurecko (commented):
I have only one suggestion, the rest is LGTM

internal/pkg/log/cli_test.go
jachym-tousek-keboola: I don't mind using this though I do wish Go had something like PHP's newdoc syntax which can be indented.

It's changed accordingly now.

michaljurecko (approved):
LGTM 

---------------

Please. after making changes, resubmit it for review:
image

PR #1452, opened 2023-12-05
Add --log-format option

jachym-tousek-keboola:


internal/pkg/log/logformat.go
```diff
@@ -0,0 +1,29 @@
...
+
+import (
+	"github.com/keboola/keboola-as-code/internal/pkg/utils/errors"
+)
+
+type LogFormat int
+
+const (
+	Console LogFormat = iota
+	JSON
```
jachym-tousek-keboola, line 11: Should I prefix the names to be LogFormatConsole and LogFormatJSON instead? Console and JSON feel like they might easily conflict with something in the future.

internal/pkg/log/logformat.go
```diff
@@ -0,0 +1,29 @@
...
+// Creates LogFormat from string.
+// On invalid value Console is used as default with an error.
+func NewLogFormat(format string) (LogFormat, error) {
+	switch format {
+	case "console":
+		return Console, nil
+	case "json":
+		return JSON, nil
+	default:
+		return Console, errors.New(`log format must be "console" or "json"`)
```
jachym-tousek-keboola, line 27: Not entirely sure how to handle error so I took inspiration from how you handle log file and tried to do it similarly to be consistent. But if you have a better idea please share. :)

michaljurecko: This line is ok, but it will probably not be needed with string type.

internal/pkg/service/cli/options/options.go
```diff
@@ -43,6 +44,7 @@ func New() *Options {
 		envNaming: envNaming,
 		setBy:     make(map[string]cliconfig.SetBy),
 		parser:    viper.New(),
+		LogFormat: "console",
```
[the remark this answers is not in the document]
michaljurecko: Valid default values are always good practice.

Did it fall in NewLogFormat? (there will be no problem with the string type).

----

It the switch bellow, you preffered default encoder over error. It is IMHO ok, for style flags like log format.
So now there is no place for it to fall?

image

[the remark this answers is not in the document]
jachym-tousek-keboola: > Did it fall in NewLogFormat? (there will be no problem with the string type).

Yes it did because without the default set here it got empty string as input. I don't see why this would change with string type. Inputs other than "json" and "console" are still invalid.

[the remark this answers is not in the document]
jachym-tousek-keboola: > So now there is no place for it to fall?

Yeah the default branch is technically unnecessary. Should I remove it? I'm unsure what the best practice is around enums and switches in Go. In PHP we have static analysis that checks that every such switch uses all the possible values of the enum and has no default branch but I don't think such advanced rule exists for Go linter.

internal/pkg/log/logformat.go
```diff
@@ -0,0 +1,25 @@
...
+// On invalid value Console is used as default with an error.
+func NewLogFormat(format string) (LogFormat, error) {
+	switch format {
+	case "console":
+		return LogFormatConsole, nil
+	case "json":
+		return LogFormatJSON, nil
+	default:
+		return LogFormatConsole, errors.New(`log format must be "console" or "json"`)
+	}
```
jachym-tousek-keboola, line 24: Is there some way to simplify this? I'm unsure if there is a way to use `LogFormat(format)` - that syntax doesn't seem to return an "ok" value to detect invalid input value.

internal/pkg/service/cli/options/options.go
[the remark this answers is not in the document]
michaljurecko: > but I don't think such advanced rule exists for Go linter.
No, there is no such linter.

Default branch should always be there, you have 3 options:
- panic(errors.New("this is a programmer error"))
- return errors.New("this error can also occur during normal operation (for example user input)")
- default behaviour - I don't use this almost at all, but with log-format it makes sense

internal/pkg/log/logformat.go
michaljurecko: You can remove `NewLogFormat` method at all.

Right place for validation is here:
https://github.com/keboola/keboola-as-code/blob/ea0fa61c4e81094bf448240bffb50a8bb3116684/internal/pkg/log/encoder.go#L42-L44

There is a `panic` call.
- Panic is OK for programmer errors (if user cannot affect the value), as I wrote sooner, but here it is an user input.
- So you should return the error (and also at all higher levels).
- Is there any problem to do it this way? If so, I'll be happy to help.

michaljurecko (commented):
There is the last issue:

internal/pkg/log/logformat.go
jachym-tousek-keboola: I think we have a misunderstanding. The string passed into NewLogFormat is a user input. User can use `--log-format=unknown`. That's what the validation in NewLogFormat is for.

michaljurecko:
```
type Config struct {
  LogFormat LogFormat `validate:"one_of=console json"`
}
```

michaljurecko: `LogFormat.Validate() error`

```
func (v LogFormat) Validate() error {
  switch v {
    case Console, Json: return nil
    default return errors.New()
  }
}

func NewLogFormat(s string) error {
  switch LogFormat(s) {
    case Console, Json: return nil
    default return errors.New()
  }
}

LogFormat("foo")
```

michaljurecko (approved):
LGTM

PR #1453, opened 2023-12-05
Fix docker usage documentation

jachym-tousek-keboola:


docs/development.md
```diff
@@ -25,6 +25,12 @@ Staging storage can be `s3`, `abs` or `gcs`, according to the stack.
 
 Start an interactive console in a container, run:
 ```
+docker compose run --rm --service-ports dev bash
+```
+
+With older docker you may need to add `-u "$UID:$GID"` while using `docker compose run` for correct file permissions.
+However, with latest Docker Desktop it is no longer needed and will actually cause problems.
```
jachym-tousek-keboola, line 32: Should we add some link for more details? The [article](https://medium.com/nttlabs/cgroup-v2-596d035be4d7) you sent is a bit too technical to be used here imo.

michaljurecko (approved):
LGTM

PR #1454, opened 2023-12-06
Add context methods to logger

jachym-tousek-keboola:


build/ci/golangci.yml
```diff
@@ -51,6 +51,8 @@ linters-settings:
       - ^fmt\.Errorf$(# Use errors\.Errorf instead to include stack trace)?
       # No "gonanoid" package, use "internal/pkg/idgenerator" package instead
       - ^gonanoid\..*$(# Use "internal/pkg/idgenerator" package instead)?
+      # No logger calls without context
+      #- ^logger\.(Log|Debug|Info|Warn|Error|Debugf|Infof|Warnf|Errorf)$(# Use a method with Ctx suffix instead)?
```
jachym-tousek-keboola, line 55: Is there some better way to disallow the old methods? This depends on the variable name `logger` and doesn't catch cases such as `root.logger.Infof`. It should not matter how the logger is obtained.

michaljurecko: https://stackoverflow.com/questions/59627756/is-there-a-way-to-block-the-usage-of-a-specific-function-from-a-package-when-lin

It may be ok to disable all Info/Info calls.
These method names we probably only use within the logger.

internal/pkg/service/cli/cmd/cmd.go
```diff
@@ -302,36 +302,36 @@ func (root *RootCommand) printError(errRaw error) {
...
 			modifiedErrs.Append(errors.Wrapf(err, `none of this and parent directories is repository dir`))
 		case errors.Is(err, dependencies.ErrTemplateManifestNotFound):
-			root.logger.Infof(`Template directory must contain the "%s" file.`, templateManifest.Path())
-			root.logger.Infof(`You are in the template repository, but not in the template directory.`)
-			root.logger.Infof(`Please change working directory to a template directory, for example "template/v1".`)
-			root.logger.Infof(`Or use the "template create" command.`)
+			root.logger.InfofCtx(root.Context(), `Template directory must contain the "%s" file.`, templateManifest.Path())
+			root.logger.InfofCtx(root.Context(), `You are in the template repository, but not in the template directory.`)
+			root.logger.InfofCtx(root.Context(), `Please change working directory to a template directory, for example "template/v1".`)
+			root.logger.InfofCtx(root.Context(), `Or use the "template create" command.`)
```
jachym-tousek-keboola, line 334: Are you sure you want to always use the ctx methods? In cases like this logging the context values multiple times doesn't seem that great to me. But it's true that if we don't disallow the non-context methods completely it will be easy to forget to pass the context... I'm considering keeping the no context methods but renaming them:

```
logger.Infof(ctx, ...)
logger.InfofNoCtx(...)
```

jachym-tousek-keboola: Will consider it later. :)

internal/pkg/log/logger.go
```diff
@@ -52,6 +54,86 @@ func (l *zapLogger) Log(level string, args ...any) {
...
+	}
+	return fmt.Sprint(args...)
+}
+
+func (l *zapLogger) LogCtx(ctx context.Context, level string, args ...any) {
+	message := getMessage(args)
+	fields := make([]zap.Field, 0)
+	switch level {
+	case "debug", "DEBUG":
+		l.Desugar().Debug(message, fields...)
```
jachym-tousek-keboola, line 71: There is no way to pass both Sprintf args and fields into SugaredLogger so I'll need to use this lower level API.

internal/pkg/log/logger.go
```diff
@@ -52,6 +54,86 @@ func (l *zapLogger) Log(level string, args ...any) {
...
 }
 
+func getMessage(args []any) string {
+	if len(args) == 1 {
+		if str, ok := args[0].(string); ok {
+			return str
+		}
+	}
+	return fmt.Sprint(args...)
+}
```
jachym-tousek-keboola, line 64: Logic copied from zap since there is no method I could use directly.

internal/pkg/log/logger.go
```diff
@@ -52,6 +55,86 @@ func (l *zapLogger) Log(level string, args ...any) {
...
+
+func (l *zapLogger) LogCtx(ctx context.Context, level string, args ...any) {
+	message := getMessage(args)
```
michaljurecko, line 68: Could we please just add the unused ctx argument to the beginning of the method in this step?

We have not yet agreed on what the final methods will look like, see our discussion on slack.



image

PR #1463, opened 2023-12-11
feat: Add ctx to filesystem API

jachym-tousek-keboola:
https://keboola.atlassian.net/browse/PSGO-315


michaljurecko (approved):
:+1:

michaljurecko:
Side note: We should add a linter to check there are no unintendet uses of `context.Background()` or `context.TODO()` in all Go files, except tests.

I created task: https://keboola.atlassian.net/browse/PSGO-322

PR #1449, opened 2023-11-29
Finish configuration structure of the Stream API

michaljurecko:
Part of: https://keboola.atlassian.net/browse/PSGO-286

Changes:
- Finished configuration structure of the Stream API + related fixes/modifications.

--------------------
Default configuration looks like:
https://github.com/keboola/keboola-as-code/blob/115ab2d2bdb5e99169c85278fd5972fcc90f2a36/internal/pkg/service/buffer/config/config_test.go#L19-L140

PR #1464, opened 2023-12-11
Simplify servicectx package

michaljurecko:
Jira: https://keboola.atlassian.net/browse/PSGO-313
Related to: https://github.com/keboola/keboola-as-code/pull/1460

Changes:
- Removed `servicectx.Process.ctx` field and `Process.Ctx()` method.
  - It is [anti-pattern](https://go.dev/blog/context-and-structs).
- Simplified code and fixed edge-cases.

### DIFF
--- a/internal/pkg/service/common/httpserver/httpserver.go
+++ b/internal/pkg/service/common/httpserver/httpserver.go
@@ -48,19 +48,20 @@ func Start(ctx context.Context, d dependencies, cfg Config) error {
 	// Start HTTP server
 	srv := &http.Server{Addr: cfg.ListenAddress, Handler: handler, ReadHeaderTimeout: readHeaderTimeout}
 	proc := d.Process()
-	proc.Add(func(ctx context.Context, shutdown servicectx.ShutdownFn) {
+	proc.Add(func(shutdown servicectx.ShutdownFn) {
 		// Start HTTP server in a separate goroutine.
 		logger.InfofCtx(ctx, "HTTP server listening on %q", cfg.ListenAddress)
-		shutdown(ctx, srv.ListenAndServe())
+		serverErr := srv.ListenAndServe() // ListenAndServe blocks while the server is running
+		shutdown(context.Background(), serverErr)
 	})
 
 	// Register graceful shutdown
 	proc.OnShutdown(func(ctx context.Context) {
-		logger.InfofCtx(ctx, "shutting down HTTP server at %q", cfg.ListenAddress)
-
 		// Shutdown gracefully with a timeout.
 		ctx, cancel := context.WithTimeout(ctx, gracefulShutdownTimeout)
 		defer cancel()
+
+		logger.InfofCtx(ctx, "shutting down HTTP server at %q", cfg.ListenAddress)
 
 		if err := srv.Shutdown(ctx); err != nil {
 			logger.ErrorfCtx(ctx, `HTTP server shutdown error: %s`, err)
--- a/internal/pkg/service/common/servicectx/servicectx.go
+++ b/internal/pkg/service/common/servicectx/servicectx.go
@@ -111,10 +111,8 @@ func New(opts ...Option) (*Process, error) {
 		shutdown:    make(chan context.Context, 1),
 	}
 
-	// Execute OnShutdown callbacks and wait for them
-	v.wg.Add(1)
+	// Execute OnShutdown callbacks and then, after all work, unblock WaitForShutdown via done channel
 	go func() {
-		defer v.wg.Done()
 
 		// Wait for shutdown, see Shutdown function
 		shutdownCtx := <-v.shutdown
@@ -123,6 +121,15 @@ func New(opts ...Option) (*Process, error) {
 		for i := len(v.onShutdown) - 1; i >= 0; i-- {
 			v.onShutdown[i](shutdownCtx)
 		}
+
+		// Wait for all work
+		v.wg.Wait()
+
+		// Log message after successful termination
+		v.logger.InfoCtx(shutdownCtx, "exited")
+
+		// Unblock WaitForShutdown method calls
+		close(v.done)
 	}()
 
 	// Setup interrupt handler, so SIGINT and SIGTERM signals trigger shutdown.
@@ -140,19 +147,8 @@ func New(opts ...Option) (*Process, error) {
 		}
 	}()
 
-	// Finalizer
-	go func() {
-		// Wait for all work
-		v.wg.Wait()
-
-		// Log message after successful termination
-		v.logger.Info("exited")
-
-		// Unblock WaitForShutdown method calls
-		close(v.done)
-	}()
-
-	v.logger.Infof(`process unique id "%s"`, v.UniqueID())
+	// The unique id will be removed from the package.
+	v.logger.InfofCtx(context.TODO(), `process unique id "%s"`, v.UniqueID())
 	return v, nil
 }
 
@@ -185,7 +181,7 @@ func (v *Process) Shutdown(ctx context.Context, err error) {
 		return
 	default:
 		close(v.terminating)
-		v.logger.Infof("exiting (%v)", err)
+		v.logger.InfofCtx(ctx, "exiting (%v)", err)
 		v.shutdown <- ctx
 		close(v.shutdown)
 	}
--- a/internal/pkg/service/common/servicectx/servicectx.go
+++ b/internal/pkg/service/common/servicectx/servicectx.go
@@ -41,6 +41,7 @@ type Process struct {
 	shutdown chan context.Context
 	done     chan struct{}
 
+	// lock synchronizes Add, OnShutdown and Shutdown methods, so these methods are atomic.
 	lock        *sync.Mutex
 	terminating chan struct{}
 	onShutdown  []OnShutdownFn
@@ -134,7 +135,7 @@ func New(opts ...Option) (*Process, error) {
 			// Trigger shutdown on signal
 			v.Shutdown(context.Background(), errors.Errorf("%s", sig))
 		case <-v.terminating:
-			// The process was shut down by another trigger, stop goroutine
+			// The Process was shut down by another trigger, stop goroutine
 			return
 		}
 	}()
--- a/internal/pkg/service/common/servicectx/servicectx.go
+++ b/internal/pkg/service/common/servicectx/servicectx.go
@@ -14,24 +14,45 @@ import (
 	"github.com/keboola/keboola-as-code/internal/pkg/utils/errors"
 )
 
+const (
+	// ctxShutdownReasonKey stores the error that triggered shutdown.
+	ctxShutdownReasonKey = ctxKey("shutdownReason")
+)
+
+// Process is a stack of shutdown callbacks.
+// Callbacks are invoked sequentially, in LIFO order.
+// This makes it possible to register resource graceful shutdown callback when creating the resource.
+// It is simple approach for complex applications.
+//
+// Callbacks are registered via the OnShutdown method.
+//
+// Root goroutines are registered via the Add method.
+// Graceful shutdown waits for all root goroutines and shutdown callbacks.
+//
+// Graceful shutdown can be triggered by:
+//   - Signals SIGINT, SIGTERM
+//   - Call of the Shutdown method.
+//   - Call of the ShutdownFn function provided by the Add method.
 type Process struct {
 	uniqueID string
 
 	logger   log.Logger
 	wg       *sync.WaitGroup
-	errCh    chan error
+	shutdown chan context.Context
+	done     chan struct{}
 
-	lock          *sync.Mutex
-	terminating   bool
-	onShutdown    []OnShutdownFn
-	shutdownCtxCh chan context.Context
+	lock        *sync.Mutex
+	terminating chan struct{}
+	onShutdown  []OnShutdownFn
 }
 
-type Option func(c *config)
+type ctxKey string
 
-type OnShutdownFn func(context.Context)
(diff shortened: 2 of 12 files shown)


internal/pkg/service/common/servicectx/servicectx.go
```diff
@@ -79,50 +100,59 @@ func New(opts ...Option) (*Process, error) {
 		c.uniqueID = fmt.Sprintf(`%s-%05d`, hostname, pid)
 	}
 
-	// Create channel used by both the signal handler and service goroutines
-	// to notify the main goroutine when to stop the server.
-	errCh := make(chan error, 1)
-
-	// Create channel used to pass the shutdown context to OnShutdownFn callbacks.
-	shutdownCtxCh := make(chan context.Context, 1)
```
michaljurecko, line 88: These two channels have been replace by the `Process.shutdown` - `chan context.Context`.

See `ShutdownReason` function.

The context carries with him `reason` - shutdown trigger error.

BTW: error value was not used anywhere in the previous implementation, so `errCh` could be deleted away.

internal/pkg/service/common/servicectx/servicectx.go
```diff
@@ -79,50 +100,59 @@ func New(opts ...Option) (*Process, error) {
...
+		lock:        &sync.Mutex{},
+		terminating: make(chan struct{}),
+		shutdown:    make(chan context.Context, 1),
 	}
 
-	// Register onShutdown operation
-	proc.Add(func(ctx context.Context, _ ShutdownFn) {
-		<-ctx.Done()
-		shutdownCtx := <-proc.shutdownCtxCh
+	// Execute OnShutdown callbacks and wait for them
```
michaljurecko, line 113: The goroutine executes shutdown callbacks.

internal/pkg/service/common/servicectx/servicectx.go
```diff
@@ -79,50 +100,59 @@ func New(opts ...Option) (*Process, error) {
...
-			proc.onShutdown[i](shutdownCtx)
+		for i := len(v.onShutdown) - 1; i >= 0; i-- {
+			v.onShutdown[i](shutdownCtx)
 		}
-	})
+	}()
 
-	// Setup interrupt handler,
-	// so SIGINT and SIGTERM signals cause the services to stop gracefully.
+	// Setup interrupt handler, so SIGINT and SIGTERM signals trigger shutdown.
```
michaljurecko, line 127: The goroutine listen for signals an triggers shutdown.

internal/pkg/service/common/servicectx/servicectx.go
```diff
@@ -79,50 +100,59 @@ func New(opts ...Option) (*Process, error) {
...
+			v.Shutdown(context.Background(), errors.Errorf("%s", sig))
+		case <-v.terminating:
+			// The process was shut down by another trigger, stop goroutine
+			return
 		}
 	}()
 
-	proc.logger.Infof(`process unique id "%s"`, proc.UniqueID())
-	return proc, nil
+	// Finalizer
```
michaljurecko, line 142: The goroutine logs `exited` message and unblock `WaitForShutdown` method calls.

In the previous version, the code was part of the first `WaitForShutdown` call.
Here it is more clear, and `WaitForShutdown` method is simple.

internal/pkg/service/common/servicectx/servicectx.go
```diff
@@ -144,64 +174,66 @@ func NewForTest(t *testing.T, opts ...Option) *Process {
...
-	}()
+// Add a root operation.
+// The operation can terminate the entire process by calling the shutdown callback with an error.
+func (v *Process) Add(operation func(ShutdownFn)) {
+	v.lock.Lock()
+	defer v.lock.Unlock()
+
+	select {
+	case <-v.terminating:
+		v.logger.Errorf(`cannot Add operation: the Process is terminating`)
```
michaljurecko, line 206: Added missing check of the unexpected state.

michaljurecko: `Add` / `OnShutdown` methods should be called only at app startup.

internal/pkg/service/common/servicectx/servicectx.go
```diff
@@ -144,64 +174,66 @@ func NewForTest(t *testing.T, opts ...Option) *Process {
...
+	default:
+		v.onShutdown = append(v.onShutdown, fn)
 	}
-	v.onShutdown = append(v.onShutdown, fn)
-	v.lock.Unlock()
+}
+
+func (v *Process) WaitForShutdown() {
+	<-v.done
+}
```
michaljurecko, line 233: Simplified.

internal/pkg/service/common/servicectx/servicectx.go
```diff
@@ -144,64 +174,66 @@ func NewForTest(t *testing.T, opts ...Option) *Process {
...
+
+// ShutdownReason gets the shutdown reason error.
+func ShutdownReason(ctx context.Context) error {
+	v, _ := ctx.Value(ctxShutdownReasonKey).(error)
+	return v
 }
```
michaljurecko, line 239: Added.

internal/pkg/service/common/servicectx/servicectx.go
```diff
@@ -144,64 +174,66 @@ func NewForTest(t *testing.T, opts ...Option) *Process {
...
-	v.lock.Unlock()
+}
+
+func (v *Process) WaitForShutdown() {
+	<-v.done
+}
+
+// ShutdownReason gets the shutdown reason error.
+func ShutdownReason(ctx context.Context) error {
+	v, _ := ctx.Value(ctxShutdownReasonKey).(error)
```
michaljurecko, line 237: This is typical method to set/get context value.
- There is custom private type `ctxKey string`, so no other package can set same key :exclamation: 
- Context value has `any` type, but we know that it is `error`, so private key constant + public getter function is the best approach.

internal/pkg/service/common/servicectx/servicectx_test.go
```diff
@@ -15,71 +16,83 @@ import (
...
-		<-ctx.Done()
-		logger.InfoCtx(ctx, "end1")
-		op1.Done()
+	// OpCtx simulates long running operations
+	opCtx, opCancel := context.WithCancel(context.Background())
+
+	// There are some parallel operations
+	opWg := &sync.WaitGroup{}
+	opWg.Add(3)
+	proc.Add(func(shutdown ShutdownFn) {
```
michaljurecko, line 30: Test has been modified:
- `Process.Add` goroutines run in parallel, this test tests it (second test `TestProcess_Shutdown` is unmodified).

internal/pkg/service/common/servicectx/servicectx_test.go
```diff
@@ -15,71 +16,83 @@ import (
...
+
+	// Add more shutdown callbacks
 	proc.OnShutdown(func(ctx context.Context) {
-		logger.InfoCtx(ctx, "onShutdown1")
+		logger.InfoCtx(ctx, "onShutdown1") // STEP 7, LIFO
+
+		// Shutdown reason can be retrieved from the shutdown context
+		if err := ShutdownReason(ctx); assert.Error(t, err) {
+			assert.Equal(t, "operation failed", err.Error())
+		}
```
michaljurecko, line 60: Added.

internal/pkg/service/common/servicectx/servicectx_test.go
```diff
@@ -15,71 +16,83 @@ import (
...
+	// There are some parallel operations
+	opWg := &sync.WaitGroup{}
+	opWg.Add(3)
+	proc.Add(func(shutdown ShutdownFn) {
+		<-opCtx.Done()
+		logger.InfoCtx(opCtx, "end") // STEP 4, see <-opCtx.Done()
+		opWg.Done()
 	})
-	proc.Add(func(ctx context.Context, shutdown ShutdownFn) {
-		<-ctx.Done()
```
michaljurecko, line 37: The `Add` method originally provided the "root" context (which ends when the shutdown was requested).

But in practice we didn't use it like that:

For example, there are two sub-contexts, which must be closed in a specified order:
https://github.com/keboola/keboola-as-code/blob/7404898578f8b28f17f30a3bf08ac0cf46984b90/internal/pkg/service/common/distribution/node.go#L64-L75

internal/pkg/service/common/httpserver/httpserver.go
```diff
@@ -48,20 +48,21 @@ func Start(ctx context.Context, d dependencies, cfg Config) error {
 	// Start HTTP server
 	srv := &http.Server{Addr: cfg.ListenAddress, Handler: handler, ReadHeaderTimeout: readHeaderTimeout}
 	proc := d.Process()
-	proc.Add(func(ctx context.Context, shutdown servicectx.ShutdownFn) {
+	proc.Add(func(shutdown servicectx.ShutdownFn) {
 		// Start HTTP server in a separate goroutine.
 		logger.InfofCtx(ctx, "HTTP server listening on %q", cfg.ListenAddress)
-		shutdown(ctx, srv.ListenAndServe())
+		serverErr := srv.ListenAndServe() // ListenAndServe blocks while the server is running
+		shutdown(context.Background(), serverErr)
```
michaljurecko, line 55: :exclamation: :exclamation: :exclamation: shutdown starts here, so the shutdown context is created here :exclamation: :exclamation: :exclamation:

internal/pkg/service/common/servicectx/servicectx.go
```diff
@@ -14,26 +14,46 @@ import (
...
+// Callbacks are registered via the OnShutdown method.
+//
+// Root goroutines are registered via the Add method.
+// Graceful shutdown waits for all root goroutines and shutdown callbacks.
+//
+// Graceful shutdown can be triggered by:
+//   - Signals SIGINT, SIGTERM
+//   - Call of the Shutdown method.
+//   - Call of the ShutdownFn function provided by the Add method.
 type Process struct {
```
michaljurecko, line 36: There are many changes, I recommend looking at the resulting code.

internal/pkg/service/common/servicectx/servicectx.go
```diff
@@ -80,139 +100,140 @@ func New(ctx context.Context, cancel context.CancelFunc, opts ...Option) (*Proce
...
-	// Wait for signal
-	_, ok := <-v.errCh
-	if !ok {
-		// Channel is closed, the process is already terminating, wait for completion
-		v.wg.Wait()
+	select {
+	case <-v.terminating:
 		return
+	default:
+		close(v.terminating)
```
michaljurecko, line 186: https://github.com/keboola/keboola-as-code/pull/1464/files#r1423442477

michaljurecko:
TODO:
- Remove unrelated `uniqueID` field from the `servicectx.Process` struct.
  - It is prepared in Stream API RFC, see https://github.com/keboola/keboola-as-code/pull/1449/commits/fb1b5da6a8e534b00dabc27a58f7a9a48c0ee2a0

internal/pkg/service/common/servicectx/servicectx.go
```diff
@@ -144,64 +174,66 @@ func NewForTest(t *testing.T, opts ...Option) *Process {
...
-	v.wg.Add(1)
-	go func() {
-		defer v.wg.Done()
-		operation(v.ctx, v.Shutdown)
-	}()
+// Add a root operation.
+// The operation can terminate the entire process by calling the shutdown callback with an error.
+func (v *Process) Add(operation func(ShutdownFn)) {
+	v.lock.Lock()
+	defer v.lock.Unlock()
```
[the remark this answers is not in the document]
michaljurecko: I will add comment.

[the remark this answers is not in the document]
michaljurecko: Done @jachym-tousek-keboola https://github.com/keboola/keboola-as-code/pull/1464/commits/c19f58789d115c8f939d6f7e799734bc10869fc7

internal/pkg/service/common/servicectx/servicectx.go
```diff
@@ -80,139 +101,140 @@ func New(ctx context.Context, cancel context.CancelFunc, opts ...Option) (*Proce
...
 
-	proc.logger.Infof(`process unique id "%s"`, proc.UniqueID())
-	return proc, nil
+	// Finalizer
+	go func() {
+		// Wait for all work
+		v.wg.Wait()
+
+		// Log message after successful termination
+		v.logger.Info("exited")
```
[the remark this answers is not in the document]
michaljurecko: Sorry, I'll send fix.

internal/pkg/service/common/servicectx/servicectx.go
```diff
@@ -200,7 +203,7 @@ func (v *Process) Add(operation func(ShutdownFn)) {
 
 	select {
 	case <-v.terminating:
-		v.logger.Errorf(`cannot Add operation: the Process is terminating`)
+		v.logger.ErrorfCtx(v.shutdownCtx, `cannot Add operation: the Process is terminating`)
```
michaljurecko, line 206: Two log message are problematic, there are 2 possible solutions:
1. Add a `ctx` argument to the `Add` and `OnShutdown` method, so the message will be logged in the parent context.
2. Store `shutdown` context and use it here - it is anti-pattern and we try to avoid it.
3. Use `context.Background()` - it is an unexpected state, but we dont want to stop shutdown via `panic`.   
  So we want DataDog to generate an alert that an error has occurred, we don't need more.

------------

I asked myself **where I want to see this log message?**

And it makes sense to me to see that something unexpected happens during the shutdown operation (sleep), so I prefered solution `2.` (or `3.`).

PR #1465, opened 2023-12-12
refator: Use any instead of interface{}

jachym-tousek-keboola:
700 changed lines looks like a lot but it's just a bulk replace + `make fix`.


michaljurecko (approved):
LGTM

cool, I've been planning this (and similar things like `.editorconfig`) for a long time, so it's nice that it happened. :slightly_smiling_face: 
​

PR #1469, opened 2023-12-12
feat: Add generic result to Put and Atomic operation

michaljurecko:
Part of: https://keboola.atlassian.net/browse/PSGO-286

**Changes:**
- Atomic etcd operation didn't have result, now it can carry a result value - result of the operation.

**Intro**
- We are using [etcd](https://etcd.io/) dabase in our Stream and Templates API.
- It is a KV store used also by the Kubernetes.
- [Etcd Watch API](https://etcd.io/docs/v3.5/learning/api/#watch-api) is important for us in Stream API to achieve fast operation without locks and pooling (without regularly checking changes from the client side).
- Etcd supports [transactions](https://etcd.io/docs/v3.5/learning/api/#transaction), but they are different from RDBS.
  - Transaction in etcd is a `if/then/else` tree.
- [etcdop](https://github.com/keboola/keboola-as-code/tree/michaljurecko-PSGO-286-atomic-op-result/internal/pkg/service/common/etcdop) and other packages in the directory provide high-level API for etcd:
  - It is something like ORM for RDBS.
  - It provides: serialization/deserialization to/from JSON, callbacks, joining operations into transactions:
    - Please read [documentation](https://github.com/keboola/keboola-as-code/blob/cad3e78bcfe887ab91ee9228ba61aaad2df9ddb0/internal/pkg/service/common/etcdop/etcdop.go#L1-L34).
    - And go through packages files to get an overview of them.
    - Read documentation about [AtomicOp](https://github.com/keboola/keboola-as-code/blob/cad3e78bcfe887ab91ee9228ba61aaad2df9ddb0/internal/pkg/service/common/etcdop/op/atomic.go#L14-L34).
 - In mentioned packages, [the generic programing](https://go.dev/doc/tutorial/generics) is heavily used. 
​

Test in the RFC branch are disabled now.
- As we finish the individual parts, we also fix the tests.
- That's why I'm attaching a screenshot of the tests regarding the changed packages.

image

### DIFF
--- a/internal/pkg/service/buffer/sink/tablesink/statistics/repository/delete.go
+++ b/internal/pkg/service/buffer/sink/tablesink/statistics/repository/delete.go
@@ -11,8 +11,8 @@ import (
 // Delete returns an etcd operation to delete all statistics associated with the object key.
 // Statistics for the storage.LevelTarget are not deleted but are rolled up to the parent object.
 // This operation should not be used separately but atomically together with the deletion of the object.
-func (r *Repository) Delete(objectKey fmt.Stringer) *op.AtomicOp {
-	ops := op.Atomic(r.client)
+func (r *Repository) Delete(objectKey fmt.Stringer) *op.AtomicOp[op.NoResult] {
+	ops := op.Atomic(r.client, &op.NoResult{})
 	for _, level := range storage.AllLevels() {
 		// Object prefix contains all statistics related to the object
 		objectPfx := r.schema.InLevel(level).InObject(objectKey)
--- a/internal/pkg/service/common/etcdop/etcdop.go
+++ b/internal/pkg/service/common/etcdop/etcdop.go
@@ -11,7 +11,7 @@
 // A new operation can be defined as a:
 // New<Operation>(<operation factory>, <response processor>) function.
 //
-// On Operation.Do(ctx, etcdClient) call :
+// On Operation.Do(ctx) call :
 //   - The <operation factory> is executed, result is an etcd operation.
 //   - The etcd operation is executed, result is an etcd response.
 //   - The <response processor> is executed, to process the etcd response.
--- a/internal/pkg/service/common/etcdop/key_test.go
+++ b/internal/pkg/service/common/etcdop/key_test.go
@@ -2,6 +2,7 @@ package etcdop
 
 import (
 	"context"
+	"github.com/stretchr/testify/require"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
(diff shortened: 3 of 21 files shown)


internal/pkg/service/common/etcdop/key_test.go
```diff
@@ -110,82 +111,84 @@ func TestTypedKeyOperations(t *testing.T) {
...
-	assert.NoError(t, err)
+	require.NoError(t, err)
 	assert.False(t, found)
 
 	// ------
 	// Put
-	assert.NoError(t, k.Put(client, "bar").Do(ctx).Err())
+	result, err := k.Put(client, "bar").Do(ctx).ResultOrErr()
+	require.NoError(t, err)
+	assert.Equal(t, fooType("bar"), result)
```
michaljurecko, line 126: Added assert in the unit test.

internal/pkg/service/common/etcdop/etcdop.go
```diff
@@ -11,7 +11,7 @@
 // A new operation can be defined as a:
 // New<Operation>(<operation factory>, <response processor>) function.
 //
-// On Operation.Do(ctx, etcdClient) call :
+// On Operation.Do(ctx) call :
```
michaljurecko, line 14: Typo.

internal/pkg/service/buffer/sink/tablesink/statistics/repository/delete.go
```diff
@@ -11,8 +11,8 @@ import (
 // Delete returns an etcd operation to delete all statistics associated with the object key.
 // Statistics for the storage.LevelTarget are not deleted but are rolled up to the parent object.
 // This operation should not be used separately but atomically together with the deletion of the object.
-func (r *Repository) Delete(objectKey fmt.Stringer) *op.AtomicOp {
-	ops := op.Atomic(r.client)
+func (r *Repository) Delete(objectKey fmt.Stringer) *op.AtomicOp[op.NoResult] {
+	ops := op.Atomic(r.client, &op.NoResult{})
```
michaljurecko, line 15: Fixes in statistics/repository.

internal/pkg/service/common/etcdop/key_test.go
```diff
@@ -22,81 +23,81 @@ func TestKeyOperations(t *testing.T) {
 
 	// Get - not found
 	kv, err := k.Get(client).Do(ctx).ResultOrErr()
-	assert.NoError(t, err)
+	require.NoError(t, err)
 	assert.Nil(t, kv)
```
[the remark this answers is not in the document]
michaljurecko: Yes. 

`assert` is default approach in go tests.

`require` should be use if there is no reason to continue, or there is even a threat of panic somewhere further in tests.
For example, an error has been expected, and there is no error, and there is some assert `err.Error()` later (`err == nil`).

PR #1475, opened 2023-12-13
feat: Add iterator WithResultTo, AndOnFirstPage and AndOnPage methods

michaljurecko:
Part of: https://keboola.atlassian.net/browse/PSGO-286

Changes:
- Added `WithResultTo`, `AndOnFirstPage` and `AndOnPage` methods to the typed iterator.

Note:
- `With` means `Set`
- `And` means `Add`
- if the value is immutable

---------------------

image

---------------------

## Intro

`Iterator` provides pagination over etcd get operations results.
- Each page is provided by one get request.
- It is primarily designed for the gradual processing of records, so that we do not have to load them all into memory.
- In some cases, we know that there will not be a lot of records and we need to load them all, the `All` method is used for that.
- To simplify code, I added `WithResultTo`, it is similar to `Add`, but records are stored to the slice provided as an argument.
- Other operations already have `WithResultTo` method, but it was missing for the iterator.

### DIFF
--- a/internal/pkg/service/buffer/sink/tablesink/storage/retry_test.go
+++ b/internal/pkg/service/buffer/sink/tablesink/storage/retry_test.go
@@ -1,24 +1,18 @@
 package storage
 
 import (
-	"context"
-	"strings"
+	"github.com/keboola/keboola-as-code/internal/pkg/service/buffer/sink/tablesink/storage/test/testvalidation"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
 
 	"github.com/keboola/keboola-as-code/internal/pkg/service/common/utctime"
-	"github.com/keboola/keboola-as-code/internal/pkg/validator"
 )
 
 func TestRetryable_Validation(t *testing.T) {
 	t.Parallel()
 
-	cases := []struct {
-		Name          string
-		ExpectedError string
-		Value         Retryable
-	}{
+	cases := testvalidation.TestCases[Retryable]{
 		{
 			Name:  "empty",
 			Value: Retryable{},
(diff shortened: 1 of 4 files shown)


internal/pkg/service/buffer/sink/tablesink/storage/retry_test.go
```diff
@@ -1,24 +1,18 @@
...
 
 func TestRetryable_Validation(t *testing.T) {
 	t.Parallel()
 
-	cases := []struct {
-		Name          string
-		ExpectedError string
-		Value         Retryable
-	}{
+	cases := testvalidation.TestCases[Retryable]{
```
michaljurecko, line 15: Unrelated fix, we have common type for this use-case.

internal/pkg/service/common/etcdop/iterator/iterator_typed.go
```diff
@@ -38,7 +39,37 @@ func (v DefinitionT[T]) Do(ctx context.Context, opts ...op.Option) *IteratorT[T]
...
+	return v.
+		ForEachOp(func(value T, header *Header) error {
+			*slice = append(*slice, value)
+			return nil
+		}).
+		AndOnFirstPage(func(response *etcd.GetResponse) error {
+			*slice = nil
+			return nil
+		})
+}
```
michaljurecko, line 56: I need to use iterator together with `AtomicOp`, so I need to store all results to the slice defined by the pointer.
Therefore I added `WithResultTo` method.

michaljurecko: `AtomicOp` performs retries on transaction failures, so at the beginning I have to reset the slice.
Therefore, I create `AndOnFirstPage` method.

internal/pkg/service/common/etcdop/iterator/iterator_typed.go
```diff
@@ -38,7 +39,37 @@ func (v DefinitionT[T]) Do(ctx context.Context, opts ...op.Option) *IteratorT[T]
...
+
+// AndOnPage registers a callback that is executed after each page is successfully loaded.
+func (v *ForEachOpT[T]) AndOnPage(fn onPageFn) *ForEachOpT[T] {
+	clone := *v
+	clone.onPage = append(clone.onPage, fn)
+	return &clone
 }
 
 func (v *ForEachOpT[T]) Op(ctx context.Context) (op.LowLevelOp, error) {
```
michaljurecko, line 75: `Op` can be used to include the operation in `TxnOp`/`AtomicOp`.

`Do` bellow can be use to perform the operation directly.

PR #1481, opened 2023-12-15
feat: Implement ctxattr package

jachym-tousek-keboola:
https://keboola.atlassian.net/browse/PSGO-326

### DIFF
--- a/internal/pkg/log/log.go
+++ b/internal/pkg/log/log.go
@@ -53,12 +53,14 @@ type baseLogger interface {
 }
 
 type contextLogger interface {
-	// Debug, Info, Warn and Error accept a template string which can contain placeholders to be replaced by ctxattr attributes.
-	// For example if the template contains "%status%" then this placeholder is replaced by an attribute with the key "status".
-	Debug(ctx context.Context, template string)
-	Info(ctx context.Context, template string)
-	Warn(ctx context.Context, template string)
-	Error(ctx context.Context, template string)
+	// Debug logs message in the debug level, you can use an attribute %placeholder% defined by the ctxattr package.
+	Debug(ctx context.Context, message string)
+	// Info logs message in the debug level, you can use an attribute %placeholder% defined by the ctxattr package.
+	Info(ctx context.Context, message string)
+	// Warn logs message in the debug level, you can use an attribute %placeholder% defined by the ctxattr package.
+	Warn(ctx context.Context, message string)
+	// Error logs message in the debug level, you can use an attribute %placeholder% defined by the ctxattr package.
+	Error(ctx context.Context, message string)
 
 	LogCtx(ctx context.Context, level string, args ...any)
 	DebugCtx(ctx context.Context, args ...any)
--- a/internal/pkg/log/logger.go
+++ b/internal/pkg/log/logger.go
@@ -46,30 +46,30 @@ func formatMessageUsingAttributes(message string, set *attribute.Set) string {
 	return strings.NewReplacer(replacements...).Replace(message)
 }
 
-func (l *zapLogger) Debug(ctx context.Context, template string) {
+func (l *zapLogger) Debug(ctx context.Context, message string) {
 	l.sugaredLogger.Desugar().Debug(
-		formatMessageUsingAttributes(template, ctxattr.Attributes(ctx)),
+		formatMessageUsingAttributes(message, ctxattr.Attributes(ctx)),
 		ctxattr.ZapFields(ctx)...,
 	)
 }
 
-func (l *zapLogger) Info(ctx context.Context, template string) {
+func (l *zapLogger) Info(ctx context.Context, message string) {
 	l.sugaredLogger.Desugar().Info(
-		formatMessageUsingAttributes(template, ctxattr.Attributes(ctx)),
+		formatMessageUsingAttributes(message, ctxattr.Attributes(ctx)),
 		ctxattr.ZapFields(ctx)...,
 	)
 }
 
-func (l *zapLogger) Warn(ctx context.Context, template string) {
+func (l *zapLogger) Warn(ctx context.Context, message string) {
 	l.sugaredLogger.Desugar().Warn(
-		formatMessageUsingAttributes(template, ctxattr.Attributes(ctx)),
+		formatMessageUsingAttributes(message, ctxattr.Attributes(ctx)),
 		ctxattr.ZapFields(ctx)...,
 	)
 }
 
-func (l *zapLogger) Error(ctx context.Context, template string) {
+func (l *zapLogger) Error(ctx context.Context, message string) {
 	l.sugaredLogger.Desugar().Error(
-		formatMessageUsingAttributes(template, ctxattr.Attributes(ctx)),
+		formatMessageUsingAttributes(message, ctxattr.Attributes(ctx)),
 		ctxattr.ZapFields(ctx)...,
 	)
 }


internal/pkg/ctxattr/ctxattr.go
```diff
@@ -38,6 +38,10 @@ func ContextWith(ctx context.Context, attributes ...attribute.KeyValue) context.
 }
 
 func Attributes(ctx context.Context) *attribute.Set {
+	if ctx == nil {
```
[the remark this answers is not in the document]
michaljurecko: Interface can be nil. https://go.dev/tour/methods/13

Nil context is a problem, I prefer `panic`, and `context.Background()` in tests.

internal/pkg/ctxattr/ctxattr.go
```diff
@@ -0,0 +1,90 @@
...
+
+func ZapFields(ctx context.Context) []zap.Field {
+	if ctx == nil {
+		return []zap.Field{}
+	}
+
+	value := ctx.Value(contextZapFields)
+	if value != nil {
+		return value.([]zap.Field)
+	}
```
michaljurecko, line 61: Type assertion can be written in two ways:
- `value := interface.(myType)` - if the interface is not the `myType`, a panic occurs
- `value, ok := interface.(myType)` - if the interface is not the `myType`, the value is empty and `ok=false`
  - this is safer
  
---------------------
  
So I prefer:
```
if v, ok := ctx.Value(contextZapFields).([]zap.Field); ok {
    return v
}
```

But in your code, in this particular case, both ways are safe.

internal/pkg/ctxattr/ctxattr.go
```diff
@@ -0,0 +1,90 @@
+package ctxattr
```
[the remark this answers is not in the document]
michaljurecko: Please add a package comment, `// Package ctxattr provides ....`

internal/pkg/ctxattr/ctxattr_test.go
```diff
@@ -0,0 +1,90 @@
...
+)
+
+func TestContextAttributes(t *testing.T) {
+	t.Parallel()
+
+	ctx := context.Background()
+
+	set := Attributes(ctx)
+	_, ok := set.Value("isValid")
+	assert.False(t, ok)
```
michaljurecko, line 20: I don't know if these 2 lines are necessary here, assert on `Len` is enough.

```
_, ok := set.Value("isValid")
assert.False(t, ok)
```

internal/pkg/ctxattr/ctxattr_test.go
```diff
@@ -0,0 +1,90 @@
...
+	require.True(t, ok)
+	assert.Equal(t, "false", value.Emit())
+
+	value, ok = set.Value("status")
+	require.True(t, ok)
+	assert.Equal(t, "success", value.Emit())
+
+	value, ok = set.Value("count")
+	require.True(t, ok)
+	assert.Equal(t, "5", value.Emit())
```
michaljurecko, line 42: Can be simplified.

From:
```
assert.Equal(t, 3, set.Len())

value, ok := set.Value("isValid")
require.True(t, ok)
assert.Equal(t, "false", value.Emit())

value, ok = set.Value("status")
require.True(t, ok)
assert.Equal(t, "success", value.Emit())

value, ok = set.Value("count")
require.True(t, ok)
assert.Equal(t, "5", value.Emit())
```

To:
```
assert.Equal(t, attribute.NewSet(
	attribute.Bool("isValid", false), // last value wins
	attribute.String("status", "success"),
	attribute.Int("count", 5),
), *set)
```

-------------

Or
```
expected := attribute.NewSet(
	attribute.Bool("isValid", false), // last value wins
	attribute.String("status", "success"),
	attribute.Int("count", 5),
)
assert.True(t, expected.Equals(Attributes(ctx)))
```

internal/pkg/ctxattr/ctxattr_test.go
```diff
@@ -0,0 +1,90 @@
...
+	assert.True(t, fields[2].Equals(zap.String("status", "success")))
+}
+
+func TestAttributeConversion(t *testing.T) {
+	t.Parallel()
+
+	assert.Equal(t, zap.Bool("isValid", true), convertAttributeToZapField(attribute.Bool("isValid", true)))
+	assert.Equal(t, zap.Int("count", 5), convertAttributeToZapField(attribute.Int("count", 5)))
+	assert.Equal(t, zap.Float64("result", 3.14), convertAttributeToZapField(attribute.Float64("result", 3.14)))
+	assert.Equal(t, zap.String("result", "success"), convertAttributeToZapField(attribute.String("result", "success")))
```
michaljurecko, line 58: Table driven test would be better for reading, checking and editing.

```
t.Parallel()

	cases := []struct {
		Name     string
		In       attribute.KeyValue
		Expected zap.Field
	}{
		{
			Name:     "bool",
			In:       attribute.Bool("isValid", true),
			Expected: zap.Bool("isValid", true),
		},
		{
			Name:     "int",
			In:       attribute.Int("count", 5),
			Expected: zap.Int("count", 5),
		},
		{
			Name:     "float64",
			In:       attribute.Float64("result", 3.14),
			Expected: zap.Float64("result", 3.14),
		},
		{
			Name:     "string",
			In:       attribute.String("result", "success"),
			Expected: zap.String("result", "success"),
		},
		{
			Name:     "booleans",
			In:       attribute.BoolSlice("booleans", []bool{true, false}),
			Expected: zap.Bools("booleans", []bool{true, false}),
		},
		{
			Name:     "integers",
			In:       attribute.IntSlice("integers", []int{0, 5, -10}),
			Expected: zap.Int64s("integers", []int64{0, 5, -10}),
		},
		{
			Name:     "floats",
			In:       attribute.Float64Slice("floats", []float64{0.0, 5.4, -10.7}),
			Expected: zap.Float64s("floats", []float64{0.0, 5.4, -10.7}),
		},
		{
			Name:     "strings",
			In:       attribute.StringSlice("strings", []string{"success", "error"}),
			Expected: zap.Strings("strings", []string{"success", "error"}),
		},
	}

	for _, tc := range cases {
		assert.Equal(t, tc.Expected, convertAttributeToZapField(tc.In), tc.Name)
	}
```

------------

- `name` is optional, you can use test case index in the assert error message.
- you can simply convert it to parallel sub tests (not needed here)
```
	for _, tc := range cases {
		t.Run(tc.Name, func(t *testing.T) {
		        t.Parallel()
			assert.Equal(t, tc.Expected, convertAttributeToZapField(tc.In))	
		})
	}
```

internal/pkg/ctxattr/ctxattr.go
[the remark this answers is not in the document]
michaljurecko: image

internal/pkg/ctxattr/ctxattr.go
```diff
@@ -0,0 +1,90 @@
...
+	if ctx == nil {
+		return []zap.Field{}
+	}
+
+	value := ctx.Value(contextZapFields)
+	if value != nil {
+		return value.([]zap.Field)
+	}
+
+	return []zap.Field{}
```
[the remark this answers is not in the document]
michaljurecko: Maybe some linter can check it.

internal/pkg/ctxattr/ctxattr_test.go
```diff
@@ -0,0 +1,90 @@
...
+	value, ok = set.Value("count")
+	require.True(t, ok)
+	assert.Equal(t, "5", value.Emit())
+
+	fields = ZapFields(ctx)
+	assert.Len(t, fields, 3)
+
+	assert.True(t, fields[0].Equals(zap.Int64("count", 5)))
+	assert.True(t, fields[1].Equals(zap.Bool("isValid", false)))
+	assert.True(t, fields[2].Equals(zap.String("status", "success")))
```
michaljurecko, line 49: In can be simplified

From:
```
assert.Len(t, fields, 3)
assert.True(t, fields[0].Equals(zap.Int64("count", 5)))
assert.True(t, fields[1].Equals(zap.Bool("isValid", false)))
assert.True(t, fields[2].Equals(zap.String("status", "success")))
```

To:
```
assert.Equal(t, []zap.Field{
  zap.Int64("count", 5),
  zap.Bool("isValid", false),
  zap.String("status", "success"),
}, fields)
```


Pros: 
- It's shorter, it reads better.
- It is easier to edit.

internal/pkg/ctxattr/ctxattr_test.go
```diff
@@ -80,3 +81,10 @@ func TestAttributeConversion(t *testing.T) {
 		convertAttributeToZapField(attribute.StringSlice("strings", []string{"success", "error"})),
 	)
 }
+
+func TestNilContext(t *testing.T) {
+	t.Parallel()
+
+	assert.Equal(t, 0, Attributes(nil).Len()) // nolint: staticcheck
+	assert.Len(t, ZapFields(nil), 0)          // nolint: staticcheck
```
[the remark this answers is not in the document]
michaljurecko: https://github.com/stretchr/testify/issues/289

internal/pkg/log/log.go
```diff
@@ -53,6 +53,11 @@ type baseLogger interface {
 }
 
 type contextLogger interface {
+	Debug(ctx context.Context, template string)
+	Info(ctx context.Context, template string)
+	Warn(ctx context.Context, template string)
+	Error(ctx context.Context, template string)
+
```
michaljurecko, line 60: Please add some comments here, how it works.

internal/pkg/log/logger.go
```diff
@@ -33,57 +38,103 @@ func (l *zapLogger) AddPrefix(prefix string) Logger {
 	return clone
 }
 
+func formatMessageUsingAttributes(message string, set *attribute.Set) string {
+	for _, keyValue := range set.ToSlice() {
+		message = strings.ReplaceAll(message, "%"+string(keyValue.Key)+"%", keyValue.Value.Emit())
+	}
```
michaljurecko, line 44: BTW there is also `strings.Replacer`, but it is not issue here.
https://medium.com/@vikram.ingawale91/golang-strings-replace-vs-strings-replacer-a7b2d2b71593

internal/pkg/log/log.go
michaljurecko: Changes in the `logger` package can be a separate PR, but can be also here.

**But I am missing tests for the logger changes.**

michaljurecko (commented):
Good work! I like it.

I have some notes:

internal/pkg/ctxattr/ctxattr.go
michaljurecko: image

internal/pkg/log/logger.go
michaljurecko: Yes, bu overall, there won't be so many logs that it will have a big impact (I think).

For `strings.Replacer`, you has to allocate `oldNew` slice, which also costs something.

I mentioned it mainly so you know it exists.

internal/pkg/log/log.go
```diff
@@ -53,6 +53,13 @@ type baseLogger interface {
 }
 
 type contextLogger interface {
+	// Debug, Info, Warn and Error accept a template string which can contain placeholders to be replaced by ctxattr attributes.
+	// For example if the template contains "%status%" then this placeholder is replaced by an attribute with the key "status".
+	Debug(ctx context.Context, template string)
+	Info(ctx context.Context, template string)
+	Warn(ctx context.Context, template string)
+	Error(ctx context.Context, template string)
+
```
michaljurecko, line 62: Each method should have its own comment in the format: <methodName> ...., it doesn't matter if it is duplicated.

`Debug logs message in the debug level, you can use an attribute %placeholder% defined by the ctxattr package.`

Maybe I would choose the variable name `message` and not `template`, but I'll leave that up to you.

michaljurecko (approved):
LGTM, one note:

internal/pkg/log/log.go
michaljurecko: Thanks.

PR #1487, opened 2023-12-19
build: Reenable linters

jachym-tousek-keboola:


internal/pkg/service/common/cliconfig/flags.go
```diff
@@ -64,7 +64,8 @@ func flagsFromStruct(config any, fs *pflag.FlagSet, parents []string) error {
 
 		// Get optional usage tag
 		usage := field.Tag.Get("usage")
-		fieldPath := append(parents, partialName)
+		fieldPath := parents
+		fieldPath = append(fieldPath, partialName)
```
michaljurecko, line 68: I think there is no change in functionality.

I think the warning from the linter correctly indicates a problem in the code, and this is not a fix.

Can you please copy here what the linter wrote to you?

-----

In Go, it is not obvious how slices work in detail, try looking at this example.

Guess what the program prints and then run it.

https://go.dev/play/p/iiGLYofQ8N4

internal/pkg/service/common/etcdop/watch.go
```diff
@@ -371,21 +372,22 @@ func wrapStreamWithRestart(ctx context.Context, channelFactory func(ctx context.
...
 						continue
-					} else {
+					default:
 						// Convert initialization error
 						// from an 1+ attempt to a common error and restart watch.
 						resp.Err = err
 						resp.InitErr = nil
 						stream.channel <- resp
-						break
+						break ChannelLoop
```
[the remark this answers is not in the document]
michaljurecko: > This is the only way to satisfy gocritic here but maybe it would be better to revert this change and just ignore the error?

yes, in this case is IF simpler solution

internal/pkg/service/common/etcdop/op/txn.go
```diff
@@ -140,7 +140,7 @@ func (v *TxnOpDef) Txn(ctx context.Context) *TxnOp {
 		txn.Else(newInlineOp(etcd.OpTxn(subCmps, []etcd.Op{}, subElse), nil, item.MapResponse))
 	}
 
-	txn.processors = v.processors[:]
+	txn.processors = v.processors
```
michaljurecko, line 143: There was (and is) logical error, slice should be copied, so `v.processors` and `txn.processors` are decoupled.

We can discuss it on standup call.

```
txn.processors = append([]txnProcessor(nil), v.processors...) // (copy slice)
```

build/ci/golangci.yml
```diff
@@ -199,3 +197,8 @@ issues:
       text: "string `https?` has .+ occurrences, make it a constant"
       linters:
         - goconst
+    # Allow overriding ctx even if not used yet
+    - path: ".*"
+      text: "assigned to ctx, but never used afterwards"
+      linters:
+        - wastedassign
```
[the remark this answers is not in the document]
michaljurecko: > it would be really easy to forget to change it back once the ctx gets used for something.

yes! :+1:

michaljurecko (commented):
LGTM

we need to resolve issues with slices

internal/pkg/service/common/etcdop/op/txn.go
michaljurecko: (I didn't realize that `[:]` opeartor doesn't copy slices either.)

michaljurecko: Yes, therefore, it is always safe to append only to the same variable, either a new slice is created or the existing unused capacity in the old slice is used.

This validates the linter you turned on. It can be simply forgotten.

This characteristic of slices is intensively used, for example, in buffers. Memory allocation is an expensive operation.

michaljurecko:
```
txn.processors = append([]txnProcessor(nil), v.processors...)
```

PR #1495, opened 2024-01-04
feat: Use json encoder in NewDebugLogger

jachym-tousek-keboola:
Jira: https://keboola.atlassian.net/browse/PSGO-337

**Changes:**
- Debug logger is using json format internally

-----------

### DIFF
--- a/internal/pkg/service/cli/cmd/cmd_test.go
+++ b/internal/pkg/service/cli/cmd/cmd_test.go
@@ -7,6 +7,7 @@ import (
 	"strings"
 	"testing"
 
+	"github.com/keboola/go-utils/pkg/wildcards"
 	"github.com/spf13/pflag"
 	"github.com/stretchr/testify/assert"
 
@@ -172,7 +173,10 @@ func TestTearDown_Panic(t *testing.T) {
 	root.logger = logger
 	exitCode := root.tearDown(0, errors.New("panic error"))
 	assert.Equal(t, 1, exitCode)
-	expected := `INFO  
+	expected := `
+DEBUG  Unexpected panic: panic error
+%A
+INFO  
 ---------------------------------------------------
 Keboola CLI had a problem and crashed.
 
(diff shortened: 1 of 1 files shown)


internal/pkg/encoding/json/schema/schema_test.go
```diff
@@ -281,16 +269,8 @@ func TestValidateObjects_InvalidSchema_Warning2(t *testing.T) {
...
-WARN  config row JSON schema of the component "foo.bar" is invalid, please contact support:
-- anyOf failed:
-  - doesn't validate with "/definitions/simpleTypes":
-    - "properties.foo.type" is invalid: value must be one of "array", "boolean", "integer", "null", "number", "object", "string"
-  - "properties.foo.type" is invalid: expected array, but got string
+{"level":"warn","message":"config JSON schema of the component \"foo.bar\" is invalid, please contact support:\n- anyOf failed:\n  - doesn't validate with \"/definitions/simpleTypes\":\n    - \"properties.foo.type\" is invalid: value must be one of \"array\", \"boolean\", \"integer\", \"null\", \"number\", \"object\", \"string\"\n  - \"properties.foo.type\" is invalid: expected array, but got string"}
+{"level":"warn","message":"config row JSON schema of the component \"foo.bar\" is invalid, please contact support:\n- anyOf failed:\n  - doesn't validate with \"/definitions/simpleTypes\":\n    - \"properties.foo.type\" is invalid: value must be one of \"array\", \"boolean\", \"integer\", \"null\", \"number\", \"object\", \"string\"\n  - \"properties.foo.type\" is invalid: expected array, but got string"}
```
michaljurecko, line 273: for discussion: in most cases in this PR it works beautifully, ... but in these CLI related things it is quite difficult to read.
Would it be a problem to add an `AllMessagesTxt()` method that would return only messages? To simplify these assertions for the CLI? so there will be no change in the format here.

michaljurecko (approved):
LGTM

PR #1511, opened 2024-01-15
refactor: Remove baseLogger and LevelWriter

jachym-tousek-keboola:
Jira: https://keboola.atlassian.net/browse/PSGO-350

**Changes:**
- Removed baseLogger and LevelWriter
- Existing usages of LevelWriter replaced by os.Stdout, os.Stderr or Logger

-----------


pkg/lib/operation/project/local/encrypt/operation.go
```diff
@@ -37,7 +38,7 @@ func Run(ctx context.Context, projectState *project.State, o Options, d dependen
 
 	// Log plan
 	if !plan.Empty() || o.LogEmpty {
-		plan.Log(logger)
+		plan.Log(os.Stdout)
 	}
```
michaljurecko, line 43: This singleton/global way is simpler, but dependencies ways is better.

We used it in the past for: envs, logger, telemetry ... and it was always bad choice.

If you didn't run into any problems with tests, then this is an acceptable solution for me. 

But I think that one day it will block someone that he can't verify stdout/stderr.

It is also spamming tests output in verbose mode. Other ways how to print to stdout/stderr are blocked by the linter.
​

michaljurecko: Or are you planning it in the next PR? What do you think?

michaljurecko (approved):
LGTM

PR #1519, opened 2024-01-16
fix: Invoke processors for nested transactions at first

michaljurecko:
Part of: https://keboola.atlassian.net/browse/PSGO-352

**Changes:**
- Fixed calling order of nested and top-level transactions callbacks.
  - Nested callback must be performed first, it may have an effect that affects top-level callback.

-----------

### DIFF
--- a/internal/pkg/service/common/etcdop/op/txn.go
+++ b/internal/pkg/service/common/etcdop/op/txn.go
@@ -141,10 +141,6 @@ func (v *TxnOp[R]) lowLevelTxn(ctx context.Context) (*lowLevelTxn[R], error) {
 	out := &lowLevelTxn[R]{result: v.result, client: v.client, thenOps: make([]etcd.Op, 0), elseOps: make([]etcd.Op, 0)}
 	errs := errors.NewMultiError()
 
-	// Copy processors
-	out.processors = make([]func(ctx context.Context, r *TxnResult[R]), len(v.processors))
-	copy(out.processors, v.processors)
-
 	// Copy IFs
 	out.ifs = make([]etcd.Cmp, len(v.ifs))
 	copy(out.ifs, v.ifs)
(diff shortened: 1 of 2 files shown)


internal/pkg/service/common/etcdop/op/txn.go
```diff
@@ -141,10 +141,6 @@ func (v *TxnOp[R]) lowLevelTxn(ctx context.Context) (*lowLevelTxn[R], error) {
 	out := &lowLevelTxn[R]{result: v.result, client: v.client, thenOps: make([]etcd.Op, 0), elseOps: make([]etcd.Op, 0)}
 	errs := errors.NewMultiError()
 
-	// Copy processors
-	out.processors = make([]func(ctx context.Context, r *TxnResult[R]), len(v.processors))
-	copy(out.processors, v.processors)
-
 	// Copy IFs
```
michaljurecko, line 144: Here's another minor bug I found.

Callbacks from a nested transaction must be performed before callbacks from a parent transaction (if transactions are merged).

Because, a callback from a nested transaction can modify the result, for a callback from a parent transaction,

The error did not appear in tests because there is another bug, see next PR.

-----

So this code is moved to the end of the method, after `andOps` callbacks.

internal/pkg/service/common/etcdop/op/txn_test.go
```diff
@@ -412,8 +412,12 @@ func TestTxnOp_And_RealExample(t *testing.T) {
...
 	txn.Else(etcdop.Key("key/txn/succeeded").Put(client, "false"))
-	txn.OnResult(func(r *TxnResult[NoResult]) {
-		_, _ = fmt.Fprintf(&log, "txn succeeded: %t\n", r.Succeeded())
+	txn.AddProcessor(func(ctx context.Context, r *TxnResult[NoResult]) {
+		if err := r.Err(); err != nil {
+			_, _ = fmt.Fprintf(&log, "txn succeeded: error: %s", strings.ReplaceAll(err.Error(), "\n", ";"))
+		} else {
+			_, _ = fmt.Fprintf(&log, "txn succeeded: %t\n", r.Succeeded())
+		}
 	})
```
michaljurecko, line 421: Updated tests:

- `OnResult` callback is invoked if there is no error.
- Error can be generated by a previous callback.
- Therefore, because the top-level callbacks have been moved to the end, now they may be called after some other callback that generate errors.
- Result was that the `txn succeeded: ` message disappeared from some tests.

So I replaced `OnResult` shortcut with `AddProcessor` method - the callback is called always.
- I also distinguish between an error and a failed transaction (some IF condition was not met).
- See changes bellow, it should be clear.

PR #1521, opened 2024-01-17
fix: Replace misused Txn.Then with Txn.Merge(And)

michaljurecko:
Jira: https://keboola.atlassian.net/browse/PSGO-352

--------------

**Issue:**
- In multiple places in the code was used `txn.Then(subTxn)` instead of `txn.Merge/And(subTxn)`.
- - `Then` and `Merge` operations are valid and tested, but was misused.
- If all IF conditions in transactions are fulfilled, then both operations behave same :exclamation: 
- The problem is that if SOME IF conditions are not met, the sub transaction is applied, if `Then` is used.
- With `Merge/And` ALL IF conditions, from all sub-transactions must be meet, then all sub-txns are applied.
- The misused `Then/Merge` calls were also in the `AtomicOp` impl.

---------------------------

**Changes:**
- `Txn.And` merhod has been renamed to `Txn.Merge` to be more clear.
- `Then` method cannot be called with a transaction.
  - I created a new method `ThenTxn` - it can be called only with a transaction.
  - **In this way, the method interface helps us to avoid mistakes.**
  - If you want to add a transaction, you have to make a decision, and select `ThenTxn` or `Merge` method.
- AtomicOp uses the `Merge` instead of the `Then`, :arrow_left: there was also a mistake here.
- Added more unit tests.

---------------------------

**Details:**
[etcd transaction](https://etcd.io/docs/v3.5/learning/api/#transaction):
  - Has `0..n` IF conditions.
  - Has `0..n` THEN operations.
  - Has `0..n` ELSE operations.
  - THEN and ELSE branches may contain nested transaction.

Our high-level interface `op.TxnOp` match low-level interface, has `If`, `Then` and `Else` methods.
- So these methods must be preserved.

Our use-case is to always `Merge` transactions, and do either everything or nothing. 

But nested transactions via `ThenTxn` can also be a valid use-case, but we don't have it yet.

---------------

**Tests**

See tests: `TestTxnOp_Then_Simple`, `TestTxnOp_Merge_Simple`, `TestTxnOp_Then_Simple_Ifs` and `TestTxnOp_Merge_Simple_Ifs` to see differences between `Then` and `Merge` ops.

Main difference is here:
`Then` - `TestTxnOp_Then_Simple_Ifs`
https://github.com/keboola/keboola-as-code/blob/00e5cbdb503513707adca3984f75c484e99dd4f7/internal/pkg/service/common/etcdop/op/txn_test.go#L597-L607

`Merge` - `TestTxnOp_Merge_Simple_Ifs`
https://github.com/keboola/keboola-as-code/blob/00e5cbdb503513707adca3984f75c484e99dd4f7/internal/pkg/service/common/etcdop/op/txn_test.go#L703-L715

-----------

**Future improvements:**
- In a next PR, I will lock File/Slice rotation transaction by a snapshot in a test.
  - To prevent future bugs in this important Stream API part.

### DIFF
--- a/internal/pkg/service/buffer/sink/tablesink/storage/repository/file_repo.go
+++ b/internal/pkg/service/buffer/sink/tablesink/storage/repository/file_repo.go
@@ -163,7 +163,7 @@ func (r *FileRepository) StateTransition(now time.Time, fileKey storage.FileKey,
 
 				// Save modified value, if there is a difference
 				if !reflect.DeepEqual(oldSliceState, slice) {
-					txn.Then(r.all.slice.updateTxn(oldSliceState, slice))
+					txn.Merge(r.all.slice.updateTxn(oldSliceState, slice))
 				}
 			}
 
@@ -187,7 +187,7 @@ func (r *FileRepository) Delete(k storage.FileKey) *op.TxnOp[op.NoResult] {
 	txn := op.Txn(r.client)
 
 	// Delete entity from All prefix
-	txn.And(
+	txn.Merge(
 		r.schema.
 			AllLevels().ByKey(k).DeleteIfExists(r.client).
 			WithEmptyResultAsError(func() error {
@@ -201,7 +201,7 @@ func (r *FileRepository) Delete(k storage.FileKey) *op.TxnOp[op.NoResult] {
 	}
 
 	// Delete all slices
-	txn.And(r.all.slice.deleteAll(k))
+	txn.Merge(r.all.slice.deleteAll(k))
 
 	return txn
 }
@@ -349,7 +349,7 @@ func (r *FileRepository) rotateAllIn(rb rollback.Builder, now time.Time, parentK
 					return nil, serviceError.NewResourceAlreadyExistsError("file", oldFile.FileKey.String(), "sink")
 				} else if modified, err := oldFile.WithState(now, storage.FileClosing); err == nil {
 					// Switch the old file from the state storage.FileWriting to the state storage.FileClosing
-					txn.And(r.updateTxn(oldFile, modified))
+					txn.Merge(r.updateTxn(oldFile, modified))
 				} else {
 					errs.Append(err)
 				}
@@ -359,7 +359,7 @@ func (r *FileRepository) rotateAllIn(rb rollback.Builder, now time.Time, parentK
 			for _, oldSlice := range openedSlicesPerSink[sink.SinkKey] {
 				if modified, err := oldSlice.WithState(now, storage.SliceClosing); err == nil {
 					// Switch the old slice from the state storage.SliceWriting to the state storage.SliceClosing
-					txn.And(r.all.slice.updateTxn(oldSlice, modified))
+					txn.Merge(r.all.slice.updateTxn(oldSlice, modified))
 				} else {
 					errs.Append(err)
 				}
@@ -387,14 +387,14 @@ func (r *FileRepository) rotateAllIn(rb rollback.Builder, now time.Time, parentK
 				// Open slices in the assigned volumes
 				for _, volumeID := range file.Assignment.Volumes {
 					if slice, err := newSlice(now, file, volumeID, maxUsedDiskSpace[file.SinkKey]); err == nil {
-						txn.And(r.all.slice.createTxn(slice))
+						txn.Merge(r.all.slice.createTxn(slice))
 					} else {
 						errs.Append(err)
 					}
 				}
 
 				// Open file
-				txn.And(r.createTxn(file).OnResult(func(result *op.TxnResult[op.NoResult]) {
+				txn.Merge(r.createTxn(file).OnResult(func(result *op.TxnResult[op.NoResult]) {
 					if result.Succeeded() {
 						newFiles = append(newFiles, file)
 					}
--- a/internal/pkg/service/common/etcdop/op/atomic.go
+++ b/internal/pkg/service/common/etcdop/op/atomic.go
@@ -202,9 +202,10 @@ func (v *AtomicOp[R]) Do(ctx context.Context, opts ...Option) AtomicResult[R] {
 	var err error
 
 	for {
-		if ok, err = v.DoWithoutRetry(ctx, opts...); err != nil {
-			break
-		} else if !ok {
+		txnResult := v.DoWithoutRetry(ctx, opts...)
+		ok = txnResult.Succeeded()
+		err = txnResult.Err()
+		if err == nil && !ok {
 			attempt++
 			if delay := b.NextBackOff(); delay == backoff.Stop {
 				break
@@ -227,17 +228,15 @@ func (v *AtomicOp[R]) Do(ctx context.Context, opts ...Option) AtomicResult[R] {
 	return AtomicResult[R]{result: v.result, error: err, attempt: attempt, elapsedTime: elapsedTime}
 }
 
-func (v *AtomicOp[R]) DoWithoutRetry(ctx context.Context, opts ...Option) (bool, error) {
+func (v *AtomicOp[R]) DoWithoutRetry(ctx context.Context, opts ...Option) *TxnResult[R] {
 	tracker := NewTracker(v.client)
 
 	// Create READ operations
 	readTxn := Txn(tracker)
 	for _, opFactory := range v.readPhase {
-		op, err := opFactory(ctx)
-		if err != nil {
-			return false, err
-		}
-		if op != nil {
+		if op, err := opFactory(ctx); err != nil {
+			return newErrorTxnResult[R](err)
+		} else if op != nil {
 			readTxn.Merge(op)
 		}
 	}
@@ -245,7 +244,7 @@ func (v *AtomicOp[R]) DoWithoutRetry(ctx context.Context, opts ...Option) (bool,
 	// Run READ phase, track used keys/prefixes
 	readResult := readTxn.Do(ctx, opts...)
 	if err := readResult.Err(); err != nil {
-		return false, err
+		return newErrorTxnResult[R](err)
 	}
 
 	// Create IF part of the transaction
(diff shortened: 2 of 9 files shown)


internal/pkg/service/common/etcdop/op/txn.go
```diff
@@ -29,8 +29,9 @@ type TxnOp[R any] struct {
 	processors []func(ctx context.Context, r *TxnResult[R])
 	ifs        []etcd.Cmp
 	thenOps    []Op
+	thenTxnOps []Op // thenTxnOps are separated from thenOps to avoid confusing between Then and Merge
+	mergeOps   []Op
 	elseOps    []Op
-	andOps     []Op
```
michaljurecko, line 33: `andOps` -> `mergeOps`, added `thenTxnOps`.

internal/pkg/service/common/etcdop/op/txn.go
```diff
@@ -155,17 +164,29 @@ func (v *TxnOp[R]) lowLevelTxn(ctx context.Context) (*lowLevelTxn[R], error) {
...
 			out.addThen(lowLevel.Op, lowLevel.MapResponse)
+		}
+	}
+	for i, op := range v.thenTxnOps {
+		// Create low-level operation
+		if lowLevel, err := op.Op(ctx); err != nil {
+			errs.Append(errors.PrefixErrorf(err, "cannot create operation [thenTxn][%d]", i))
+		} else if !lowLevel.Op.IsTxn() {
+			errs.Append(errors.Errorf(`cannot create operation [thenTxn][%d]: operation "%T" is not a transaction, use Then, not ThenTxn`, i, lowLevel.Op))
 		} else {
```
michaljurecko: `Then/ThenTxn` checks are implemented in convertion to a low-level op.
  - Rather, we do not know whether it is a transaction, our `TxnOp` high-level operation may not be used.

internal/pkg/service/common/etcdop/op/txn_test.go
```diff
@@ -75,9 +75,9 @@ func TestTxnOp_OpError_Create(t *testing.T) {
 		Then(op).
 		Else(op).
 		Else(op).
-		And(op).
-		And(op).
-		And(Txn(client).Then(op))
+		Merge(op).
+		Merge(op).
+		Merge(Txn(client).Then(op))
```
michaljurecko, line 81: Updated Txn tests.

internal/pkg/service/common/etcdop/op/atomic.go
```diff
@@ -228,22 +228,23 @@ func (v *AtomicOp[R]) Do(ctx context.Context, opts ...Option) AtomicResult[R] {
...
+	readTxn := Txn(tracker)
 	for _, opFactory := range v.readPhase {
 		op, err := opFactory(ctx)
 		if err != nil {
 			return false, err
 		}
 		if op != nil {
-			getOps = append(getOps, op)
+			readTxn.Merge(op)
 		}
```
michaljurecko: Read phase: `Merge` instead of `Then` bellow, and removed unnecessery aux slice.

internal/pkg/service/common/etcdop/op/atomic.go
```diff
@@ -291,35 +292,35 @@ func (v *AtomicOp[R]) DoWithoutRetry(ctx context.Context, opts ...Option) (bool,
...
 		if op != nil {
-			writeOps = append(writeOps, op)
+			writeTxn.Merge(op)
 		}
 	}
 
 	// Create transaction
-	return TxnWithResult[R](v.client, v.result).Then(writeOps...).AddProcessor(func(ctx context.Context, r *TxnResult[R]) {
+	return writeTxn.AddProcessor(func(ctx context.Context, r *TxnResult[R]) {
```
michaljurecko: Write phase: `Merge` instead of `Then` bellow, and removed unnecessery aux slice.

internal/pkg/service/common/etcdop/op/atomic.go
```diff
@@ -227,25 +228,23 @@ func (v *AtomicOp[R]) Do(ctx context.Context, opts ...Option) AtomicResult[R] {
 	return AtomicResult[R]{result: v.result, error: err, attempt: attempt, elapsedTime: elapsedTime}
 }
 
-func (v *AtomicOp[R]) DoWithoutRetry(ctx context.Context, opts ...Option) (bool, error) {
+func (v *AtomicOp[R]) DoWithoutRetry(ctx context.Context, opts ...Option) *TxnResult[R] {
```
michaljurecko: Here, we can use `TxnResult`, the type has been added in some previous PR.

internal/pkg/service/common/etcdop/op/atomic_test.go
```diff
@@ -69,10 +69,10 @@ func TestAtomicUpdate(t *testing.T) {
 			key3.Put(client, "value"),
 			op.Txn(client).
 				If(etcd.Compare(etcd.Value("key4"), "=", "value")).
-				Then(
+				Merge(
 					key5.Get(client),
 					op.Txn(client).
-						If(etcd.Compare(etcd.Value("checkMissing"), "=", "value")).
+						If(etcd.Compare(etcd.Version("missing"), "=", 0)).
 						Then().
```
michaljurecko, line 76: Improved test, to use `Merge` operation - because we use it more often, `if` condition now pass by default.

internal/pkg/service/buffer/sink/tablesink/storage/repository/file_repo.go
```diff
@@ -163,7 +163,7 @@ func (r *FileRepository) StateTransition(now time.Time, fileKey storage.FileKey,
 
 				// Save modified value, if there is a difference
 				if !reflect.DeepEqual(oldSliceState, slice) {
-					txn.Then(r.all.slice.updateTxn(oldSliceState, slice))
+					txn.Merge(r.all.slice.updateTxn(oldSliceState, slice))
```
michaljurecko, line 166: In this commit, misused `Then` calls are fixed.

internal/pkg/service/common/etcdop/op/txn.go
```diff
@@ -170,7 +170,7 @@ func (v *TxnOp[R]) lowLevelTxn(ctx context.Context) (*lowLevelTxn[R], error) {
 		if lowLevel, err := op.Op(ctx); err != nil {
 			errs.Append(errors.PrefixErrorf(err, "cannot create operation [then][%d]", i))
 		} else if lowLevel.Op.IsTxn() {
-			errs.Append(errors.Errorf(`cannot create operation [then][%d]: operation "%T" is a transaction, use ThenTxn, not Then`, i, lowLevel.Op))
+			errs.Append(errors.Errorf(`cannot create operation [then][%d]: operation is a transaction, use ThenTxn, not Then`, i))
 		} else {
```
michaljurecko: Type is always `clientv3.Op`.

internal/pkg/service/common/etcdop/op/txn_test.go
```diff
@@ -363,6 +363,83 @@ func TestTxnOp_ThenTxn_CalledWithoutTxn(t *testing.T) {
 	}
 }
 
+func TestTxnOp_Then_Simple(t *testing.T) {
+	t.Parallel()
```
michaljurecko: There are already enough tests here, but I added another one to show the difference between ThenTxn and Merge.

internal/pkg/service/common/etcdop/op/txn.go
```diff
@@ -155,17 +164,29 @@ func (v *TxnOp[R]) lowLevelTxn(ctx context.Context) (*lowLevelTxn[R], error) {
...
 
-	// Map THEN operations
+	// Map Then and ThenTxn operations
 	for i, op := range v.thenOps {
 		// Create low-level operation
-		if lowLevel, err := op.Op(ctx); err == nil {
+		if lowLevel, err := op.Op(ctx); err != nil {
+			errs.Append(errors.PrefixErrorf(err, "cannot create operation [then][%d]", i))
+		} else if lowLevel.Op.IsTxn() {
+			errs.Append(errors.Errorf(`cannot create operation [then][%d]: operation "%T" is a transaction, use ThenTxn, not Then`, i, lowLevel.Op))
```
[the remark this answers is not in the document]
michaljurecko: Sure, it would be nice to check it sooner.

But to create a low level operation you need (in some cases) the context, which we don't want here.
- Composing and invoking phases are strictly separated.

I could check the *TxnOp type there, which will cover 99% of cases.
- I will do it now.

But you will still be able to do your own high-level operation, 
e.g. AcquireLock, which is implemented using a low-level transaction.

internal/pkg/service/common/etcdop/op/txn.go
```diff
@@ -75,6 +83,14 @@ func (v *TxnOp[R]) If(cs ...etcd.Cmp) *TxnOp[R] {
...
 // To add a transaction to the Then branch, use ThenTxn, or use Merge to merge transactions.
 func (v *TxnOp[R]) Then(ops ...Op) *TxnOp[R] {
+	// Check common high-level transaction types.
+	// Bulletproof check of the low-level transaction is in the "lowLevelTxn" method.
+	for i, op := range ops {
+		if _, ok := op.(txnInterface); ok {
+			panic(errors.Errorf(`invalid operation[%d]: op is a transaction, use ThenTxn, not Then`, i))
+		}
+	}
+
```
michaljurecko, line 93: I added the additional check, see https://github.com/keboola/keboola-as-code/pull/1521#discussion_r1455367562

internal/pkg/service/common/etcdop/op/txn.go
```diff
@@ -34,6 +34,11 @@ type TxnOp[R any] struct {
 	elseOps    []Op
 }
 
+// txnInterface is a marker interface for generic type TxnOp.
+type txnInterface interface {
+	txn()
+}
+
```
michaljurecko, line 41: You can check specific type, for example `v, ok := value.(*TxnOp[NoResult])`, but not all variants `v, ok := value.(*TxnOp[*])`, therefore I added the `txnInterface` - a marker interface with private method (so it can only be fulfilled from the same package).

michaljurecko (commented):
@jachym-tousek-keboola please check the improvement

internal/pkg/service/common/etcdop/op/txn.go
michaljurecko: No, because only `*TxnOp` will match it, but not your custom `AcquireLock` operation from your custom package.

PR #1523, opened 2024-01-17
feat: Append attributes from the context to the span

michaljurecko:
Jira: https://keboola.atlassian.net/browse/PSGO-367

**Changes:**
- Minor fixes.
- Added missing assignment of the `ctxattrs` to each span + updated test.

-----------

### DIFF
--- a/build/ci/golangci.yml
+++ b/build/ci/golangci.yml
@@ -1,5 +1,5 @@
 run:
-  timeout: 5m
+  timeout: 10m
   concurrency: 8
   tests: true # check test files
   skip-files: [ ]
--- a/internal/pkg/service/common/httpserver/middleware/logger.go
+++ b/internal/pkg/service/common/httpserver/middleware/logger.go
@@ -4,11 +4,9 @@ import (
 	"net/http"
 	"time"
 
-	"go.opentelemetry.io/otel/attribute"
 	goaHttpMiddleware "goa.design/goa/v3/http/middleware"
 
 	"github.com/keboola/keboola-as-code/internal/pkg/log"
-	"github.com/keboola/keboola-as-code/internal/pkg/service/common/ctxattr"
 	"github.com/keboola/keboola-as-code/internal/pkg/utils/ip"
 )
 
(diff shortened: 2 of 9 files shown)


internal/pkg/service/common/httpserver/middleware/logger.go
```diff
@@ -26,7 +27,6 @@ func Logger(baseLogger log.Logger) Middleware {
 
 			// Log
 			userAgent := req.Header.Get("User-Agent")
-			logger := baseLogger.WithComponent("http")
 			logger.Infof(
```
michaljurecko, line 30: Small memory optimalization.

internal/pkg/telemetry/ddtracer.go
```diff
@@ -18,19 +18,29 @@ type wrappedDDTracer struct {
...
+		)
+
+		span = &wrappedDDSpan{Span: span, tp: t.tp}
+		ctx = trace.ContextWithSpan(ctx, span)
+	}
+
+	// Append attributes from the context
+	if attrs := ctxattr.Attributes(ctx).ToSlice(); len(attrs) > 0 {
+		span.SetAttributes(attrs...)
+	}
```
michaljurecko, line 43: I think we forgot this.

internal/pkg/telemetry/ddtracer_test.go
```diff
@@ -21,39 +23,44 @@ func TestContextAttributes(t *testing.T) {
...
 
+	// Add some common context attribute, it should appear in span and log record
 	ctx := context.Background()
+	ctx = ctxattr.ContextWith(ctx, attribute.String("foo", "bar"))
 
+	// Create span
 	ctx, span := tel.Tracer().Start(ctx, "keboola.go.test")
 	defer span.End(nil)
 
+	// Tracer added TraceID and SpanID attributes to the context
```
michaljurecko, line 44: BTW: I know that I use comments more than is standard. But otherwise I can't quickly scan a code longer than 20 lines.

build/ci/golangci.yml
```diff
@@ -1,5 +1,5 @@
 run:
-  timeout: 5m
+  timeout: 10m
   concurrency: 8
```
michaljurecko, line 3: `5 minutes` is the limit value without cache

PR #1527, opened 2024-01-19
 fix: Fix iterator in TxnOp, don't load empty page

michaljurecko:
Jira: https://keboola.atlassian.net/browse/PSGO-370

**Changes:**
- Iterator in a tranction is no more loading an additional/unnecessary empty page.

-----------

See new - failing test:
image

### DIFF
--- a/internal/pkg/service/common/etcdop/iterator/iterator.go
+++ b/internal/pkg/service/common/etcdop/iterator/iterator.go
@@ -271,13 +271,11 @@ func (v *Iterator) moveToPage(resp *etcd.GetResponse) bool {
 		}
 	}
 
-	// Handle empty result
 	v.values = kvs
 	v.header = header
 	v.lastIndexOnPage = len(v.values) - 1
-	if v.lastIndexOnPage == -1 {
-		return false
-	}
+	v.indexOnPage = -1
+	v.page++
 
 	// Prepare next page
 	if more {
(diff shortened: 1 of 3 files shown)


internal/pkg/service/common/etcdop/iterator/iterator.go
```diff
@@ -292,9 +290,8 @@ func (v *Iterator) moveToPage(resp *etcd.GetResponse) bool {
 		v.start = end
 	}
 
-	v.indexOnPage = -1
-	v.page++
-	return true
+	// Stop on empty result
+	return v.lastIndexOnPage != -1
 }
```
michaljurecko, line 295: Iterator can be used in a TXN.
- Then, the first page is loaded withing TXN and next pages are loaded using `WithRevision`, so iteration is atomic, but also memory optimal.

But, I found, there are some unnecessary request in logs.

Problem:
- The `if v.lastIndexOnPage == -1 {` condition is in the code before `v.start = end`.
- So first call of the `Next/movePageTo` methods returns `false`.
- But next call of the `nextPage` don't stop on `v.start == end` (if `Next` is called again).
- The code usually iterates until `Next` returns `false` the first time, so we didn't notice it.

-------------

But in the TXN, the first page is injected here (`movePageTo` is part of the `Next` call).

https://github.com/keboola/keboola-as-code/blob/1448a99fa91c047e5bf616cc511752c4da4048ff/internal/pkg/service/common/etcdop/iterator/iterator.go#L96-L109

The `moveToPage` returns ^^^ here `false`, but it is not checked/important.

Later, in the code, the `Next` method calls `nextPage`, :arrow_right: :arrow_right: **so `v.start = end` must be set ALSO IF THE FIRST PAGE IS EMPTY.** :arrow_left: :arrow_left:

PR #1528, opened 2024-01-19
fix: Improve AtomicOP IF conditions and tests

michaljurecko:
Jira: https://keboola.atlassian.net/browse/PSGO-357

**Motivation:**
- As we found some issues regarding transactions, I thought it would be good to also check `AtomicOp` in order to avoid difficult-to-debug problems in production.

**Changes:**
- Fixed conditions in AtomicOp, some cases were not covered, added tests.

-----------

Commented per commit.

### DIFF
--- a/internal/pkg/service/common/etcdop/op/atomic.go
+++ b/internal/pkg/service/common/etcdop/op/atomic.go
@@ -250,44 +250,49 @@ func (v *AtomicOp[R]) DoWithoutRetry(ctx context.Context, opts ...Option) *TxnRe
 	// Create IF part of the transaction
 	var cmps []etcd.Cmp
 	for _, op := range removeOpsOverlaps(tracker.Operations()) {
-		switch op.Type {
-		case DeleteOp:
+		mustExist := (op.Type == GetOp || op.Type == PutOp) && op.Count > 0
+		mustNotExist := op.Type == DeleteOp || op.Count == 0
+		switch {
+		case mustExist:
+			// IF: 0 < modification version <= Read Phase revision
+			// Key/range exists and has not been modified since the Read Phase.
+			//
+			// Note: we cannot check that nothing was deleted from the prefix.
+			// The condition IF count == n is needed, and it is not implemented in etcd.
+			// We can verify that an individual key was deleted, its MOD == 0.
 			cmps = append(cmps,
-				// The key/prefix must be missing, version must be equal to 0.
+				// The key/prefix must exist, version must be NOT equal to 0.
 				etcd.Cmp{
-					Result:      etcdserverpb.Compare_EQUAL,
-					Target:      etcdserverpb.Compare_VERSION,
-					TargetUnion: &etcdserverpb.Compare_Version{Version: 0},
+					Target:      etcdserverpb.Compare_MOD,
+					Result:      etcdserverpb.Compare_GREATER,
+					TargetUnion: &etcdserverpb.Compare_ModRevision{ModRevision: 0},
+					Key:         op.Key,
+					RangeEnd:    op.RangeEnd, // may be empty
+				},
+				// The key/prefix cannot be modified between GET and UPDATE phase.
+				// Mod revision of the item must be less or equal to header.Revision.
+				etcd.Cmp{
+					Target:      etcdserverpb.Compare_MOD,
+					Result:      etcdserverpb.Compare_LESS, // see +1 bellow, so less or equal to header.Revision
+					TargetUnion: &etcdserverpb.Compare_ModRevision{ModRevision: readResult.Header().Revision + 1},
 					Key:         op.Key,
 					RangeEnd:    op.RangeEnd, // may be empty
 				},
 			)
-		case GetOp:
-			if op.Count > 0 {
-				cmps = append(cmps,
-					// The key/prefix must exist, version must be NOT equal to 0.
-					etcd.Cmp{
-						Result:      etcdserverpb.Compare_GREATER,
-						Target:      etcdserverpb.Compare_VERSION,
-						TargetUnion: &etcdserverpb.Compare_Version{Version: 0},
-						Key:         op.Key,
-						RangeEnd:    op.RangeEnd, // may be empty
-					})
-			}
-			fallthrough
-		case PutOp:
+		case mustNotExist:
 			cmps = append(cmps,
-				// The key/prefix cannot be modified between GET and UPDATE phase.
-				// Mod revision of the item must be less or equal to header.Revision.
+				// IF: modification version == 0
+				// The key/range doesn't exist.
 				etcd.Cmp{
-					Result:      etcdserverpb.Compare_LESS, // see +1 bellow, so less or equal to header.Revision
 					Target:      etcdserverpb.Compare_MOD,
-					TargetUnion: &etcdserverpb.Compare_ModRevision{ModRevision: readResult.Header().Revision + 1},
+					Result:      etcdserverpb.Compare_EQUAL,
+					TargetUnion: &etcdserverpb.Compare_ModRevision{ModRevision: 0},
 					Key:         op.Key,
 					RangeEnd:    op.RangeEnd, // may be empty
-				})
+				},
+			)
 		default:
-			panic(errors.Errorf(`unexpected operation type "%v"`, op.Type))
+			panic(errors.Errorf(`unexpected state, operation type "%v"`, op.Type))
 		}
 	}
 
--- a/internal/pkg/service/common/etcdop/op/atomic.go
+++ b/internal/pkg/service/common/etcdop/op/atomic.go
@@ -32,11 +32,12 @@ import (
 //
 // Retries on network errors are always performed.
 type AtomicOp[R any] struct {
-	client     etcd.KV
-	result     *R
-	readPhase  []HighLevelFactory
-	writePhase []HighLevelFactory
-	processors processors[R]
+	client         etcd.KV
+	result         *R
+	readPhase      []HighLevelFactory
+	writePhase     []HighLevelFactory
+	processors     processors[R]
+	checkPrefixKey bool // checkPrefixKey - see SkipPrefixKeysCheck method documentation
 }
 
 type AtomicOpInterface interface {
@@ -45,7 +46,22 @@ type AtomicOpInterface interface {
 }
 
 func Atomic[R any](client etcd.KV, result *R) *AtomicOp[R] {
-	return &AtomicOp[R]{client: client, result: result}
+	return &AtomicOp[R]{client: client, result: result, checkPrefixKey: true}
+}
+
+// SkipPrefixKeysCheck disables the feature.
+//
+// By default, the feature is enabled and checks that each loaded key within the Read Phase, from a prefix, exists in Write Phase.
+// This can be potentially SLOW and generate a lot of IF conditions, if there are a large number of keys in the prefix.
+// Therefore, this feature can be turned off by the method.
+//
+// Modification of a key in the prefix is always detected,
+// this feature is used to detect the deletion of a key from the prefix.
+//
+// See TestAtomicOp:GetPrefix_DeleteKey_SkipPrefixKeysCheck.
+func (v *AtomicOp[R]) SkipPrefixKeysCheck() *AtomicOp[R] {
+	v.checkPrefixKey = false
+	return v
 }
 
 func (v *AtomicOp[R]) AddFrom(ops ...AtomicOpInterface) *AtomicOp[R] {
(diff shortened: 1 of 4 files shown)


internal/pkg/service/common/etcdop/op/atomic_test.go
```diff
@@ -13,10 +14,290 @@ import (
...
+	Name               string
+	Prepare            func(t *testing.T, client etcd.KV) []op.Op
+	ReadPhase          func(t *testing.T, client etcd.KV) []op.Op
+	BreakingChange     func(t *testing.T, client etcd.KV) []op.Op
+	ExpectedWritePhase string
+}
+
+func TestAtomicOp(t *testing.T) {
+	t.Parallel()
+
```
michaljurecko, line 35: In the first commit a added better tests for the AtomicOp.

Each case it tested twice:
- First run tests the AtomicOp without the `BreakingChange` between Read and Write phase.
- Second run tests it with the `BreakingChange`.

internal/pkg/service/common/etcdop/op/atomic_test.go
```diff
@@ -13,10 +14,290 @@ import (
...
+			ReadPhase:      getKey,
+			BreakingChange: putKey,
+			ExpectedWritePhase: `
+➡️  TXN
+  ➡️  IF:
+  001 "key/1" MOD EQUAL 0
+  ➡️  THEN:
+  001 ➡️  PUT "foo"
+
+✔️  TXN | succeeded: false
```
michaljurecko, line 74: TXN snapshot of the Write Phase:

`key/1` has been read by the get operation in the Read Phase - but it doesn't exist.
So in the Write Phase it must still not exist.

internal/pkg/service/common/etcdop/op/atomic_test.go
```diff
@@ -13,10 +14,290 @@ import (
...
+			// Read Phase gets a key, it exists, but before the Write Phase, it is modified.
+			Name:           "GetKey_ModifyKey",
+			Prepare:        putKey,
+			ReadPhase:      getKey,
+			BreakingChange: putKey,
+			ExpectedWritePhase: `
+➡️  TXN
+  ➡️  IF:
+  001 "key/1" MOD GREATER 0
+  002 "key/1" MOD LESS %d
```
michaljurecko, line 87: `key/1` has been read by the get operation in the Read Phase - it exists.
So in the Write Phase it must still exist and must be same.

- `MOD GREATER 0` - exists
- `MOD LESS <read phase revision + 1>` - it has been last modified by the read phase or an operation before
  - `+1` means convert `<` operator to `<=`, see next commits.

michaljurecko: Same logic is applied in all other cases, ask if a condition is not clear.

internal/pkg/service/common/etcdop/op/atomic.go
```diff
@@ -250,44 +250,49 @@ func (v *AtomicOp[R]) DoWithoutRetry(ctx context.Context, opts ...Option) *TxnRe
 	// Create IF part of the transaction
 	var cmps []etcd.Cmp
 	for _, op := range removeOpsOverlaps(tracker.Operations()) {
-		switch op.Type {
```
michaljurecko, line 253: `tracker` returns recorded operations from the Read Phase.

internal/pkg/service/common/etcdop/op/atomic.go
```diff
@@ -250,44 +250,49 @@ func (v *AtomicOp[R]) DoWithoutRetry(ctx context.Context, opts ...Option) *TxnRe
...
-		case DeleteOp:
+		mustExist := (op.Type == GetOp || op.Type == PutOp) && op.Count > 0
+		mustNotExist := op.Type == DeleteOp || op.Count == 0
+		switch {
+		case mustExist:
+			// IF: 0 < modification version <= Read Phase revision
+			// Key/range exists and has not been modified since the Read Phase.
+			//
```
michaljurecko, line 275: Original implementation didin't covered all cases. Original tests have been keept, but new tests cover all possible combinations of operations.

Changes:


**1. If the key does not exist, it can be written in different ways:**
  - `MOD == 0`, `VERSIO == 0`, `CREATE == 0`.
  - I replaced `VERSION == 0` by `MOD == 0`, because `MOD` is used also in other conditions.

**2. `put` method in Read Phase must generate also confition "key must exists", I have no idea why it was missing there.**
https://github.com/keboola/keboola-as-code/blob/317d484d8d1cfab4c96f91b636e4727c8a3c8c5c/internal/pkg/service/common/etcdop/op/atomic.go#L167-L177

michaljurecko: You can easily review it as a new code, the original tests remained.

internal/pkg/service/common/etcdop/op/atomic_test.go
```diff
@@ -141,8 +141,11 @@ func TestAtomicOp(t *testing.T) {
...
 		{
 			// Read Phase gets a keys prefix, but before the Write Phase, an existing key is deleted in the prefix.
-			Name:           "GetPrefix_DeleteKey",
-			Prepare:        putTwoKeys,
+			Name: "GetPrefix_DeleteKey",
+			Prepare: func(t *testing.T, client etcd.KV) []op.Op {
+				t.Skip("not implemented/impossible: we cannot check that a key was deleted from the prefix")
+				return putTwoKeys(t, client)
+			},
 			ReadPhase:      getPrefix,
```
michaljurecko: But there is a problem, one test-case is not covered by the logic.

I have a prefix "key"` with keys "key/1", "key/2".

I read the prefix in get phase, so write phase something depends on the all prefix values.

But `BrakingChange` removed `"key/2"`.

Write Phase conditions are MET, because:
- There is at least one key in the prefix `MOD > 0`.
- No key has been modified `MOD < read phase revision + 1`.

In the etcd is missing a `COUNT` operation in TXN IF condition, that would help us solve the issue.
https://github.com/Revolution1/etcd3-py/issues/76

You can only check, the prefix is empty, via `MOD == 0`.

internal/pkg/service/common/etcdop/op/tracker.go
```diff
@@ -32,6 +33,8 @@ type TrackedOp struct {
 	Key []byte
 	// RangeEnd, optional, used if the WithPrefix or WithRange option is used
 	RangeEnd []byte
+	// KVs, optional, contains loaded KVs, only for Get operation
+	KVs []*mvccpb.KeyValue
 }
```
michaljurecko, line 38: There is a workaround, to block the Write Phase, if a key is delete from the tracked prefix.

**We can track all loaded keys.**

internal/pkg/service/common/etcdop/op/tracker.go
```diff
@@ -117,8 +120,8 @@ func (v *trackedTxn) Commit() (*etcd.TxnResponse, error) {
 	return r, err
 }
 
-func (v *TrackerKV) track(t TrackedOpType, key, rangeEnd []byte, count int64) {
-	record := TrackedOp{Type: t, Key: key, RangeEnd: rangeEnd, Count: count}
+func (v *TrackerKV) track(t TrackedOpType, key, rangeEnd []byte, count int64, kvs []*mvccpb.KeyValue) {
+	record := TrackedOp{Type: t, Key: key, RangeEnd: rangeEnd, Count: count, KVs: kvs}
```
michaljurecko, line 124: The slice is already in the memory, so the best approach is to store reference to the slice.

Slices are always passed by a reference:
- To be exact, slice header is copied (`firstElementAddr, len, cap`), but underlying array not.

internal/pkg/service/common/etcdop/op/tracker_test.go
```diff
@@ -64,6 +64,17 @@ func TestTracker(t *testing.T) {
 		Commit()
 	assert.NoError(t, err)
 
+	// Get records
+	operations := tracker.Operations()
+	for i := range operations {
+		op := &operations[i]
+		// Clear complex KVs slice, assert count
+		if op.Type == GetOp {
+			assert.Len(t, op.KVs, int(op.Count))
```
michaljurecko, line 73: `KVs` are comples values, so we are not asserting them, we are asserting only the items count.

internal/pkg/service/common/etcdop/op/atomic.go
```diff
@@ -45,7 +46,22 @@ type AtomicOpInterface interface {
...
+//
+// By default, the feature is enabled and checks that each loaded key within the Read Phase, from a prefix, exists in Write Phase.
+// This can be potentially SLOW and generate a lot of IF conditions, if there are a large number of keys in the prefix.
+// Therefore, this feature can be turned off by the method.
+//
+// Modification of a key in the prefix is always detected,
+// this feature is used to detect the deletion of a key from the prefix.
+//
+// See TestAtomicOp:GetPrefix_DeleteKey_SkipPrefixKeysCheck.
+func (v *AtomicOp[R]) SkipPrefixKeysCheck() *AtomicOp[R] {
```
michaljurecko, line 62: We can now check that each loaded key from the prefix exists, but the approach has some limitations.

michaljurecko: By default the feature is enabled, so there is by default a 100% guarantee of atomicity.

internal/pkg/service/common/etcdop/op/atomic.go
```diff
@@ -279,6 +295,20 @@ func (v *AtomicOp[R]) DoWithoutRetry(ctx context.Context, opts ...Option) *TxnRe
...
+					for _, kv := range op.KVs {
+						cmps = append(cmps, etcd.Cmp{
+							Target:      etcdserverpb.Compare_MOD,
+							Result:      etcdserverpb.Compare_GREATER,
+							TargetUnion: &etcdserverpb.Compare_ModRevision{ModRevision: 0},
+							Key:         kv.Key,
+						})
+					}
+				}
+			}
```
michaljurecko, line 311: Implementation.

`op.RangeEnd != nil` it is a Get operation over range/prefix, not a single key get.

internal/pkg/service/common/etcdop/op/atomic_test.go
```diff
@@ -21,11 +21,13 @@ import (
...
-	ReadPhase          func(t *testing.T, client etcd.KV) []op.Op
-	BreakingChange     func(t *testing.T, client etcd.KV) []op.Op
-	ExpectedWritePhase string
+	Name                string
+	SkipPrefixKeysCheck bool
+	Prepare             func(t *testing.T, client etcd.KV) []op.Op
+	ReadPhase           func(t *testing.T, client etcd.KV) []op.Op
+	BreakingChange      func(t *testing.T, client etcd.KV) []op.Op
+	ExpectedWritePhase  string
+	ExpectedlyDontWork  bool
```
michaljurecko, line 30: Test case, added `SkipPrefixKeysCheck` and `ExpectedlyDontWork`.

internal/pkg/service/common/etcdop/op/atomic_test.go
```diff
@@ -141,22 +145,40 @@ func TestAtomicOp(t *testing.T) {
...
+			ExpectedlyDontWork:  true,
+			ExpectedWritePhase: `
+➡️  TXN
+  ➡️  IF:
+  001 ["key/", "key0") MOD GREATER 0
+  002 ["key/", "key0") MOD LESS %d
+  ➡️  THEN:
+  001 ➡️  PUT "foo"
+
+✔️  TXN | succeeded: true
```
michaljurecko, line 181: If the feature is disabled, the collision is not detected.

internal/pkg/service/common/etcdop/op/atomic_test.go
```diff
@@ -141,22 +145,40 @@ func TestAtomicOp(t *testing.T) {
 		},
 		{
 			// Read Phase gets a keys prefix, but before the Write Phase, an existing key is deleted in the prefix.
-			Name: "GetPrefix_DeleteKey",
-			Prepare: func(t *testing.T, client etcd.KV) []op.Op {
-				t.Skip("not implemented/impossible: we cannot check that a key was deleted from the prefix")
```
michaljurecko: Removed skip.

internal/pkg/service/common/etcdop/op/atomic_test.go
```diff
@@ -141,22 +145,40 @@ func TestAtomicOp(t *testing.T) {
...
 ➡️  TXN
   ➡️  IF:
   001 ["key/", "key0") MOD GREATER 0
   002 ["key/", "key0") MOD LESS %d
+  003 "key/1" MOD GREATER 0
+  004 "key/2" MOD GREATER 0
   ➡️  THEN:
   001 ➡️  PUT "foo"
 
 ✔️  TXN | succeeded: false
```
michaljurecko, line 162: By default, the colision is detected by new conditions.

internal/pkg/service/common/etcdop/op/atomic_test.go
```diff
@@ -251,80 +251,82 @@ func TestAtomicOp(t *testing.T) {
 	}
 
 	for _, tc := range cases {
-		tc.Run(t)
+		tc := tc
```
michaljurecko, line 254: Here is a bug :see_no_evil: - missing `tc := tc`.

I added `t.Parallel()` at the work end, so I didn't notice it.

I realized it during next work. The last test-case was started repeatedly, instead of other test cases.

Linter didn't reveal it, because callbacks are wrapped in methods.

The solution is to add `tc := tc` or remove the pointer receiver from `*atomicOpTestCase` methods. 

I did both, and tweaked it a bit.

image

michaljurecko (commented):
I added the last 2 commits here, I found some issues during the next work.

PR #1529, opened 2024-01-20
fix: Remove broken removeOpsOverlaps

michaljurecko:
Part of: https://keboola.atlassian.net/browse/PSGO-357

**Changes:**
- ...

-----------


internal/pkg/service/common/etcdop/op/atomic.go
```diff
@@ -265,7 +265,7 @@ func (v *AtomicOp[R]) DoWithoutRetry(ctx context.Context, opts ...Option) *TxnRe
 
 	// Create IF part of the transaction
 	var cmps []etcd.Cmp
-	for _, op := range removeOpsOverlaps(tracker.Operations()) {
+	for _, op := range tracker.Operations() {
```
michaljurecko, line 268: I found bug in the `removeOpsOverlaps` function, see example:

image

image

-------------------

Function purpose was to remove overlaps in tracked operations from Read Phase.

For example:
- Range `GET [aaa/, xxx0)` contains range  `GET [bbb/, ccc0)`, so `[bbb/, ccc0)` can be ignored, form Write Phase conditions.

In the previous iteration of the Stream API it was used, but now it is not used at all (see etcd snapshots in the next PR).

So I decided to remove this feature at all. It is complex and there is no reason to maintain it.

----------------------

B

internal/pkg/service/common/etcdop/op/util.go
```diff
@@ -1,55 +0,0 @@
...
-	// Start must be same or greater
-	if bytes.Compare(op.Key, parent.Key) == -1 {
-		// Key is smaller
-		return false
-	}
-	// End must be same or smaller
-	if bytes.Compare(op.RangeEnd, parent.RangeEnd) == 1 {
-		// End is greater
-		return false
-	}
```
michaljurecko, line 53: Bug is here, in the `isOpSubsetOf`, and similar also in the `sort` above.

If only one key is loaded, not a range, then `RangeEnd == nil`.
- In this case, we should set `RangeEnd = Key` to fix comparison.

---------------

For example, two keys `A,B`:
image

- // Operation type must be same :heavy_check_mark:  
- // Start (Key) B must be same or greater than A :heavy_check_mark: 
- // End B must be same or smaller than A :heavy_check_mark: - **both are keys, RangeEnd == nil** :lady_beetle: 

-------------

See https://goplay.tools/snippet/xvLLsQEcIDU

PR #1538, opened 2024-01-24
feat: Add missing ConfigPatch structs

michaljurecko:
Jira: https://keboola.atlassian.net/browse/PSGO-387

**Changes:**
- `ConfigPatch` struct were missing in some deeper configuration levels.
   - Fixed, now it is posible to patch each individual configuration field - from the Table Sink definition.
   - I need it to finish Sink Endpoints Definitions.

-----------


internal/pkg/service/buffer/sink/tablesink/storage/compression/config.go
```diff
@@ -15,42 +15,69 @@ const (
...
 		Type: TypeGZIP,
-		GZIP: DefaultGZIPConfig().GZIP,
-		ZSTD: DefaultZSTDConfig().ZSTD,
+		GZIP: NewGZIPConfig().GZIP,
+		ZSTD: NewZSTDConfig().ZSTD,
 	}
 }
 
-func DefaultNoneConfig() Config {
+func NewNoneConfig() Config {
```
michaljurecko, line 74: Unified naming, on other places is already `NewConfig`.

internal/pkg/service/buffer/sink/tablesink/storage/level/local/writer/disksync/config.go
```diff
@@ -39,18 +39,18 @@ type Config struct {
 	// Mode defines sync mode: more durable ModeDisk or faster ModeCache.
 	Mode Mode `json:"mode" configKey:"mode" validate:"required,oneof=disabled disk cache" configUsage:"Sync mode: \"disabled\", \"cache\" or \"disk\"."`
 	// Wait defines whether the operation should wait for sync.
-	Wait bool `json:"wait" configKey:"wait" validate:"excluded_if= Mode disabled" configUsage:"Wait for sync to disk OS cache or to disk hardware, depending on the mode."`
+	Wait bool `json:"wait" configKey:"wait" configUsage:"Wait for sync to disk OS cache or to disk hardware, depending on the mode."`
```
michaljurecko, line 42: In this config struct, additional values were forbidden.

It can cause probles, for example, we will modify global Stream API configruation and a Sink config patch will be no more valid.

internal/pkg/service/buffer/sink/tablesink/storage/compression/config.go
```diff
@@ -15,42 +15,69 @@ const (
...
 	ZSTD *ZSTDConfig `json:"zstd,omitempty" configKey:"zstd" validate:"required_if=Type zstd"`
 }
 
+// ConfigPatch is same as the Config, but with optional/nullable fields.
+// It is part of the definition.TableSink structure to allow modification of the default configuration.
+type ConfigPatch struct {
+	Type *Type            `json:"type,omitempty"`
+	GZIP *GZIPConfigPatch `json:"gzip,omitempty"`
+	ZSTD *ZSTDConfigPatch `json:"zstd,omitempty"`
+}
```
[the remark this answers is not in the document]
michaljurecko: I haven't needed `nil` anywhere yet, ... usually there is some kind of switch, for example `Wait bool`, `Enabled bool`, ... so it is not one field, but switch + N fields.

For patching a `nil` config value (`*string`), you would need a double pointer (`**string`). 
And I think the method with multiples fields is better.

hosekpeter (approved):
LGTM

PR #1543, opened 2024-01-26
feat: Add configpatch.ToKVs - dump config as key-value pairs

michaljurecko:
Jira: https://keboola.atlassian.net/browse/PSGO-388

**Changes:**
- `configmap.Visit` utility visits also `nil` pointers to deep.
-  Added `configpatch.ToKVs` function.

-----------

#### What do we need to achieve?

- There is a global Stream Service configuration, see [here](https://github.com/keboola/keboola-as-code/blob/7aad8567ea8075608724c21e265cbfbd96b25d8b/internal/pkg/service/buffer/config/config_test.go#L21-L39).
- These settings can be customized per Sink - **via the API**. 
  - Some settings cannot be patched - they are not in the `ConfigPatch` structure. 
  - Some settings can only be changed by the superadmin - `protected:true` tag.
  - And other: a normal user can change them.

In this PR, I want to achieve that we can present the status of these settings **via the API**:
- Default value.
- Actual value.
- Is the key overwritten by in the Sink definition? Or is the default value used?
- Is the key protected? Or can a normal user modify it?
- Plus, list of validation rules, maybe it will be used dynamically in the UI.

See dump of the Stream API config, in the new test bellow:
https://github.com/keboola/keboola-as-code/blob/1bceaecd1d8572a9cc6e0d544682c4db189e6024/internal/pkg/service/buffer/sink/tablesink/config_test.go#L37-L93

### DIFF
--- a/internal/pkg/service/buffer/config/config_test.go
+++ b/internal/pkg/service/buffer/config/config_test.go
@@ -77,7 +77,7 @@ sink:
             volumeAssignment:
                 # Volumes count simultaneously utilized per sink. Validation rules: required,min=1,max=100
                 count: 1
-                # List of preferred volume types, start with the most preferred. Validation rules: min=1
+                # List of preferred volume types, start with the most preferred. Validation rules: required,min=1
                 preferredTypes:
                     - default
             volumeRegistration:
@@ -110,7 +110,7 @@ sink:
                     wait: true
                     # Minimal interval between syncs. Validation rules: min=0,maxDuration=2s,required_if=Mode disk,required_if=Mode cache
                     checkInterval: 5ms
-                    # Written records count to trigger sync. Validation rules: max=1000000,required_if=Mode disk,required_if=Mode cache
+                    # Written records count to trigger sync. Validation rules: min=0,max=1000000,required_if=Mode disk,required_if=Mode cache
                     countTrigger: 500
                     # Written size to trigger sync. Validation rules: maxBytes=100MB,required_if=Mode disk,required_if=Mode cache
                     bytesTrigger: 1MB
--- a/internal/pkg/service/buffer/sink/tablesink/config_test.go
+++ b/internal/pkg/service/buffer/sink/tablesink/config_test.go
@@ -1,0 +1,249 @@
+package tablesink_test
+
+import (
+	"strings"
+	"testing"
+
+	"github.com/c2h5oh/datasize"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	"github.com/keboola/keboola-as-code/internal/pkg/encoding/json"
+	"github.com/keboola/keboola-as-code/internal/pkg/service/buffer/sink/tablesink"
+	"github.com/keboola/keboola-as-code/internal/pkg/service/buffer/sink/tablesink/storage"
+	"github.com/keboola/keboola-as-code/internal/pkg/service/buffer/sink/tablesink/storage/level/local"
+	"github.com/keboola/keboola-as-code/internal/pkg/service/buffer/sink/tablesink/storage/level/local/writer/diskalloc"
+	"github.com/keboola/keboola-as-code/internal/pkg/service/buffer/sink/tablesink/storage/test"
+	"github.com/keboola/keboola-as-code/internal/pkg/service/common/configpatch"
+)
+
+func TestConfig_ToKVs(t *testing.T) {
+	t.Parallel()
+
+	kvs, err := configpatch.DumpKVs(
+		tablesink.NewConfig(),
+		tablesink.ConfigPatch{
+			Storage: &storage.ConfigPatch{
+				Local: &local.ConfigPatch{
+					DiskAllocation: &diskalloc.ConfigPatch{
+						Size: test.Ptr(456 * datasize.MB),
+					},
+				},
+			},
+		},
+	)
+	require.NoError(t, err)
+
+	assert.Equal(t, strings.TrimSpace(`
+[
+  {
+    "key": "storage.local.compression.gzip.blockSize",
+    "value": "256KB",
+    "defaultValue": "256KB",
+    "overwritten": false,
+    "protected": false,
+    "validation": "required,minBytes=16kB,maxBytes=100MB"
+  },
+  {
+    "key": "storage.local.compression.gzip.concurrency",
+    "value": 0,
+    "defaultValue": 0,
+    "overwritten": false,
+    "protected": false
+  },
+  {
+    "key": "storage.local.compression.gzip.implementation",
+    "value": "parallel",
+    "defaultValue": "parallel",
+    "overwritten": false,
+    "protected": false,
+    "validation": "required,oneof=standard fast parallel"
+  },
+  {
+    "key": "storage.local.compression.gzip.level",
+    "value": 1,
+    "defaultValue": 1,
+    "overwritten": false,
+    "protected": false,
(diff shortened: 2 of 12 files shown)


internal/pkg/service/common/configmap/flags.go
```diff
@@ -74,7 +74,7 @@ func StructToFlags(fs *pflag.FlagSet, v any, outFlagToField map[string]orderedma
 			case bool:
 				fs.Bool(flagName, v, usage)
 			case string:
-				if vc.Value.IsZero() {
+				if !vc.Value.IsValid() || vc.Value.IsZero() {
 					// Don't set the default Value, if the original Value is empty.
```
michaljurecko, line 78: Here, if a parent pointer was `nil`, the child value (that doesn't exists) is marked as invalid by the reflection.

internal/pkg/service/common/configmap/flags.go
```diff
@@ -103,8 +103,8 @@ func StructToFlags(fs *pflag.FlagSet, v any, outFlagToField map[string]orderedma
 	})
 }
 
-func mapAndFilterField() func(field reflect.StructField) (fieldName string, ok bool) {
-	return func(field reflect.StructField) (fieldName string, ok bool) {
+func mapAndFilterField() func(field reflect.StructField, path orderedmap.Path) (fieldName string, ok bool) {
+	return func(field reflect.StructField, path orderedmap.Path) (fieldName string, ok bool) {
 		// Field must have tag
```
michaljurecko, line 108: See `OnField` change bellow.

internal/pkg/service/common/configmap/visit.go
```diff
@@ -10,7 +10,10 @@ import (
 )
 
 type VisitConfig struct {
-	OnField func(field reflect.StructField) (fieldName string, ok bool)
+	// OnField maps field to a custom field name, for example from a tag.
+	// If ok == false, then the field is ignored.
+	OnField func(field reflect.StructField, path orderedmap.Path) (fieldName string, ok bool)
```
michaljurecko, line 15: Added `path` argument, needed by the code in the next commits.

internal/pkg/service/common/configmap/visit.go
```diff
@@ -38,6 +41,7 @@ type VisitContext struct {
 	Validate string
 }
 
+// Visit each nested structure field.
 func Visit(value reflect.Value, cfg VisitConfig) error {
 	vc := &VisitContext{}
```
michaljurecko, line 46: I refactored the `configmap.Visit` function that already visits configuration structure.

But I need to visit also `nil` pointers, in the `ConfigPatch` structures, to get the full list of all keys.
See test in the next commit.

-----------

Original tests are unmodified. 

In the tests, all logic is covered, missing coverage is for different base types: `int8`, `int16`, ... and some rare errors.

image

-----------------

The code contains same blocks, but they were reorganized, and handling for `nil` and `invalid` cases have been added.

internal/pkg/service/common/configmap/visit_test.go
```diff
@@ -0,0 +1,51 @@
...
+		"Nested1",
+		"Nested1.Ignored",
+		"Nested1.Foo",
+		"Nested1.Bar",
+		// Visit function visits also empty pointers, see Nested2 field
+		"Nested2",
+		"Nested2.Ignored",
+		"Nested2.Foo",
+		"Nested2.Bar",
+	}, keys)
```
michaljurecko, line 51: I added the new unit test, `Nested2 *Nested` visitation is now covered.

internal/pkg/service/common/configpatch/dump.go
```diff
@@ -0,0 +1,174 @@
...
+	Protected bool `json:"protected"`
+	// Validation contains validation rules of the field, if any.
+	Validation string `json:"validation,omitempty"`
+}
+
+// DumpKVs generates key-value pairs from a configuration structure and a patch structure.
+// Only keys found in both, configuration and patch structure, are processed.
+// The structure is flattened, keys are joined with a dot "." separator.
+// Each key-value pair contains information whether the value was overwritten from the patch or not.
+func DumpKVs(configStruct, patchStruct any, opts ...Option) ([]DumpKV, error) {
```
michaljurecko, line 35: This new function generates key-value pairs, see real world example, in the PR description.

internal/pkg/service/common/configpatch/dump_test.go
```diff
@@ -0,0 +1,288 @@
...
+type ConfigPatchInvalid struct {
+	Key1 string  `json:"foo1"`
+	Key2 *string `json:"foo2"`
+	Key8 *string `json:"foo8"`
+	Key9 *int    `json:"foo9"`
+}
+
+func TestDumpKVs_EmptyPatch(t *testing.T) {
+	t.Parallel()
+
```
michaljurecko, line 55: Here are the boring but necessary tests for `confipatch.ToKVs` function.

The code is fully covered, except for one panic.
image

internal/pkg/service/buffer/sink/tablesink/config_test.go
```diff
@@ -0,0 +1,249 @@
...
+
+	assert.Equal(t, strings.TrimSpace(`
+[
+  {
+    "key": "storage.local.compression.gzip.blockSize",
+    "value": "256KB",
+    "defaultValue": "256KB",
+    "overwritten": false,
+    "protected": false,
+    "validation": "required,minBytes=16kB,maxBytes=100MB"
```
michaljurecko, line 45: Test for real Stream API `Config`/`ConfigPatch`.

internal/pkg/service/buffer/sink/tablesink/config_test.go
```diff
@@ -0,0 +1,249 @@
...
+    "protected": false
+  },
+  {
+    "key": "storage.local.diskAllocation.size",
+    "value": "456MB",
+    "defaultValue": "100MB",
+    "overwritten": true,
+    "protected": false,
+    "validation": "required"
+  },
```
michaljurecko, line 115: This key is patched.

`protected` tags are not set yet.

internal/pkg/service/buffer/config/config_test.go
```diff
@@ -77,7 +77,7 @@ sink:
             volumeAssignment:
                 # Volumes count simultaneously utilized per sink. Validation rules: required,min=1,max=100
                 count: 1
-                # List of preferred volume types, start with the most preferred. Validation rules: min=1
+                # List of preferred volume types, start with the most preferred. Validation rules: required,min=1
```
michaljurecko, line 80: In the last commit, a slighty updated some validarion rules, so that they are not confusing.

`min=1` is sufficient validation, but I prefer to mark all required values with the `required` rule.

internal/pkg/service/common/configmap/flags.go
```diff
@@ -103,8 +103,8 @@ func StructToFlags(fs *pflag.FlagSet, v any, outFlagToField map[string]orderedma
 	})
 }
 
-func mapAndFilterField() func(field reflect.StructField) (fieldName string, ok bool) {
-	return func(field reflect.StructField) (fieldName string, ok bool) {
+func mapAndFilterField() func(field reflect.StructField, path orderedmap.Path) (fieldName string, ok bool) {
+	return func(field reflect.StructField, path orderedmap.Path) (fieldName string, ok bool) {
```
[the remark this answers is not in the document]
michaljurecko: :man_shrugging: we can, I'll put it in the next PR

PR #1553, opened 2024-02-01
feat: Make the delete instance endpoint async

hosekpeter:
Jira: https://keboola.atlassian.net/browse/PSGO-325

**Changes:**
- delete instance endpoint is async

-----------


michaljurecko:
> Code looks okay, just one question... is it okay to change the semantics of an existing endpoint to async or should it be a new endpoint to avoid BC break? With sync endpoint code 200 means the operation succeeded, for async one you just know the request was accepted but not the result. So this technically is a BC break, just unsure whether or not it matters in our case.

Yesm it is BC and it is ok, ... it is now used only by our UI, there is some note about it, in Jira (about deployment).

test/templates/api/projects/instance-delete/004-delete-instance-ok/expected-response.json
```diff
@@ -1 +1,8 @@
-{}
+{
+  "id": "template.delete/%s",
+  "type": "template.delete",
+  "url": "https://templates.keboola.local/v1/tasks/template.delete/%s",
+  "status": "processing",
+  "isFinished": false,
+  "createdAt": "%s"
+}
```
michaljurecko, line 8: You should also check the task result, in both test.

See example in the Jira.

michaljurecko (approved):
LGTM

PR #1556, opened 2024-02-05
fix: Invalid link

jachym-tousek-keboola:
Pipelines are [failing](https://github.com/keboola/keboola-as-code/actions/runs/7784399276/job/21224868229) because one of the links used in documentation is no longer valid. Apparently PHPUnit author took down the documentation for the old version.


hosekpeter (approved):
LGTM

PR #1562, opened 2024-02-07
fix: Remove old Stream API code

michaljurecko:
Jira: https://keboola.atlassian.net/browse/PSGO-288

**Goal:**
- Make cleanup, so unit tests will work and we can merge the `RFC` to the `main` branch (part 1).

**Changes:**
- Removed `stream/worker` package + E2E tests:
  - In the previous implementation, there were 2 compontents: `api` and `worker`.
  - In the new implementation there are 5 components: 
    - `api`, `storage-writer`, `storage-reader`, `http-source`, `coordinator`.
    - They can be deployed as one pod, or multiple pods, according to the scaling requirements.
    - Communication is handled via `etcd`.
- Removed `stream/watcher` package
  - Responsibility divided into several packages, will be finished in a next PR.
- Removed `stream/condition`
  - Full rewriting needed
- Removed `test/stream/api` E2E tests, except `base`.
  - Full rewriting needed, endpoints significantly differ.
-----------


hosekpeter (approved):
LGTM

PR #1565, opened 2024-02-07
feat: Improve logs in the storage layer

michaljurecko:

**Changes:**
- Modified logger method: `With(args ...any)` to `With(attrs ...attribute.KeyValue)`
- Added log attributes to the storage layer - volume,writer,reader

-----------

### DIFF
--- a/internal/pkg/log/log.go
+++ b/internal/pkg/log/log.go
@@ -5,6 +5,7 @@ import (
 	"io"
 
 	"github.com/stretchr/testify/assert"
+	"go.opentelemetry.io/otel/attribute"
 	"go.uber.org/zap/zapcore"
 )
 
@@ -17,7 +18,7 @@ const (
 
 type Logger interface {
 	contextLogger
-	withFields
+	withAttributes
 }
 
 type loggerWithZapCore interface {
@@ -63,7 +64,7 @@ type contextLogger interface {
 	Sync() error
 }
 
-type withFields interface {
-	With(args ...any) Logger
+type withAttributes interface {
+	With(attrs ...attribute.KeyValue) Logger
 	WithComponent(component string) Logger
 }
--- a/internal/pkg/log/memory_test.go
+++ b/internal/pkg/log/memory_test.go
@@ -4,6 +4,8 @@ package log
 import (
 	"context"
 	"testing"
+
+	"go.opentelemetry.io/otel/attribute"
 )
 
 func TestMemoryLogger(t *testing.T) {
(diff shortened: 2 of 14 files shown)


internal/pkg/log/log.go
```diff
@@ -63,7 +64,7 @@ type contextLogger interface {
 	Sync() error
 }
 
-type withFields interface {
-	With(args ...any) Logger
+type withAttributes interface {
+	With(attrs ...attribute.KeyValue) Logger
 	WithComponent(component string) Logger
 }
```
michaljurecko, line 70: I prefer a one ways how to work with attributes in logger, tracer and `ctxattrs`, it was discussed, but no final decision was made.

If we specify the data type, we can refactor it more easily in the future to a different logger than Zap.

internal/pkg/log/memory_test.go
```diff
@@ -12,7 +14,7 @@ func TestMemoryLogger(t *testing.T) {
 	mem := NewMemoryLogger()
 	mem.Debug(context.Background(), `Debug message.`)
 	mem.Info(context.Background(), `Info message.`)
-	memWithCtx := mem.With("key1", "value1", "key2", "value2")
+	memWithCtx := mem.With(attribute.String("key1", "value1"), attribute.String("key2", "value2"))
```
michaljurecko, line 17: Example usage.

internal/pkg/service/stream/sink/tablesink/storage/level/local/reader/volume/reader.go
```diff
@@ -26,7 +28,14 @@ func (v *Volume) NewReaderFor(slice *storage.Slice) (r reader.Reader, err error)
 	}
 
 	// Setup logger
-	logger := v.logger
+	logger := v.logger.With(
+		attribute.String("projectId", slice.ProjectID.String()),
```
michaljurecko, line 32: I update the code in the storage layer, log messages now contain attributes.

See the next commit with tests.

There is a long-lived logger, and I want to log the attributes with all operations, regardless of context, therefore I am not using `ctxattrs` (they are better for `userID` and similar - request based - attributes).

internal/pkg/service/stream/sink/tablesink/storage/level/local/reader/volume/reader_test.go
```diff
@@ -37,6 +37,15 @@ func TestVolume_NewReaderFor_Ok(t *testing.T) {
 
 	assert.NoError(t, w.Close())
 	assert.Len(t, tc.Volume.readers, 0)
+
+	// Check logs
+	tc.AssertLogs(`
+{"level":"info","message":"opening volume"}
+{"level":"info","message":"opened volume"}
+{"level":"debug","message":"opened file","volume.id":"my-volume","file.path":"%s","projectId":"123","branchId":"456","sourceId":"my-source","sinkId":"my-sink","fileId":"2000-01-01T19:00:00.000Z","sliceId":"2000-01-01T20:00:00.000Z"}
+{"level":"debug","message":"closing chain"}
```
michaljurecko, line 46: Petr: not all attributes must be specified in the logs asserts, therefore I only modified some tests.

hosekpeter (approved):
LGTM

internal/pkg/log/log.go
hosekpeter: refactor to `log/slog`?

michaljurecko: yes, we discussed the slog with Jachym, but we didn't want to complicate the logs refactoring now

PR #1569, opened 2024-02-09
feat: Improve interface of the resistant etcd session

michaljurecko:

**Changes:**
- Improved interface of the `etcdop.Session`:
  - Multiple `OnSession` callbacks.
  - `Session` and `WaitForSession` getters.
  - Default configuration values for TTL seconds and grant timeout.

-----------

**Intro**

There is a "resistent session" in the `etcdop` package.

The `Session` wraps etcd low-level `*concurrency.Session` - it provides: `LeaseID`, distributed locks, ... 

The problem with  etcd low-level `*concurrency.Session` is, that if there is an network outage, if TTL seconds expire, the session is canceled.

But in our apps: we need a retry and the session recreation, so we created "resistant session" implementation.

In this PR, I improved the package interface, so it can be used better.

----------------

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -43,6 +43,7 @@ require (
 	github.com/relvacode/iso8601 v1.3.0
 	github.com/santhosh-tekuri/jsonschema/v5 v5.3.0
 	github.com/schollz/progressbar/v3 v3.13.1
+	github.com/shopify/toxiproxy v2.1.4+incompatible
 	github.com/sirupsen/logrus v1.9.3
 	github.com/spf13/afero v1.9.5
 	github.com/spf13/cast v1.5.1
@@ -146,6 +147,7 @@ require (
 	github.com/DataDog/go-libddwaf/v2 v2.2.3 // indirect
 	github.com/DataDog/go-tuf v1.0.2-0.5.2 // indirect
 	github.com/MichaelMure/go-term-text v0.3.1 // indirect
+	github.com/Shopify/toxiproxy v2.1.4+incompatible // indirect
 	github.com/agext/levenshtein v1.2.1 // indirect
 	github.com/alecthomas/chroma v0.10.0 // indirect
 	github.com/andybalholm/brotli v1.0.6 // indirect
@@ -194,6 +196,7 @@ require (
 	github.com/google/wire v0.5.0 // indirect
 	github.com/googleapis/enterprise-certificate-proxy v0.3.2 // indirect
 	github.com/googleapis/gax-go/v2 v2.12.0 // indirect
+	github.com/gorilla/mux v1.8.0 // indirect
 	github.com/grpc-ecosystem/go-grpc-middleware v1.3.0 // indirect
 	github.com/grpc-ecosystem/go-grpc-prometheus v1.2.0 // indirect
 	github.com/grpc-ecosystem/grpc-gateway v1.16.0 // indirect
@@ -246,5 +249,6 @@ require (
 	google.golang.org/genproto/googleapis/api v0.0.0-20231212172506-995d672761c0 // indirect
 	google.golang.org/genproto/googleapis/rpc v0.0.0-20240108191215-35c7eff3a6b1 // indirect
 	gopkg.in/natefinch/lumberjack.v2 v2.0.0 // indirect
+	gopkg.in/tomb.v1 v1.0.0-20141024135613-dd632973f1e7 // indirect
 	inet.af/netaddr v0.0.0-20230525184311-b8eac61e914a // indirect
 )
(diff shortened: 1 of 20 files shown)


go.mod
```diff
@@ -43,6 +43,7 @@ require (
 	github.com/relvacode/iso8601 v1.3.0
 	github.com/santhosh-tekuri/jsonschema/v5 v5.3.0
 	github.com/schollz/progressbar/v3 v3.13.1
+	github.com/shopify/toxiproxy v2.1.4+incompatible
```
michaljurecko, line 46: For tests.

internal/pkg/service/common/etcdop/session_test.go
```diff
@@ -2,66 +2,159 @@ package etcdop
...
-	client := cluster.Client(0)
+	// Get credentials
+	etcdCfg := etcdhelper.TmpNamespace(t)
+
+	// Setup proxy to drop etcd connection
+	proxy := toxiproxy.NewProxy()
+	proxy.Name = "etcd-bridge"
+	proxy.Upstream = etcdCfg.Endpoint
+	require.NoError(t, proxy.Start())
+	defer proxy.Stop()
```
michaljurecko, line 34: The etcd `integration` package is not stable (its interace can be modified anytime), it is used internally for etcd testing.

So I replaced it with the `toxiproxy`, it is a switchable proxy between test client and real `etcd` database.
It is simpler and better match real world.

internal/pkg/service/common/etcdop/session_test.go
```diff
@@ -2,66 +2,159 @@ package etcdop
...
+	// Drop connection
+	proxy.Stop()
+	assert.Eventually(t, func() bool {
+		return logger.CompareJSONMessages(`
+{"level":"info","message":"etcd session canceled"}
+{"level":"info","message":"creating etcd session"}
+{"level":"info","message":"cannot create etcd session: context deadline exceeded"}
+{"level":"info","message":"waiting %s before the retry"}
+`) == nil
+	}, 15*time.Second, 100*time.Millisecond)
```
michaljurecko, line 72: Session is canceled after the TTL seconds.

internal/pkg/utils/etcdhelper/fortest.go
```diff
@@ -107,7 +107,7 @@ func clientForTest(t testOrBenchmark, ctx context.Context, cfg etcdclient.Config
 				BaseDelay:  100 * time.Millisecond,
 				Multiplier: 1.5,
 				Jitter:     0.2,
-				MaxDelay:   15 * time.Second,
+				MaxDelay:   5 * time.Second,
```
michaljurecko, line 110: To speedup failure of a broken test.

internal/pkg/service/common/task/orchestrator/orchestrator_test.go
```diff
@@ -113,18 +113,18 @@ func TestOrchestrator(t *testing.T) {
...
-{"level":"info","message":"assigned \"1000/my-receiver/some.task/ResourceID\"","component":"orchestrator","task":"some.task"}
-{"level":"info","message":"started task","component":"task","task":"1000/my-receiver/some.task/ResourceID/%s"}
-{"level":"debug","message":"lock acquired \"runtime/lock/task/1000/my-receiver/ResourceID\"","component":"task","task":"1000/my-receiver/some.task/ResourceID/%s"}
-{"level":"info","message":"message from the task","component":"task","task":"1000/my-receiver/some.task/ResourceID/%s"}
-{"level":"info","message":"task succeeded (%s): ResourceID","component":"task","task":"1000/my-receiver/some.task/ResourceID/%s"}
-{"level":"debug","message":"lock released \"runtime/lock/task/1000/my-receiver/ResourceID\"","component":"task","task":"1000/my-receiver/some.task/ResourceID/%s"}
+{"level":"info","message":"assigned \"1000/my-prefix/some.task/ResourceID\"","component":"orchestrator","task":"some.task"}
+{"level":"info","message":"started task","component":"task","task":"1000/my-prefix/some.task/ResourceID/%s"}
+{"level":"debug","message":"lock acquired \"runtime/lock/task/custom-lock\"","component":"task","task":"1000/my-prefix/some.task/ResourceID/%s"}
```
michaljurecko, line 118: There were some outdated tests. I didn't noticed the problem, after the update from the `main` branch.

internal/pkg/service/common/etcdop/session.go
```diff
@@ -10,112 +10,289 @@ import (
...
+
+// NewMutex returns *concurrency.Mutex that implements the sync Locker interface with etcd.
+// IsOwner().
+func (s *Session) NewMutex(name string) (*concurrency.Mutex, error) {
+	if session, err := s.Session(); err == nil {
+		return concurrency.NewMutex(session, name), nil
+	} else {
+		return nil, err
+	}
+}
```
michaljurecko, line 214: New method, a shortcut.

michaljurecko:
@jachym-tousek-keboola, we will how to log the duration (minor think), but you can review the rest.

PR #1570, opened 2024-02-09
tests: Minor fixes: ignore new etcd metrics, fix ctx, fix iterator

michaljurecko:

**Changes:**
- Fixed failing unit tests:
  - Ignored new `rpc.*` etcd client metrics.
  - Fixed bat ctx in the `task` pkg.
  - Fixed iterator tests - removed new dynamic values.

-----------

### DIFF
--- a/internal/pkg/service/common/etcdop/iterator/iterator_test.go
+++ b/internal/pkg/service/common/etcdop/iterator/iterator_test.go
@@ -489,12 +489,18 @@ func TestIterator_ForEach(t *testing.T) {
 	// Run op
 	assert.NoError(t, getAllOp.Do(ctx).Err())
 
+	// Clear loaded KVs before assert
+	operations := tracker.Operations()
+	for i := range operations {
+		operations[i].KVs = nil
+	}
+
 	// All requests can be tracked by the TrackerKV
 	assert.Equal(t, []op.TrackedOp{
 		{Type: op.GetOp, Key: []byte("some/prefix/"), RangeEnd: []byte("some/prefix0"), Count: 5},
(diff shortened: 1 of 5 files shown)


internal/pkg/service/common/etcdop/iterator/iterator_test.go
```diff
@@ -489,12 +489,18 @@ func TestIterator_ForEach(t *testing.T) {
 	// Run op
 	assert.NoError(t, getAllOp.Do(ctx).Err())
 
+	// Clear loaded KVs before assert
+	operations := tracker.Operations()
+	for i := range operations {
+		operations[i].KVs = nil
+	}
+
```
michaljurecko, line 497: There were some update of the `etcdop.Tracer`, so it now traces also loaded KVs, so we need clear them, in some tests.

internal/pkg/service/common/task/node_test.go
```diff
@@ -901,9 +901,13 @@ func newTestTelemetryWithFilter(t *testing.T) telemetry.ForTest {
...
 		NewForTest(t).
-		SetSpanFilter(func(ctx context.Context, spanName string, opts ...trace.SpanStartOption) bool {
+		AddSpanFilter(func(ctx context.Context, spanName string, opts ...trace.SpanStartOption) bool {
 			// Ignore etcd spans
 			return !strings.HasPrefix(spanName, "etcd")
+		}).
+		AddMetricFilter(func(metric metricdata.Metrics) bool {
+			// Ignore etcd metrics
+			return !strings.HasPrefix(metric.Name, "rpc.")
 		})
```
michaljurecko, line 911: There some new etcd `rpc.*` metrics after the latest update, so we need a way how to ignore them.

michaljurecko: So I added `AddMetricFilter` helper.

PR #1571, opened 2024-02-09
feat: Misconfigured apps

jachym-tousek-keboola:
Jira: https://keboola.atlassian.net/browse/PSGO-408

**Changes:**
- Changed behavior of misconfigured apps to error 403. We can customize the page later.
- Fixed code review comments from previous PRs.

-----------

### DIFF
--- a/internal/pkg/service/appproxy/http/router_test.go
+++ b/internal/pkg/service/appproxy/http/router_test.go
@@ -10,6 +10,7 @@ import (
 	"net/http"
 	"net/http/httptest"
 	"net/url"
+	"os"
 	"strings"
 	"testing"
 	"time"
@@ -26,6 +27,7 @@ import (
 	"github.com/keboola/keboola-as-code/internal/pkg/service/appproxy/config"
 	proxyDependencies "github.com/keboola/keboola-as-code/internal/pkg/service/appproxy/dependencies"
 	mockoidcCustom "github.com/keboola/keboola-as-code/internal/pkg/service/appproxy/mockoidc"
+	"github.com/keboola/keboola-as-code/internal/pkg/utils/errors"
 )
 
 type testCase struct {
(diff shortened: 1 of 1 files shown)


internal/pkg/service/appproxy/http/router.go
```diff
@@ -95,65 +96,87 @@ func (r *Router) CreateHandler() http.Handler {
 	return mux
 }
 
-func (r *Router) createDataAppHandler(app DataApp) (http.Handler, error) {
+func (r *Router) createForbiddenHandler() http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, req *http.Request) {
+		w.WriteHeader(http.StatusForbidden)
+		fmt.Fprintln(w, "Application has misconfigured OAuth2 provider.")
+	})
+}
```
michaljurecko, line 104: `forbiddenHandler` -> `configErrorHandler`?  forbidden sounds to me like it's the user's fault

internal/pkg/service/appproxy/http/router_test.go
```diff
@@ -63,6 +63,20 @@ func TestAppProxyRouter(t *testing.T) {
 				assert.Equal(t, `Application "unknown" not found.`, string(body))
 			},
 		},
+		{
+			name: "broken-app",
+			run: func(t *testing.T, client *http.Client, m []*mockoidc.MockOIDC, appServer *appServer) {
+				// Request to unknown app
```
michaljurecko, line 69: `// Request to unknown app` -> `// Request to broken app`.

michaljurecko (approved):
LGTM

internal/pkg/service/appproxy/http/router.go
```diff
@@ -95,65 +96,87 @@ func (r *Router) CreateHandler() http.Handler {
...
+	return r.protectedAppHandler(ctx, app)
 }
 
-func (r *Router) publicAppHandler(app DataApp) (http.Handler, error) {
+func (r *Router) publicAppHandler(ctx context.Context, app DataApp) http.Handler {
 	target, err := url.Parse("http://" + app.UpstreamHost)
 	if err != nil {
-		return nil, errors.Errorf(`cannot parse upstream url "%s" for app %s: %w`, app.UpstreamHost, app.ID, err)
+		r.logger.Errorf(ctx, `cannot parse upstream url "%s" for app <app> %s: %w`, app.UpstreamHost, app.Name, err.Error())
+		return r.configErrorHandler
```
michaljurecko, line 119: In future, we will need probably to include error details in the error page (at least an `exceptionId`), so the handler won't be same.

michaljurecko (approved):
LGTM

internal/pkg/service/appproxy/http/router.go
michaljurecko: See `httpserver/error.go` and `WithExceptionID` interfaces for example.

Error in the `go-client` implements `WithExceptionID` interface.

image

michaljurecko: So you should:
- Generate an exeception ID for the error.
- Log the error, with details.
- In the handler, always return the same `exceptionID`.
- So we (support) can find it in the logs, and get error details.
- Logged message and message for the user may differ.

hosekpeter (approved):

PR #1572, opened 2024-02-09
fix: Improve cookies configuration

jachym-tousek-keboola:
Jira: https://keboola.atlassian.net/browse/PSGO-414

**Changes:**
- Cookies are now using SameSite=Strict (see https://zdrojak.cz/clanky/co-jsou-samesite-cookie-a-proc-je-potrebujeme/)
- Added assertion for `_oauth2_provider` cookie.
- Added assertion that 403 page from previous PR does not set `_oauth2_provider` cookie.
- HTTPOnly and Secure are both default and were already tested for so no changes there.
- Refactored tests to use proper cookies API instead of raw headers.

-----------


michaljurecko (approved):
LGTM

hosekpeter (approved):
LGTM

PR #1576, opened 2024-02-12
refactor: Move private test helpers to the "test", to be reused

michaljurecko:

**Changes:**
- Updated code, so some private helpers can now be reused in more tests.

-----------


internal/pkg/service/stream/sink/tablesink/storage/test/slice.go
```diff
@@ -1,20 +1,34 @@
...
+	"github.com/keboola/keboola-as-code/internal/pkg/utils/errors"
 )
 
+// sliceRepository interface to prevent package import cycle.
+type sliceRepository interface {
+	Close(now time.Time, fileVolumeKey storage.FileVolumeKey) *op.AtomicOp[op.NoResult]
+	Get(k storage.SliceKey) op.WithResult[storage.Slice]
+	StateTransition(now time.Time, sliceKey storage.SliceKey, from, to storage.SliceState) *op.AtomicOp[storage.Slice]
+}
+
```
michaljurecko, line 31: I added these interface to prevent import cycles.

I want to have one `../storage/test` pkg with all helpers, not `testslice`,`testfile`, .... it would complicate imports/code.

michaljurecko: Go is very strict about this.

At first, the reason was that they wanted to simplify the implementation, 
then it turned out that it is actually super useful.

https://github.com/golang/go/issues/30247

PR #1579, opened 2024-02-12
feat: Add health-check endpoint for oauth proxy

jachym-tousek-keboola:
Jira: [keboola.atlassian.net/browse/PSGO-425](https://keboola.atlassian.net/browse/PSGO-425)

**Changes:**
- Added `/health-check` endpoint for oauth proxy.

-----------

### DIFF
--- a/internal/pkg/service/appproxy/http/router.go
+++ b/internal/pkg/service/appproxy/http/router.go
@@ -80,6 +80,11 @@ func (r *Router) CreateHandler() http.Handler {
 	mux.HandleFunc("/", func(w http.ResponseWriter, req *http.Request) {
 		appID, ok := parseAppID(req.Host)
 		if !ok {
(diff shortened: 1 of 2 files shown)


internal/pkg/service/appproxy/http/router.go
```diff
@@ -80,6 +80,11 @@ func (r *Router) CreateHandler() http.Handler {
 	mux.HandleFunc("/", func(w http.ResponseWriter, req *http.Request) {
 		appID, ok := parseAppID(req.Host)
```
michaljurecko, line 81: Host may be: `proxy.data-apps.svc.cluster.local`, you should improve the `parseAppID` function, and add some static expected host suffix to the proxy service configuration struct.

michaljurecko (approved):

hosekpeter (approved):

PR #1586, opened 2024-02-15
refactor: Improve pkg and config structure

michaljurecko:

**Changes:**
- There are many configuration options in the Stream service.
- The top level configuration is composed from low-level configurations.
- Each small part can be tested separately, as well as the whole service.
- This PR fixes hierarchy PACKAGES and CONFIGURATION structure to match application logic.
- This also removes the problems with cyclic dependencies that I ran into in other work.

-----------

Full config example:

https://github.com/keboola/keboola-as-code/blob/da629d0750c618be41706c19c078464f2f3a7ec1/internal/pkg/service/stream/config/config_test.go#L14-L155


internal/pkg/service/stream/cleanup/node_test.go
```diff
@@ -13,7 +13,7 @@ import (
 	"github.com/keboola/keboola-as-code/internal/pkg/service/stream/cleanup"
 	"github.com/keboola/keboola-as-code/internal/pkg/service/stream/config"
 	bufferDependencies "github.com/keboola/keboola-as-code/internal/pkg/service/stream/dependencies"
-	"github.com/keboola/keboola-as-code/internal/pkg/service/stream/sink/tablesink/statistics"
+	"github.com/keboola/keboola-as-code/internal/pkg/service/stream/storage/statistics"
```
michaljurecko, line 16: The `storage` layer can be used also for other sink types, in the future.

internal/pkg/service/stream/storage/repository/file_repo_state_test.go
```diff
@@ -86,7 +86,7 @@ func TestFileRepository_StateTransition(t *testing.T) {
 
 	// Create (the first file Rotate)
 	// -----------------------------------------------------------------------------------------------------------------
-	var file storage.File
+	var file model.File
```
michaljurecko, line 89: Storage entitites have been moved to the `model` package.

hosekpeter (approved):
LGTM

PR #1589, opened 2024-02-15
fix: Use usage and shorthand for configkey:",squash" field

michaljurecko:
Jira: https://keboola.atlassian.net/browse/PSGO-440

**Changes:**
- Fixed bug in the `configmap` package: there were no usage and shorthand flag metadata for `configmap.Value` type.

-----------

### DIFF
--- a/internal/pkg/service/common/configmap/flags.go
+++ b/internal/pkg/service/common/configmap/flags.go
@@ -47,8 +47,8 @@ func GenerateFlags(fs *pflag.FlagSet, v any) error {
 				return nil
 			}
 
-			shorthand := vc.StructField.Tag.Get(configShorthandTag)
-			usage := vc.StructField.Tag.Get(configUsageTag)
+			shorthand := vc.Shorthand
+			usage := vc.Usage
 
 			switch v := vc.PrimitiveValue.Interface().(type) {
 			case int:
--- a/internal/pkg/service/common/configmap/flags.go
+++ b/internal/pkg/service/common/configmap/flags.go
@@ -10,6 +10,12 @@ import (
 	"github.com/keboola/keboola-as-code/internal/pkg/utils/errors"
 )
 
+func MustGenerateFlags(fs *pflag.FlagSet, v any) {
+	if err := GenerateFlags(fs, v); err != nil {
+		panic(err)
+	}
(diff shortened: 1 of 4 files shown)


internal/pkg/service/common/configmap/visit.go
```diff
@@ -46,6 +46,8 @@ type VisitContext struct {
 	Sensitive bool
 	// Usage contains value from the "configUsage" tag, if any.
 	Usage string
+	// Usage contains value from the "configShorthand" tag, if any.
+	Shorthand string
 	// Validate contains value from the "validate" tag, if any.
```
michaljurecko, line 51: Added `VisitContext.Shorthand` field, it stores value from the parent, if `configmap.Value.Value` field is found.

internal/pkg/service/common/configmap/visit.go
```diff
@@ -131,12 +133,18 @@ func doVisit(vc *VisitContext, cfg VisitConfig) error {
 			// Mark field and all its children as sensitive according to the tag
 			field.Sensitive = vc.Sensitive || field.StructField.Tag.Get(sensitiveTag) == "true"
 
-			// Set usage from the tag, or use parent usage text
+			// Set usage from the tag, or use parent value
 			field.Usage = vc.Usage
 			if usage := field.StructField.Tag.Get(configUsageTag); usage != "" {
 				field.Usage = usage
 			}
```
michaljurecko, line 140: Implementation has been partially here, but the code for `shorthand` was missing.

internal/pkg/service/common/configmap/flags.go
```diff
@@ -41,8 +47,8 @@ func GenerateFlags(fs *pflag.FlagSet, v any) error {
 				return nil
 			}
 
-			shorthand := vc.StructField.Tag.Get(configShorthandTag)
-			usage := vc.StructField.Tag.Get(configUsageTag)
+			shorthand := vc.Shorthand
+			usage := vc.Usage
```
michaljurecko, line 51: The main part of the bug. In the `configmap.Value` use-case, we have to use parent metadata.

hosekpeter (approved):
LGTM

PR #1590, opened 2024-02-16
feat: Remove global config from Service scope

michaljurecko:

**Changes:**
- Removed top-level config from the dependencies container.

-----------

### DIFF
--- a/internal/pkg/service/stream/dependencies/dependencies.go
+++ b/internal/pkg/service/stream/dependencies/dependencies.go
@@ -34,7 +34,6 @@ package dependencies
 
 import (
 	"github.com/keboola/keboola-as-code/internal/pkg/service/common/dependencies"
-	"github.com/keboola/keboola-as-code/internal/pkg/service/stream/config"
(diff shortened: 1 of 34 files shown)


internal/pkg/service/stream/dependencies/dependencies.go
```diff
@@ -46,7 +46,6 @@ type ServiceScope interface {
 	dependencies.PublicScope
 	dependencies.EtcdClientScope
 	dependencies.TaskScope
-	Config() config.Config
 }
```
michaljurecko, line 49: Removed top-level config from the dependencies container.

It is anti-pattern (with such large app) to pass the whole configuration when only part of it is needed. It leads to cyclic dependencies between packages, since the top-level configuration refers to almost every package.

michaljurecko: Instead, when creating individual services, only the configuration that is needed is passed.

hosekpeter (approved):
LGTM

PR #1593, opened 2024-02-18
feat: Add etcdop.Mutex implementation

michaljurecko:

**Changes:**
- Crated `etcdop.Mutex`, it combines cluster and local lock.
  - So in the application, we only lock the `mutex`, and all other work on the same or another node in the cluster must wait.
  - The code in the application is simple, complex code is in one place and well tested.

-----------


internal/pkg/service/common/etcdop/mutex.go
```diff
@@ -0,0 +1,347 @@
...
+//
+// Implementation:
+//   - Mutex is only a user-facing interface to the internal struct mutexStore.
+//   - The local lock is implemented via a channel, so it is possible to cancel waiting for the lock via the context.
+//   - The "usage" counter prevents memory leaks, unused references are deleted.
+type Mutex struct {
+	store  *mutexStore
+	name   string
+	locked *activeMutex
+}
```
michaljurecko, line 25: Go native lock seem like `mutex.Lock()`, there is no` mutex.Lock(ctx)`, but all our operations have a timeout.

But context aware locks can be simple implemented via a channel:
https://h12.io/article/go-pattern-context-aware-lock

internal/pkg/service/common/etcdop/mutex.go
```diff
@@ -0,0 +1,347 @@
...
+		s.allLock.Lock()
+		mtx.dbSession = nil
+		mtx.dbMutex = nil
+		delete(s.all, mtx.name) // prevent memory leak, remove unused mutex from the map
+		close(mtx.teardown)
+		s.allLock.Unlock()
+	}
+
+	return err
+}
```
michaljurecko, line 231: This is the hard part, we should prevent memory leaks.

The lock is for example a `FileKey.String()`, so, it is used several times and then should be removed from the map.

But during removal/clear in the database, a request for the same lock may come (edge-case).

So the new request must wait for the DB unlock, a then a new lock can be acquired.

internal/pkg/service/common/etcdop/mutex_test.go
```diff
@@ -0,0 +1,239 @@
...
+
+	"github.com/keboola/keboola-as-code/internal/pkg/log"
+	"github.com/keboola/keboola-as-code/internal/pkg/utils/errors"
+	"github.com/keboola/keboola-as-code/internal/pkg/utils/etcdhelper"
+)
+
+func TestMutex_LockUnlock(t *testing.T) {
+	t.Parallel()
+
+	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
```
michaljurecko, line 23: Test for basic operations.

internal/pkg/service/common/etcdop/mutex_test.go
```diff
@@ -0,0 +1,239 @@
...
+	wg.Wait()
+	assert.ErrorIs(t, lock1.TryLock(ctx), context.Canceled)
+	assert.ErrorIs(t, lock2.TryLock(ctx), context.Canceled)
+}
+
+func TestMutex_ParallelWork(t *testing.T) {
+	t.Parallel()
+
+	type testCase struct {
+		// Sessions - count on parallel sessions - virtual nodes.
```
michaljurecko, line 95: Extreme parallel testing for edge-cases.

internal/pkg/service/common/etcdop/mutex_test.go
michaljurecko: Moved here from `session_test.go`.

provisioning/dev/docker/Dockerfile
```diff
@@ -18,6 +18,7 @@ RUN mkdir -p /tmp/build
 COPY Makefile /tmp/build/Makefile
 COPY scripts  /tmp/build/scripts
 RUN cd /tmp/build && make tools && rm -rf /tmp/build
+RUN apt update && apt install -y graphviz
```
michaljurecko, line 21: for `go tool trace`.

internal/pkg/service/common/etcdop/mutex_test.go
```diff
@@ -0,0 +1,239 @@
...
+	require.NoError(t, lock2.Unlock(ctx))
+
+	// Unlock with cancelled context
+	require.NoError(t, lock1.TryLock(ctx))
+	if err = lock1.Unlock(cancelledContext); assert.Error(t, err) {
+		assert.ErrorIs(t, err, context.Canceled)
+	}
+
+	// Close session
+	cancel()
```
hosekpeter, line 85: Do we need to use this, if defer cancel() is used above?

michaljurecko: Yes, because we are testing `TryLock` with canceled context and session:
image

`defer cancel()` in the upper part of the test could be omitted, but there is no problem, ... and the context will end correctly if the test does not reach the end due to some fatal error

PR #1609, opened 2024-02-24
feat: Remote create table command - synced with main

michaljurecko:
Jira: XXXXX

**Changes:**
- ...

-----------


michaljurecko:
Commits up to `build: Update dependencies` have been already reviewd.

internal/pkg/validator/custom.go
```diff
@@ -187,3 +189,13 @@ func (v *wrapper) registerCustomRules() {
...
 }
+
+func (v *wrapper) registerCustomTypes() {
+	// Convert value to string using String method
+	v.validator.RegisterCustomTypeFunc(func(field reflect.Value) interface{} {
+		if v, ok := field.Interface().(fmt.Stringer); ok {
+			return v.String()
+		}
+		return field
+	}, model.SemVersion{})
```
michaljurecko, line 200: There are some fields with the type `model.SemVersion`, the value is no more automatically converted to string, so we need to register the callback.

internal/pkg/validator/validator.go
```diff
@@ -60,13 +60,15 @@ type wrapper struct {
 func New(rules ...Rule) Validator {
 	// Create validator and translator
 	v := &wrapper{
-		validator:  validator.New(),
+		// WithRequiredStructEnabled: https://github.com/go-playground/validator/issues/1142
+		validator:  validator.New(validator.WithRequiredStructEnabled()),
```
michaljurecko, line 64: Default behaviour from the previous versions is now replaced by the option.

hosekpeter (approved):
LGTM

PR #1621, opened 2024-02-27
tests: Fix test timeout

michaljurecko:
Slack:

**Changes:**
- Fixed test timeout.

-----------

Test goroutines are waiting for the `<-start`.

image

### DIFF
--- a/internal/pkg/service/common/etcdop/mutex_test.go
+++ b/internal/pkg/service/common/etcdop/mutex_test.go
@@ -187,9 +187,9 @@ func TestMutex_ParallelWork(t *testing.T) {
 					// Create client and session
 					client := etcdhelper.ClientForTest(t, etcdCfg)
 					session, errCh := NewSessionBuilder().Start(ctx, sessionWg, logger, client)
+					readyWg.Done()
 					require.NotNil(t, session)
 					require.NoError(t, <-errCh)
-					readyWg.Done()
 
 					// Create N unique locks in the session
 					locksWg := &sync.WaitGroup{}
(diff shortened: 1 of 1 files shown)


internal/pkg/service/common/etcdop/mutex_test.go
```diff
@@ -205,7 +205,12 @@ func TestMutex_ParallelWork(t *testing.T) {
 								defer locksWg.Done()
 
 								// Start all work at the same time
-								<-start
+								select {
+								case <-ctx.Done():
+									require.Fail(t, "timeout when waiting for the start channel")
+								case <-start:
+									// continue
+								}
```
michaljurecko, line 213: Context has timeout, but here was infinite waiting for the channel closing.

internal/pkg/service/common/etcdop/mutex_test.go
```diff
@@ -187,9 +187,9 @@ func TestMutex_ParallelWork(t *testing.T) {
 					// Create client and session
 					client := etcdhelper.ClientForTest(t, etcdCfg)
 					session, errCh := NewSessionBuilder().Start(ctx, sessionWg, logger, client)
+					readyWg.Done()
 					require.NotNil(t, session)
 					require.NoError(t, <-errCh)
-					readyWg.Done()
```
michaljurecko, line 192: The problem probably arose here, `require.*` terminates the goroutine in case of an error, so `wg` is not unlocked.

The original issue was probably a network error, we'll see when it happens again. The timeout should be now solved.

PR #1625, opened 2024-02-28
feat: Implement appconfig package

jachym-tousek-keboola:
Jira: https://keboola.atlassian.net/browse/PSGO-386

**Changes:**
- Added appconfig package to load configuration from Sandboxes API.
- Unused for now, proxy will use it in next PR.

-----------

### DIFF
--- a/internal/pkg/service/appproxy/appconfig/api.go
+++ b/internal/pkg/service/appproxy/appconfig/api.go
@@ -12,6 +12,7 @@ import (
 
 type AppProxyConfig struct {
 	ID              string         `json:"id"`
+	Name            string         `json:"name"`
 	UpstreamAppHost string         `json:"upstreamAppHost"`
 	AuthProviders   []AuthProvider `json:"authProviders"`
 	AuthRules       []AuthRule     `json:"authRules"`
@@ -21,6 +22,7 @@ type AppProxyConfig struct {
 
 type AuthProvider struct {
 	ID           string   `json:"id"`
+	Name         string   `json:"name"`
 	Type         string   `json:"type"`
 	ClientID     string   `json:"clientId"`
 	ClientSecret string   `json:"clientSecret"`
@@ -43,8 +45,20 @@ func GetAppProxyConfig(sender request.Sender, appID string, eTag string) request
 		AndPathParam("appId", appID).
 		AndHeader("If-None-Match", eTag).
 		WithOnSuccess(func(ctx context.Context, response request.HTTPResponse) error {
+			// Use id as fallback until name is added to Sandboxes API
+			if result.Name == "" {
+				result.Name = result.ID
+			}
+			for i, provider := range result.AuthProviders {
+				if provider.Name == "" {
+					result.AuthProviders[i].Name = provider.ID
+				}
+			}
+
+			// Add ETag to result
 			result.eTag = response.ResponseHeader().Get("ETag")
 
+			// Process Cache-Control header
 			cacheControl := response.ResponseHeader().Get("Cache-Control")
 			if cacheControl == "" {
 				return nil
--- a/internal/pkg/service/appproxy/appconfig/appconfig_test.go
+++ b/internal/pkg/service/appproxy/appconfig/appconfig_test.go
@@ -73,6 +73,7 @@ func TestLoader_LoadConfig(t *testing.T) {
 					},
 					expectedConfig: AppProxyConfig{
 						ID:              "3",
+						Name:            "3",
 						UpstreamAppHost: "app.local",
 						eTag:            `"etag-value"`,
 						maxAge:          60 * time.Second,
@@ -90,6 +91,7 @@ func TestLoader_LoadConfig(t *testing.T) {
 					},
 					expectedConfig: AppProxyConfig{
 						ID:              "4",
+						Name:            "4",
 						UpstreamAppHost: "app.local",
 						eTag:            `"etag-value"`,
 						maxAge:          60 * time.Second,
@@ -98,6 +100,7 @@ func TestLoader_LoadConfig(t *testing.T) {
 				{
 					expectedConfig: AppProxyConfig{
 						ID:              "4",
+						Name:            "4",
 						UpstreamAppHost: "app.local",
 						eTag:            `"etag-value"`,
 						maxAge:          60 * time.Second,
@@ -115,6 +118,7 @@ func TestLoader_LoadConfig(t *testing.T) {
 					},
 					expectedConfig: AppProxyConfig{
 						ID:              "5",
+						Name:            "5",
 						UpstreamAppHost: "app.local",
 						eTag:            `"etag-value"`,
 						maxAge:          60 * time.Second,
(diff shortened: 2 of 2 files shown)


internal/pkg/service/appproxy/appconfig/api.go
```diff
@@ -0,0 +1,109 @@
...
+	return request.NewAPIRequest(result, req)
+}
+
+// SandboxesError represents the structure of API error.
+type SandboxesError struct {
+	Message     string `json:"error"`
+	ExceptionID string `json:"exceptionId"`
+	request     *http.Request
+	response    *http.Response
+}
```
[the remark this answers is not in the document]
michaljurecko: :man_facepalming: ok :heavy_check_mark:

internal/pkg/service/appproxy/appconfig/api.go
```diff
@@ -0,0 +1,122 @@
...
+
+			result.maxAge = time.Second * time.Duration(cacheDirectives.MaxAge)
+			return nil
+		})
+	return request.NewAPIRequest(result, req)
+}
+
+// SandboxesError represents the structure of API error.
+type SandboxesError struct {
+	Message     string `json:"error"`
```
michaljurecko, line 84: I suggest to move error code to a new `error.go`.

internal/pkg/service/appproxy/appconfig/appconfig.go
```diff
@@ -0,0 +1,106 @@
...
+func NewLoader(logger log.Logger, clock clock.Clock, baseURL string) *Loader {
+	return &Loader{
+		logger: logger,
+		clock:  clock,
+		sender: client.New().WithBaseURL(baseURL),
+		cache:  make(map[string]cacheItem),
+	}
+}
+
+const staleCacheFallbackDuration = time.Hour
```
michaljurecko, line 39: Please move `const` before types, and add some comment - `staleCacheFallbackDuration is the maximum duration for which the old configuration of the application is used if loading the new one is not possible.`

internal/pkg/service/appproxy/appconfig/appconfig.go
```diff
@@ -0,0 +1,106 @@
...
+		if err != nil {
+			return l.handleError(ctx, appID, now, err, &item)
+		}
+
+		// Update expiration and use the cached config if ETag is still the same
+		if config.eTag == item.eTag {
+			l.cache[appID] = cacheItem{
+				config:    item.config,
+				eTag:      item.eTag,
+				expiresAt: now.Add(config.maxAge),
```
michaljurecko, line 63: What if we set a maximum to be sure, e.g. `1 hour` ?

internal/pkg/service/appproxy/appconfig/api.go
```diff
@@ -0,0 +1,122 @@
...
+	AllowedRoles []string `json:"allowedRoles"`
+}
+
+type AuthRule struct {
+	Type  string   `json:"type"`
+	Value string   `json:"value"`
+	Auth  []string `json:"auth"`
+}
+
+func GetAppProxyConfig(sender request.Sender, appID string, eTag string) request.APIRequest[*AppProxyConfig] {
```
michaljurecko, line 39: Is the pointer `*AppProxyConfig` needed here? We prefer to pass "entity" as value if possible.

internal/pkg/service/appproxy/appconfig/appconfig.go
```diff
@@ -0,0 +1,106 @@
...
+		logger: logger,
+		clock:  clock,
+		sender: client.New().WithBaseURL(baseURL),
+		cache:  make(map[string]cacheItem),
+	}
+}
+
+const staleCacheFallbackDuration = time.Hour
+
+func (l *Loader) LoadConfig(ctx context.Context, appID string) (AppProxyConfig, error) {
```
michaljurecko, line 41: I think, the code can be simplified.

```go
func (l *Loader) LoadConfig(ctx context.Context, appID string) (config AppProxyConfig, modified bool, err error) {
	// Get config from cache
	eTag := ""
	now := l.clock.Now()
	if item, ok := l.cache[appID]; ok {
		eTag = item.eTag

		// Return config from cache if still valid
		if now.Before(item.expiresAt) {
			return item.config, false, nil
		}
	}

	// Config is not cached or it is expired, load a new one
	config, err = GetAppProxyConfig(l.sender, appID, eTag).Send(ctx)
	if err != nil {
		return l.handleError(ctx, appID, now, err, nil)
	}

	// Cache the new config
	modified = eTag == "" || config.eTag == eTag // covers also config.eTag == ""
	if modified {
		l.cache[appID] = cacheItem{
			config:    config,
			eTag:      config.eTag,
			expiresAt: now.Add(config.maxAge),
		}
	}

	return config, modified, nil
}
```

internal/pkg/service/appproxy/appconfig/appconfig.go
```diff
@@ -0,0 +1,106 @@
...
+	if sandboxesError != nil {
+		logger = l.logger.With(attribute.String("exceptionId", sandboxesError.ExceptionID))
+	}
+
+	// An error other than 404 is considered a temporary failure. Keep using the stale cache for staleCacheFallbackDuration as fallback.
+	if fallbackItem != nil && now.Before(fallbackItem.expiresAt.Add(staleCacheFallbackDuration)) {
+		logger.Warnf(ctx, `Using stale cache for app "%s": %s`, appID, err.Error())
+
+		return fallbackItem.config, nil
+	}
```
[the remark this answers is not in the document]
michaljurecko: :heavy_check_mark:

internal/pkg/service/appproxy/appconfig/appconfig_test.go
```diff
@@ -0,0 +1,319 @@
...
+		tc := tc
+		t.Run(tc.name, func(t *testing.T) {
+			t.Parallel()
+
+			clk := clock.NewMock()
+			transport := httpmock.NewMockTransport()
+
+			url := "https://sandboxes.keboola.com"
+			loader := NewLoader(log.NewDebugLogger(), clk, url)
+			loader.sender = loader.sender.(client.Client).WithTransport(transport)
```
michaljurecko, line 273: Please, inject the client as a dep. via constructor.

What if you need to mock the transport at a higher level?

You can also switch the test to blackbox mode, `package ..._test`, you will only test the public interface of the Loader, and it will lead to a better design.

internal/pkg/service/appproxy/appconfig/appconfig_test.go
```diff
@@ -0,0 +1,307 @@
...
+	attempts []attempt
+}
+
+type attempt struct {
+	delay             time.Duration
+	responses         []*http.Response
+	expectedETag      string
+	expectedErrorCode int
+	expectedConfig    AppProxyConfig
+}
```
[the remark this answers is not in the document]
michaljurecko: Nice test :+1:

michaljurecko (commented):
Nice, I have some notes

internal/pkg/service/appproxy/appconfig/appconfig.go
michaljurecko: It can be simple implemented using generic types, but it is not present in the std lib:
https://github.com/golang/go/issues/50019

internal/pkg/service/appproxy/appconfig/api.go
michaljurecko: Ok, it may be that it is not solved perfectly there.

michaljurecko (approved):
LGTM

PR #1631, opened 2024-03-01
feat: Use appconfig Loader

jachym-tousek-keboola:
Jira: [keboola.atlassian.net/browse/PSGO-386](https://keboola.atlassian.net/browse/PSGO-386)

**Changes:**
- App Proxy is now using the appconfig package from previous PR.
- Fixed a possible security issue with stripping `X-KBC-` headers due to case sensitivity.
- Updated OAuth2 Proxy to `7.6.0`.
- Some refactorings.

-----------

### DIFF
--- a/internal/pkg/service/appproxy/http/router.go
+++ b/internal/pkg/service/appproxy/http/router.go
@@ -138,7 +138,7 @@ func (r *Router) CreateHandler() http.Handler {
 
 func (r *Router) createConfigErrorHandler(exceptionID string) http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, req *http.Request) {
-		r.logger.Warn(req.Context(), `application has misconfigured OAuth2 provider`)
+		r.logger.With(attribute.String("exceptionId", exceptionID)).Warn(req.Context(), `application "<proxy.appid>" has misconfigured OAuth2 provider`)
 		w.WriteHeader(http.StatusForbidden)
 		fmt.Fprintln(w, "Application has misconfigured OAuth2 provider.")
 		fmt.Fprintln(w, "Exception ID: ", exceptionID)


internal/pkg/service/appproxy/http/router.go
```diff
@@ -92,22 +96,43 @@ func (r *Router) CreateHandler() http.Handler {
...
+		}
+
+		// Recreate app handler if configuration changed.
+		handler, ok := r.appHandlers[appID]
+		if !ok || handler.Etag != config.ETag {
+			handler = appHandler{
+				Handler: r.createDataAppHandler(req.Context(), config),
+				Etag:    config.ETag,
+			}
+			r.appHandlers[appID] = handler
```
michaljurecko, line 132: Isn't there a lock needed? (`sync.Mutex/RWMutex`)

michaljurecko: `sync.Map` is not generic, so it requires re-typing, therefore it is used very rarely.

michaljurecko: Because you don't want to see these errors in production. :slightly_smiling_face:

```
fatal error: concurrent map read and map
```

https://stackoverflow.com/questions/45585589/golang-fatal-error-concurrent-map-read-and-map-write

michaljurecko: > Oh I thought you wanted the lock around

That would be nice too, but it will work without it. But the lock I'm writing about is essential.

michaljurecko: Because you probably don't have a test that sends parallel request and change the configuration of the app at the same time. (?)

michaljurecko: - Sending 100/1000 parallel requests is easy (100 goroutines + waitgroup).
- It is more difficult to figure out the app configuration change: there may be some Sandboxes API mock that will always send a random ETag and thus re-configuration will always happen? (or every `count % 5 ==0` times), or something similar.

internal/pkg/service/appproxy/http/app.go
```diff
@@ -53,7 +29,7 @@ func appIDMiddleware(publicURL *url.URL) middleware.Middleware {
 	}
 }
 
-func parseAppID(publicURL *url.URL, host string) (AppID, bool) {
+func parseAppID(publicURL *url.URL, host string) (string, bool) {
```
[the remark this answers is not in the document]
michaljurecko: The benefit is that you cannot confuse different string values that have different meanings, because it is protected by the type. 

I see, it is used only in a telemetry attribute.

michaljurecko (approved):
LGTM

hosekpeter (approved):

PR #1634, opened 2024-03-04
feat: Update Source E2E + E2E tests

michaljurecko:
Jira: https://keboola.atlassian.net/browse/PSGO-287

**Changes:**
- Finished `update source` endpoint.
- Aded E2E tests.
- Improved tasks outputs.

-----------


internal/pkg/service/stream/definition/repository/source_repo.go
```diff
@@ -126,16 +126,20 @@ func (r *SourceRepository) Create(versionDescription string, input *definition.S
 }
 
 //nolint:dupl // similar code is in SinkRepository
-func (r *SourceRepository) Update(k key.SourceKey, versionDescription string, updateFn func(definition.Source) definition.Source) *op.AtomicOp[definition.Source] {
+func (r *SourceRepository) Update(k key.SourceKey, versionDescription string, updateFn func(definition.Source) (definition.Source, error)) *op.AtomicOp[definition.Source] {
 	var result definition.Source
```
michaljurecko, line 130: Here is the `SourceRepository`, the `Update` method, with the `updateFn` callback. 

Reminder: update is implemented via the callback, so operation can be retried on a collision and result is consistent.

michaljurecko: I updated the callback type, so it can return an error.

api/stream/design.go
```diff
@@ -662,6 +662,10 @@ var ListSourcesRequest = Type("ListSourcesRequest", func() {
 var UpdateSourceRequest = Type("UpdateSourceRequest", func() {
 	SourceKeyRequest()
 	SourceFieldsRW()
+	Attribute("changeDescription", String, func() {
+		Description("Description of the modification, description of the version.")
+		Example("Renamed.")
+	})
 })
```
michaljurecko, line 669: Added mission `changeDescription` API field, it describes the new version of the entity.

internal/pkg/service/common/task/node.go
```diff
@@ -299,20 +301,40 @@ func (n *Node) runTask(logger log.Logger, task Task, cfg Config) (result Result,
...
+		}
+		// Convert value to a string if possible
+		if str, err := cast.ToStringE(v); err == nil {
+			attrs = append(attrs, attribute.String("result_outputs."+k, str))
+		}
+	}
+	sort.SliceStable(attrs, func(i, j int) bool {
+		return attrs[i].Key < attrs[j].Key
+	})
+
```
michaljurecko, line 320: See `trace.go` bellow.

Task can have some `outputs` (Key-Value pairs).

These outputs were correctly added to `span` as attributes (sed `trace.go`), but in the log message, they were serialized in JSON format. I improved it, they are now mapped to attributes also in the log message.

internal/pkg/service/common/task/node_test.go
```diff
@@ -66,8 +66,11 @@ func TestSuccessfulTask(t *testing.T) {
 		Operation: func(ctx context.Context, logger log.Logger) task.Result {
 			defer close(taskDone)
 			<-taskWork
+
 			logger.Info(ctx, "some message from the task (1)")
-			return task.OkResult("some result (1)").WithOutput("key", "value")
+			return task.OkResult("some result (1)").
+				WithOutput("key", "value").
+				WithOutputsFrom(map[string]int{"int1": 1, "int2": 2})
```
michaljurecko, line 73: Added new method `WithOutputsFrom`, here is updated test.

internal/pkg/service/stream/api/mapper/task.go
```diff
@@ -11,7 +11,15 @@ import (
...
 )
 
-func (m *Mapper) NewTaskResponse(entity task.Task) *api.Task {
+type taskOutputs struct {
+	URL       *string            `mapstructure:"url,omitempty"`
+	ProjectID *keboola.ProjectID `mapstructure:"projectId,omitempty"`
+	BranchID  *keboola.BranchID  `mapstructure:"branchId,omitempty"`
+	SourceID  *key.SourceID      `mapstructure:"sourceId,omitempty"`
+	SinkID    *key.SinkID        `mapstructure:"sinkId,omitempty"`
+}
```
michaljurecko, line 20: I simplified the code bellow, there is no more a map with strings keys, but the `taskOutputs` struct.

test/stream/api/source/update-001-ok/009-update-3-type-nop/request.json
```diff
@@ -0,0 +1,12 @@
+{
+  "path": "/v1/branches/%%TEST_DEFAULT_BRANCH_ID%%/sources/my-source",
+  "method": "PATCH",
+  "headers": {
+    "Content-Type": "application/json",
+    "X-StorageApi-Token": "%%TEST_KBC_STORAGE_API_TOKEN%%"
+  },
+  "body": {
+    "changeDescription": "Super cool modification 2.",
+    "type": "http"
```
michaljurecko, line 10: Improved `update-001-ok` test, set the same value without modification.

api/stream/design.go
```diff
@@ -191,7 +191,7 @@ var _ = Service("stream", func() {
 	Method("UpdateSource", func() {
 		Meta("openapi:summary", "Update source")
 		Description("Update the source.")
-		Result(Source)
+		Result(Task)
 		Payload(UpdateSourceRequest)
```
michaljurecko, line 195: Forward compatibility, there is no blocking operation when updating `HTTP Source`, but different source types may require a long-running operation.

hosekpeter (approved):
LGTM

internal/pkg/service/common/task/node.go
```diff
@@ -299,20 +301,40 @@ func (n *Node) runTask(logger log.Logger, task Task, cfg Config) (result Result,
...
+	// Use task outputs in log message and telemetry
+	var attrs []attribute.KeyValue
+	for k, v := range task.Outputs {
+		// Skip nil values
+		if v := reflect.ValueOf(v); !v.IsValid() || (v.Kind() == reflect.Pointer && v.IsZero()) {
+			continue
+		}
+		// Convert value to a string if possible
+		if str, err := cast.ToStringE(v); err == nil {
+			attrs = append(attrs, attribute.String("result_outputs."+k, str))
```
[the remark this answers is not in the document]
michaljurecko: Yes, this solution is "good enough", and we can improve it further. But now it would have little added value compared to the time required.

PR #1642, opened 2024-03-06
fix: Concurrency issues

jachym-tousek-keboola:
Jira: https://keboola.atlassian.net/browse/PSGO-457

**Changes:**
- Fixed potential issues with concurrency.

-----------

### DIFF
--- a/internal/pkg/service/appsproxy/appconfig/appconfig.go
+++ b/internal/pkg/service/appsproxy/appconfig/appconfig.go
@@ -42,7 +42,7 @@ func NewSandboxesAPILoader(logger log.Logger, clock clock.Clock, client client.C
 		logger: logger,
 		clock:  clock,
 		sender: client.WithBaseURL(baseURL).WithHeader("X-KBC-ManageApiToken", token),
-		cache: syncmap.NewSyncMap[string, cacheItem](func() *cacheItem {
+		cache: syncmap.New[string, cacheItem](func() *cacheItem {
 			return &cacheItem{
 				updateLock: &sync.Mutex{},
 			}
--- a/internal/pkg/service/appsproxy/http/router.go
+++ b/internal/pkg/service/appsproxy/http/router.go
@@ -69,7 +69,7 @@ func NewRouter(d dependencies.ServiceScope, exceptionIDPrefix string) (*Router, 
 		config:    d.Config(),
 		clock:     d.Clock(),
 		loader:    d.Loader(),
-		appHandlers: syncmap.NewSyncMap[string, appHandler](func() *appHandler {
+		appHandlers: syncmap.New[string, appHandler](func() *appHandler {
 			return &appHandler{
 				updateLock: &sync.RWMutex{},
 			}
@@ -145,9 +145,9 @@ func (r *Router) CreateHandler() http.Handler {
 
 		if modified || httpHandler == nil {
 			handler.setHTTPHandler(func() http.Handler {
-				return r.createDataAppHandler(req.Context(), config)
+				httpHandler = r.createDataAppHandler(req.Context(), config)
+				return httpHandler
 			})
-			httpHandler = handler.getHTTPHandler()
 		}
 
 		httpHandler.ServeHTTP(w, req)
--- a/internal/pkg/service/appsproxy/syncmap/syncmap_test.go
+++ b/internal/pkg/service/appsproxy/syncmap/syncmap_test.go
@@ -12,15 +12,28 @@ import (
 
 type testStruct struct{}
 
-func TestSyncMap(t *testing.T) {
+func TestSyncMap_GetOrInit(t *testing.T) {
 	t.Parallel()
 
-	m := syncmap.NewSyncMap[string, testStruct](func() *testStruct {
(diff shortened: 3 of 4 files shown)


internal/pkg/service/appsproxy/http/router.go
```diff
@@ -118,13 +140,15 @@ func (r *Router) CreateHandler() http.Handler {
...
-		if !ok || modified {
-			handler = r.createDataAppHandler(req.Context(), config)
-			r.appHandlers[appID] = handler
+		handler := r.appHandlers.GetOrInit(appID)
+		httpHandler := handler.getHTTPHandler()
+
+		if modified || httpHandler == nil {
+			httpHandler = r.createDataAppHandler(req.Context(), config)
+			handler.setHTTPHandler(httpHandler)
 		}
```
michaljurecko, line 149: `httpHandler = r.createDataAppHandler(req.Context(), config)` part is not protected by the lock and can run more times.

Maybe we can use something like following code?
```
httpHandler := handler.getOrInit(
    return r.createDataAppHandler(req.Context(), config)
)
```

I understand that the problem is `ctx` and `config`.

michaljurecko: You can implement this use-case specially. Or modify `SafeMap` so that the `init callback` is passed as an argument to the `GetOrInit` method.

michaljurecko (commented):
Nice, there is the last issue

internal/pkg/service/appsproxy/http/router.go
```diff
@@ -63,19 +64,40 @@ func NewRouter(d dependencies.ServiceScope, exceptionIDPrefix string) (*Router,
...
+
+func (h *appHandler) setHTTPHandler(httpHandler http.Handler) {
+	h.updateLock.Lock()
+	defer h.updateLock.Unlock()
+	h.httpHandler = httpHandler
+}
+
 func (r *Router) CreateHandler() http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, req *http.Request) {
 		appIDString, ok := ctxattr.Attributes(req.Context()).Value(attrAppID)
```
michaljurecko, line 103: Suggestion:
- From the code, I don't know where the `appID` is set/defined.
- Telemetry attributes are not intended for values passing.
- You can use something like:
```
ctx = context.WithValue(ctx, appIDCtxKey, appID)

....

appID, ok := ctx.Value(appIDCtxKey).(string)
```

internal/pkg/service/appsproxy/http/router.go
```diff
@@ -143,8 +143,10 @@ func (r *Router) CreateHandler() http.Handler {
 		httpHandler := handler.getHTTPHandler()
 
 		if modified || httpHandler == nil {
-			httpHandler = r.createDataAppHandler(req.Context(), config)
-			handler.setHTTPHandler(httpHandler)
+			handler.setHTTPHandler(func() http.Handler {
+				return r.createDataAppHandler(req.Context(), config)
+			})
+			httpHandler = handler.getHTTPHandler()
```
michaljurecko: Thanks, it is better.

Some more improvements could be made here so that it is all under one lock, but let's leave that for another iteration.

internal/pkg/service/appsproxy/appconfig/appconfig.go
```diff
@@ -41,7 +42,7 @@ func NewSandboxesAPILoader(logger log.Logger, clock clock.Clock, client client.C
 		logger: logger,
 		clock:  clock,
 		sender: client.WithBaseURL(baseURL).WithHeader("X-KBC-ManageApiToken", token),
-		cache: NewSyncMap[string, cacheItem](func() *cacheItem {
+		cache: syncmap.NewSyncMap[string, cacheItem](func() *cacheItem {
```
michaljurecko, line 45: `syncmap.NewSyncMap` -> `syncmap.New`

internal/pkg/service/appsproxy/syncmap/syncmap_test.go
```diff
@@ -0,0 +1,40 @@
...
+	})
+
+	wg := sync.WaitGroup{}
+	counter := atomic.NewInt64(0)
+	for i := 0; i < 10; i++ {
+		wg.Add(1)
+		go func() {
+			defer wg.Done()
+
+			m.GetOrInit("test")
```
michaljurecko, line 29: I'm missing an assert here, e.g. that you got the same reference 10 times.

michaljurecko (approved):
LGTM

internal/pkg/service/appsproxy/syncmap/syncmap_test.go
michaljurecko: Ok, so it tests the race condition, and it doesn't test the internal function of that map.

The easiest is to do 2 separate tests (so you can add another one).
 - `TestSyncMap_GetOrInit_Race`
 - `TestSyncMap_GetOrInit`

michaljurecko (approved):
LGTM

hosekpeter (approved):

PR #1651, opened 2024-03-08
build: Fix Go cache

jachym-tousek-keboola:
Right now the Go cache action shows a lot of errors `Cannot open: File exists`. I'm not sure but my first guess is that it might be because of the duplication of directory names with `\` and `/`. So I wanna try splitting Windows from the rest to mitigate this.

### DIFF
--- a/.github/actions/go-cache/action.yml
+++ b/.github/actions/go-cache/action.yml
@@ -4,6 +4,8 @@ inputs:
   key:
     description: Cache key
     required: true
+env:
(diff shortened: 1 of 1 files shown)


michaljurecko:
@jachym-tousek-keboola but, cache is separated by the OS, isn't it?

image

---------------------

Path should be only a list of allowed paths.

image


image

michaljurecko (approved):
Ok, we can test it, if it doesn't help, please revert it.

PR #1653, opened 2024-03-11
build: Improve Go cache

jachym-tousek-keboola:
This fixes the file already exists errors. I don't see any installation of packages in the log so caching still works fine.

### DIFF
--- a/.github/workflows/check-licenses.yml
+++ b/.github/workflows/check-licenses.yml
@@ -14,15 +14,16 @@ jobs:
       - name: Checkout code
         uses: actions/checkout@v4
 
-      - name: Install Go
-        uses: actions/setup-go@v5
-        with:
-          go-version: ${{ env.GO_VERSION }}
-
       - name: Load Go cache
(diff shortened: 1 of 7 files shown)


.github/workflows/check-licenses.yml
```diff
@@ -14,16 +14,17 @@ jobs:
...
       - name: Load Go cache
         uses: ./.github/actions/go-cache
         with:
           key: lint
 
+      - name: Install Go
+        uses: actions/setup-go@v5
+        with:
+          go-version: ${{ env.GO_VERSION }}
+          cache: false
```
hosekpeter, line 26: so it's the same thing I suggested  `cache=false`

hosekpeter (approved):

.github/workflows/check-licenses.yml
michaljurecko: Yes, quite detailed research was done to find out why this is the right solution.

michaljurecko (approved):

PR #1657, opened 2024-03-12
fix: Apps Proxy bugs

jachym-tousek-keboola:
Jira: 
- https://keboola.atlassian.net/browse/PSGO-471
- https://keboola.atlassian.net/browse/PSGO-472
- https://keboola.atlassian.net/browse/PSGO-473

**Changes:**
- Added missing test.
- Improved compliance with RFC.

-----------

### DIFF
--- a/internal/pkg/service/appsproxy/http/router.go
+++ b/internal/pkg/service/appsproxy/http/router.go
@@ -301,7 +301,7 @@ func (r *Router) createProvider(ctx context.Context, authProvider appconfig.Auth
 
 	// AllowedRoles nil means there is no role requirement. Empty array doesn't make sense.
 	if authProvider.AllowedRoles != nil && len(allowedRoles) == 0 {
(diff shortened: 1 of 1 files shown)


internal/pkg/service/appsproxy/http/router.go
```diff
@@ -274,6 +297,11 @@ func (r *Router) createProvider(ctx context.Context, authProvider appconfig.Auth
 	}
 	provider.proxyConfig = proxyConfig
 
+	if authProvider.AllowedRoles != nil && len(allowedRoles) == 0 {
+		r.logger.With(attribute.String("exceptionId", exceptionID)).Warnf(ctx, `empty array of allowed roles for app "%s" "%s"`, app.ID, app.Name)
+		return provider
+	}
+
```
michaljurecko, line 304: I would say that it should be either an error or I would ignore it.

I suggest to ignore it, and switch type from `*[]string` to `[]string`.

michaljurecko: It does not make sense to me that the user sets an empty array there and warnings will come TO US.

michaljurecko (approved):
LGTM

PR #1660, opened 2024-03-12
style: Convert remaining repository tests to blackbox tests

michaljurecko:
Jira: https://keboola.atlassian.net/browse/PSGO-287

**Changes:**
  - Create new `schema` packages, to simplify testing.
    - Code moved without modification.
  - Converted remaining whitebox tests using the blackbox approach.

-----------


internal/pkg/service/stream/definition/repository/branch_limits_test.go
```diff
@@ -1,4 +1,4 @@
-package repository
+package repository_test
 
 import (
 	"context"
```
michaljurecko, line 4: I prefer blackbox testing using `repository_test` pkg, instead of whitebox testing using `repository` pkg.
  - It also prevents some types of import cycle issues.
  - I updated the remaining tests in the Stream Service repositories.

internal/pkg/service/stream/definition/repository/schema/branch_schema.go
```diff
@@ -0,0 +1,37 @@
+package schema
+
```
michaljurecko, line 2: I have to create a new `schema` packages for direct usage in tests. 
The code is moved without modifications.

hosekpeter (approved):
LGTM

PR #1673, opened 2024-03-21
feat: Fix create table BC

hosekpeter:
Jira: [PSGO-496](https://keboola.atlassian.net/browse/PSGO-496)

**Changes:**
- `--columns` accept quotes in column names
- columns are created as nullable
-----------
kbc remote create table --primary-key "" --columns id,firstname,lastname
image

kbc remote create table --primary-key "id" --columns id,firstname,lastname 
image

[PSGO-496]: https://keboola.atlassian.net/browse/PSGO-496?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/internal/pkg/service/cli/cmd/remote/create/table/dialog.go
+++ b/internal/pkg/service/cli/cmd/remote/create/table/dialog.go
@@ -4,7 +4,6 @@ import (
 	"encoding/json"
 	"fmt"
 	"os"
-	"slices"
 	"strings"
 
 	"github.com/keboola/go-client/pkg/keboola"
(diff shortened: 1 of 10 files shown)


internal/pkg/service/cli/cmd/remote/create/table/dialog.go
```diff
@@ -103,14 +100,15 @@ func AskCreateTable(args []string, branchKey keboola.BranchKey, allBuckets []*ke
 		panic(errors.New("unexpected state"))
 	}
 
-	if f.PrimaryKey.IsSet() {
+	if f.PrimaryKey.Value != "" {
 		opts.CreateTableRequest.PrimaryKeyNames = strings.Split(strings.TrimSpace(f.PrimaryKey.Value), ",")
 	} else {
```
michaljurecko, line 105: `else` -> `else if !f.PrimaryKey.IsSet()`

If the user enters `--primary-key ""`, it means, that no primary key should be created.
- So the user has explicitly specified their will and the dialog box should be skipped.

PR #1675, opened 2024-03-22
feat: don't send basetype for columns flag

hosekpeter:
Jira: [PSGO-501](https://keboola.atlassian.net/browse/PSGO-501)

**Changes:**
- basetype doesn't send for columns flag

-----------

[PSGO-501]: https://keboola.atlassian.net/browse/PSGO-501?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/internal/pkg/service/cli/cmd/remote/create/table/create_table_test.go
+++ b/internal/pkg/service/cli/cmd/remote/create/table/create_table_test.go
@@ -456,3 +456,51 @@ func TestParseJsonInput(t *testing.T) {
 		},
 	}, res)
 }
+
+func TestPossiblePrimaryKeys(t *testing.T) {
+	testCases := []struct {
+		name           string
+		exceptedResult []string
+		columns        keboola.Columns
+	}{
+		{
+			name:           "columns-from definition",
+			exceptedResult: []string{"name"},
+			columns: keboola.Columns{
+				{
+					Name: "name",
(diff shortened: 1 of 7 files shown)


internal/pkg/service/cli/cmd/remote/create/table/create_table_test.go
```diff
@@ -2,6 +2,7 @@ package table
 
 import (
 	"fmt"
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore/to"
```
michaljurecko, line 5: We don't want such unrelated dependency.

Please add internal method `ptr`, or create an helper pkg, e.g, `service/common/ptr`

```
func ptr[T any](v T) *T {
	return &v
}
```

internal/pkg/service/cli/cmd/remote/create/table/dialog.go
```diff
@@ -217,7 +216,10 @@ func defaultValue() string {
 func getColumnsName(columns []keboola.Column) []string {
```
michaljurecko, line 216: `getColumnsName` -> `possiblePrimaryKeys`

PR #1678, opened 2024-03-25
feat: Prepare DNS resolver and HTTP Transport

jachym-tousek-keboola:
Jira: https://keboola.atlassian.net/browse/PSGO-363 (part 1)

**Changes:**
- Added a DNS client optimized for our usage.
- Added a HTTP Transport to skip DNS resolution.

-----------

### DIFF
--- a/internal/pkg/service/appsproxy/dataapps/appconfig.go
+++ b/internal/pkg/service/appsproxy/dataapps/appconfig.go
@@ -90,7 +90,7 @@ func (l *sandboxesServiceClient) Notify(ctx context.Context, appID string) error
 	// Send the notification
 	_, err := NotifyAppUsage(l.sender, appID, now).Send(ctx)
 	if err != nil {
-		l.logger.Errorf(ctx, `Failed notifying Sandboxes Service about a request to app "%s": %s`, appID, err.Error())
+		l.logger.Errorf(ctx, `failed notifying Sandboxes Service about a request to app "%s": %s`, appID, err.Error())
 
 		return err
 	}
@@ -101,7 +101,7 @@ func (l *sandboxesServiceClient) Notify(ctx context.Context, appID string) error
 func (l *sandboxesServiceClient) Wakeup(ctx context.Context, appID string) error {
 	_, err := WakeupApp(l.sender, appID).Send(ctx)
 	if err != nil {
-		l.logger.Errorf(ctx, `Failed sending wakeup request to Sandboxes Service about for app "%s": %s`, appID, err.Error())
+		l.logger.Errorf(ctx, `failed sending wakeup request to Sandboxes Service about for app "%s": %s`, appID, err.Error())
 
 		return err
 	}
--- a/internal/pkg/service/appsproxy/dns/client.go
+++ b/internal/pkg/service/appsproxy/dns/client.go
@@ -41,25 +41,20 @@ func NewClientFromEtc(dialer *net.Dialer) (*Client, error) {
 		dnsServer += ":53"
 	}
 
-	return NewClient(dialer, dnsServer)
+	return NewClient(dialer, dnsServer), nil
 }
 
-func NewClient(dialer *net.Dialer, dnsServer string) (*Client, error) {
-	// DNS client
-	client := &dns.Client{
-		Net:          "udp",
-		Dialer:       dialer,
-		DialTimeout:  DialTimeout,
-		ReadTimeout:  DNSReadTimeout,
-		WriteTimeout: DNSWriteTimeout,
-	}
-
-	c := &Client{
-		client:    client,
+func NewClient(dialer *net.Dialer, dnsServer string) *Client {
+	return &Client{
+		client: &dns.Client{
+			Net:          "udp",
+			Dialer:       dialer,
+			DialTimeout:  DialTimeout,
+			ReadTimeout:  DNSReadTimeout,
+			WriteTimeout: DNSWriteTimeout,
+		},
 		dnsServer: dnsServer,
 	}
-
-	return c, nil
 }
 
 func (c *Client) Resolve(ctx context.Context, host string) (string, error) {


internal/pkg/service/appsproxy/dns/client_test.go
```diff
@@ -0,0 +1,28 @@
...
+	client, err := dns.NewClient()
+	require.NoError(t, err)
+
+	// Local DNS server will be unable to resolve keboola.com because of RecursionDesired = false.
+	ip, err := client.Resolve(context.Background(), "keboola.com")
+	assert.Equal(t, ip, "")
+
+	var dnsErr *net.DNSError
+	assert.True(t, errors.As(err, &dnsErr))
+	assert.True(t, dnsErr.IsNotFound)
```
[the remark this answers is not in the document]
michaljurecko: It is probably not possible without mocking the DNS server in the test.

Here is an example DNS server using `miekg`, if you want:
- https://github.com/miekg/dns/blob/2230854ba97edcf29ac55a1f274e49cec11bf9bb/server_test.go#L242
- https://github.com/miekg/dns/blob/2230854ba97edcf29ac55a1f274e49cec11bf9bb/server_test.go#L1004-L1024

(`:0` is a random free port)

internal/pkg/service/appsproxy/dns/client.go
```diff
@@ -0,0 +1,88 @@
...
+	return c, nil
+}
+
+func (c *Client) Resolve(ctx context.Context, host string) (string, error) {
+	msg := &dns.Msg{}
+	msg.SetQuestion(dns.Fqdn(host), dns.TypeA)
+	msg.Authoritative = true
+	msg.RecursionAvailable = false
+	msg.RecursionDesired = false
+
```
michaljurecko, line 68: Please add comment why recursion is disabled:
- It skip caching - by a side effect.
- We want to know if some service pod available in k8s, we don't need to go recursively to other servers.

internal/pkg/service/appsproxy/dns/transport.go
```diff
@@ -0,0 +1,66 @@
...
+
+// NewDNSSkippingHTTPTransport creates an HTTP transport which can skip DNS resolution if the address is already within request context.
+// Timeouts are optimized for our usage in reverse proxy.
+func NewDNSSkippingHTTPTransport(addressCtxKey any) (*http.Transport, error) {
+	dialer := newDialer()
+
+	transport := http.DefaultTransport.(*http.Transport).Clone()
+
+	transport.DialContext = func(ctx context.Context, network, addr string) (net.Conn, error) {
+		address, ok := ctx.Value(addressCtxKey).(string)
```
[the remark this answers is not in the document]
michaljurecko: `dns/transport.go` creates `http.Transport` :thinking: 

The `dns` package should contain only the creation of the `Dialer`, there the DNS layer ends.

HTTP timeouts are not related to the DNS.

Do we have a better place where we would move the transport creation?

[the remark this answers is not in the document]
michaljurecko: > I want to split the work into two parts. 

I see the following separate parts there:
- Use the custom `http.Transport` for the proxy upstream (discussed on Slack).
- Rendering error pages : rendering the "spinner" page.

> Then second is here where the DNS resolution is skipped if the resolved address already exists in the request context.
- I don't understand it, can you please explain? Or let's call shortly.

internal/pkg/service/appsproxy/http/router.go
```diff
@@ -64,12 +68,18 @@ func NewRouter(d dependencies.ServiceScope, exceptionIDPrefix string) (*Router,
 		return nil, errors.PrefixError(err, "could not parse selection template")
 	}
 
+	transport, err := dns.NewDNSSkippingHTTPTransport(AppAddressCtxKey)
+	if err != nil {
```
michaljurecko, line 72: Suggestion, related to the previous comment:
- Move `transport.go`, to the `router` package (or create a new package).
- `transport = NewHTTPTransport(dns.NewDialer())` (+ error handling)

internal/pkg/service/appsproxy/http/router.go
```diff
@@ -171,14 +181,21 @@ func (r *Router) createDataAppHandler(ctx context.Context, app dataapps.AppProxy
 		return r.createConfigErrorHandler(exceptionID)
 	}
 
+	target, err := url.Parse(app.UpstreamAppURL)
```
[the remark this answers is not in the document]
michaljurecko: Why do you want to do a manual lookup?

internal/pkg/service/appsproxy/dns/transport.go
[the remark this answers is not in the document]
michaljurecko: Ok, if you call this a `manual lookup` then it fits. I take it that the  `http.Transport` to do the lookup, so you don't do any manual lookup, you just set the Transport.

>  I'm doing is moving the manual lookup elsewhere because I can't render the reloading page from inside the transport.
So this is probably the problematic part. You can't and shouldn't do that in the Transport.

You should implement a custom PageWriter and catch the error there (see Slack discussion).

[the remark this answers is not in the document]
michaljurecko: See :two: https://keboolaglobal.slack.com/archives/C055WKPHYSD/p1711014587799899?thread_ts=1711008295.446849&cid=C055WKPHYSD

internal/pkg/service/appsproxy/http/transport.go
```diff
@@ -0,0 +1,67 @@
...
+func NewDNSSkippingHTTPTransport(addressCtxKey any) (*http.Transport, error) {
+	dialer := dns.NewDialer()
+
+	transport := http.DefaultTransport.(*http.Transport).Clone()
+
+	transport.DialContext = func(ctx context.Context, network, addr string) (net.Conn, error) {
+		address, ok := ctx.Value(addressCtxKey).(string)
+		if ok {
+			addr = address
+		}
```
michaljurecko, line 45: @jachym-tousek-keboola We probably did not understand each other here, `dialer` or `dial callback` should be used here. 

It should be pass to the `New...` function as an argument.

And then set to the transport, something like:
```
transport.DialContext = dialer.DialContext
```
Or:
```
transport.DialContext = dialContextFn
```

michaljurecko: When I write it verbally:
```
HTTP transport, use <THIS> method to resolve DNS and create a new TCP connection.
```

internal/pkg/service/appsproxy/dns/dialer.go
```diff
@@ -0,0 +1,23 @@
...
+func NewDialer() *net.Dialer {
+	return &net.Dialer{
+		Timeout:   DialTimeout,
+		KeepAlive: KeepAlive,
+		Resolver: &net.Resolver{
+			PreferGo:     true,
+			StrictErrors: true,
+		},
+	}
+}
```
michaljurecko, line 23: This dialer is used internally by the DNS client, to create connection to the DNS server, it sholdn't be public.
Also, the `NewClient` function may be private.

You need only a one public function:  `dns.NewDialer` / `dns.NewDialContextFn()` or somethind similar.

It can be confusing that there are 2 different dialers. One is used by the DNS client and another by the HTTP transport.

internal/pkg/service/appsproxy/dataapps/appconfig.go
```diff
@@ -97,6 +98,17 @@ func (l *sandboxesServiceClient) Notify(ctx context.Context, appID string) error
 	return nil
 }
 
+func (l *sandboxesServiceClient) Wakeup(ctx context.Context, appID string) error {
+	_, err := WakeupApp(l.sender, appID).Send(ctx)
+	if err != nil {
+		l.logger.Errorf(ctx, `Failed sending wakeup request to Sandboxes Service about for app "%s": %s`, appID, err.Error())
```
michaljurecko, line 104: `Failed ` -> `failed `

internal/pkg/service/appsproxy/dns/client.go
```diff
@@ -0,0 +1,93 @@
...
+const DNSReadTimeout = 2 * time.Second
+
+const DNSWriteTimeout = 2 * time.Second
+
+type Client struct {
+	client    *dns.Client
+	dnsServer string
+}
+
+func NewClientFromEtc(dialer *net.Dialer) (*Client, error) {
```
michaljurecko, line 27: :+1:

internal/pkg/service/appsproxy/dns/client.go
```diff
@@ -0,0 +1,93 @@
...
+	// DNS client
+	client := &dns.Client{
+		Net:          "udp",
+		Dialer:       dialer,
+		DialTimeout:  DialTimeout,
+		ReadTimeout:  DNSReadTimeout,
+		WriteTimeout: DNSWriteTimeout,
+	}
+
+	c := &Client{
```
michaljurecko, line 57: `c` and `client` variables :) In this situation you can embedded `client` inside the struct definition.

And `error` is not used at all.


```
return &Client{
  dnsServer: dnsServer,
  client: &dns.Client{
	Net:          "udp",
	Dialer:       dialer,
	DialTimeout:  DialTimeout,
	ReadTimeout:  DNSReadTimeout,
	WriteTimeout: DNSWriteTimeout, 
  },
}
```

internal/pkg/service/appsproxy/http/pagewriter.go
```diff
@@ -0,0 +1,50 @@
...
+
+func (pw *pageWriter) ProxyErrorHandler(w http.ResponseWriter, req *http.Request, err error) {
+	var dnsErr *net.DNSError
+	if errors.As(err, &dnsErr) && dnsErr.IsNotFound {
+		pw.dnsErrorHandler(w, req)
+		return
+	}
+
+	pw.parentWriter.ProxyErrorHandler(w, req, err)
+}
```
michaljurecko, line 50: :+1:  :100:

michaljurecko (approved):
LGTM

PR #1682, opened 2024-03-27
feat: Add "allowTargetEnv" to "manifest.json", add flag

michaljurecko:
Jira: https://keboola.atlassian.net/browse/PSGO-483

**Changes:**
- Added `kbc init --allow-target-env` flag.
- The flag set `allowTargetEnv: true` field in the `.keboola/manifest.json`.
- The field enables support for optional `KBC_PROJECT_ID` and `KBC_BRANCH_ID` ENVs.
- `project.id`/`branch.id` from the manifest is mapped to the values from ENVs
- Mapping is performed on manifest load an vice-versa on the manifest save.
- There must be exactly one branch in the manifest.
- The `kbc init` interactive mode allows selection of the only one branch too, if `--allow-target-env` is used.

-----------


hosekpeter (approved):
LGTM

PR #1690, opened 2024-04-04
feat: Add a new flag for BigQuery setting to create table command

hosekpeter:
Jira: [PSGO-508](https://keboola.atlassian.net/browse/PSGO-508)

**Changes:**
- added a new flag `options-from` for a specific setting for BigQuery
- added tests

-----------

[PSGO-508]: https://keboola.atlassian.net/browse/PSGO-508?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


internal/pkg/service/cli/cmd/remote/create/table/dialog.go
```diff
@@ -25,9 +27,17 @@ const (
 
 type columnsDefinitionMethod int
 
-func AskCreateTable(args []string, branchKey keboola.BranchKey, allBuckets []*keboola.Bucket, d *dialog.Dialogs, f Flags) (table.Options, error) {
+func AskCreateTable(args []string, branchKey keboola.BranchKey, allBuckets []*keboola.Bucket, d *dialog.Dialogs, f Flags, backends []string) (table.Options, error) {
 	opts := table.Options{}
 
+	if f.OptionsFrom.IsSet() && f.ColumnsFrom.IsSet() && slices.Contains(backends, BackendBigQuery) {
```
hosekpeter, line 33: flag `options-from` have to be combined with `columns-from` flag, because BigQuery settings must have a specific column type.

michaljurecko: :one: `IsSet` should be used only to detect if a dialog should be displayed or not, do you remember the last incident?

`--options-from ""` shoudl mean "no options"

michaljurecko: > flag `options-from` have to be combined with `columns-from` flag, because BigQuery settings must have a specific column type.

You are right, but it should be an error, not shouldn't silent ignore it.

michaljurecko:
(please rename the PR)

internal/pkg/service/cli/cmd/remote/create/table/dialog.go
```diff
@@ -30,7 +30,15 @@ type columnsDefinitionMethod int
 func AskCreateTable(args []string, branchKey keboola.BranchKey, allBuckets []*keboola.Bucket, d *dialog.Dialogs, f Flags, backends []string) (table.Options, error) {
 	opts := table.Options{}
 
-	if f.OptionsFrom.IsSet() && f.ColumnsFrom.IsSet() && slices.Contains(backends, BackendBigQuery) {
+	if f.OptionsFrom.Value != "" {
+		if !f.ColumnsFrom.IsSet() {
```
michaljurecko, line 34: Same problem here, `--columns-from ""` means no columns.

internal/pkg/service/cli/cmd/remote/create/table/dialog.go
```diff
@@ -30,7 +30,15 @@ type columnsDefinitionMethod int
...
+			return opts, errors.Errorf("columns-from must be set for bigquery settings")
+		}
+
+		if !slices.Contains(backends, BackendBigQuery) {
+			return opts, errors.Errorf("backend have to be `bigquery`")
+		}
```
michaljurecko, line 40: `"backend have to be `bigquery`"` why?

`"backend have to be `bigquery` to use --options-from"

---------

I prefer to swap the conditions, if the backend is not `bigquery` other checks have no sense.

michaljurecko: And please use the constant:

```
fmt.Sprintf(`backend have to be "bigquery" to use --options-from flag`)
```

hosekpeter:
why aren't comments written all at once so I don't have to correct them 1000 times?

michaljurecko:
@hosekpeter 

I sent you the second batch of comments proactively, as soon as you sent a new commit, where I saw that you fixed one `IsSet` but not the other. So I couldn't comment it sooner.

But yes, I could set the second batch at once, sorry.

test/cli/help/remote-create-table/expected-stdout
```diff
@@ -34,6 +34,7 @@ Flags:
       --columns strings            comma-separated list of column names
       --columns-from string        the path to the table definition file in json
       --name string                name of the table (required if the tableId argument is empty)
+      --options-from string        the path to the set of specific settings for bigquery
```
michaljurecko, line 37: Suggestions:

`--columns-from string        the path to the table definition file in json` 
- :arrow_right: `--columns-from string        the path to the columns definition file in json`


-------------------------

`--options-from string        the path to the set of specific settings for bigquery` 
- :arrow_right: `--options-from string        the path to the table definition file with backed specific options`

michaljurecko (approved):
LGTM

I'm really sorry that I found one more thing there. I preferred not to block you, rather than looking at it in detail for the first time (I had to deal with some important planning issues adhoc).

Next time, would you prefer to have a review in the morning, but all-in-one, or can I send you the biggest problems right away?

michaljurecko (approved):

PR #1692, opened 2024-04-08
feat: Add a new error page for app proxy

hosekpeter:
Jira: [PSGO-515](https://keboola.atlassian.net/browse/PSGO-515)

**Changes:**
- added a new error page file `error.html.tmpl`
- updated tests

-----------

[PSGO-515]: https://keboola.atlassian.net/browse/PSGO-515?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


internal/pkg/service/appsproxy/http/router.go
```diff
@@ -174,9 +174,9 @@ func (r *Router) Shutdown() {
 func (r *Router) createConfigErrorHandler(exceptionID string) http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, req *http.Request) {
 		r.logger.With(attribute.String("exceptionId", exceptionID)).Warn(req.Context(), `application "<proxy.appid>" has misconfigured OAuth2 provider`)
+		appID, _ := req.Context().Value(AppIDCtxKey).(string)
```
hosekpeter, line 177: Adding the app name will be solved in the next task -> [here](https://keboola.atlassian.net/browse/PSGO-499)

internal/pkg/service/appsproxy/http/router.go
```diff
@@ -715,3 +721,58 @@ func headerFromClaim(header, claim string) options.Header {
 		},
 	}
 }
+
+func (r *Router) getHTMLPage(ctx context.Context, w http.ResponseWriter, imageFilePath string, htmlFilePath string, exceptionID string, appID string) {
```
hosekpeter, line 725: Renamed, because this function has a logic that will be used for `spinner page`

internal/pkg/service/appsproxy/http/router.go
```diff
@@ -715,3 +721,58 @@ func headerFromClaim(header, claim string) options.Header {
...
+	// Encode the GIF data to Base64.
+	imageBase64, err := encodeImageDataToBase64(imageFilePath)
```
hosekpeter, line 727: The `Inline` base64 to html is too long, so we will parse it in the code and then insert it into the template as a placeholder

michaljurecko:
**Suggestions:**

**1.** 

Is `base64` enconding required? Why? The images can have their own url, thr browser will cache them.
- `_proxy/favicon.ico`
- `_proxy/foo.svg`

If `base64` will be used, it should be cached, as Jachym wrote.

--------------------------------------

**2.** 

Error handling should be same and set for all use-cases.
I suggest, to modify all error page rendering at first,
to use `pageWriter.WriteErrorPage`

The page writer is used then:
https://github.com/keboola/keboola-as-code/blob/01ebbd4a4089221edcf4e7edce6190fbf6e1f548/internal/pkg/service/appsproxy/http/router.go#L436-L437

See the `PageWriter` interface:
https://github.com/oauth2-proxy/oauth2-proxy/blob/fc701bfd6a992bb074671b91ac89b50a79d3dc3d/pkg/app/pagewriter/pagewriter.go#L8-L18

Then, the template in the `PageWriter` should be modified, to match our design.

------------------------------------

**3.**

Please, keep [Single responsibility principle](https://en.wikipedia.org/wiki/Single_responsibility_principle).

The `Router` shouldn't have `getHTMLPage` and similar methods. 
Use please `PageWriter` or create something new (new package or a new struct for these stuffs).

michaljurecko:
> PageWriter is specific to the oauth-proxy library but we need error pages for public apps as well.

Is there a problem, that causes the `PageWriter` cannot be used?

`PageWriter.WriteErrorPage`:
I think it looks universal, we can leave the Sign In url empty -> don't display it.

If there is a blocker, there should still be one `ErrorPageWriter` that we can call for both public and authorized apps.
We don't want to keep the logic and template in different places.

-------------------------_

You (Petr+Jachym) can schedule a call if needed and we can discuss it.

internal/pkg/service/appsproxy/http/router.go
```diff
@@ -55,16 +54,15 @@ type Router struct {
...
-	errorPagePath = "template/error.html.tmpl"
-	faviconPath   = "template/favicon.ico"
+	faviconPagePath   = "/_proxy/favicon.ico"
+	imagePagePath     = "/_proxy/rocket-flying.gif"
 )
 
-//go:embed template/*
-var templates embed.FS
+//go:embed template/* static/*
+var content embed.FS
```
michaljurecko, line 62: @hosekpeter  (cc @jachym-tousek-keboola) 

Here are some problems:

1. You are mixing together internal files `template/*` and public files `static/*`.
2. `static/` prefix of the files path make the implementation more complicated.
3. The main problem is: 

**You should mount all static assets in `mux` with one handler.**

**We don't want to name individual files there, it is unnecessary work.**

How exactly you achieve this is up to you.


----------------------

You should:

1. Create two variables.
```
//go:embed static
var static embed.FS

//go:embed template
var templates embed.FS
```

2. Mount all static files to the `mux`:
```go
import io/fs
...
staticDirFs := fs.Sub(static, "static")
staticHandler := http.FileServer(http.FS(staticDirFs))
mux.Handle("/_proxy/", http.StripPrefix("/_proxy/", staticHandler))
```

Read more about `fs.Sub`: https://www.iamyadav.com/blogs/a-guide-to-embedding-static-files-in-go

(maybe there are some typos in the code, I haven't tried it, I just wanted to point you in the right direction)

PR #1700, opened 2024-04-15
tests: Rename MustReplaceEnvsDir to ReplaceEnvsDir. Return error instead of panicing.

Matovidlo:
Jira: [540](https://keboola.atlassian.net/browse/PSGO-540)

**Changes:**
- Change usage of MustReplaceEnvsDir to ReplaceEnvsDir.
- Return error instead of panic.
- Use require.NoError for catching newly returned error.

-----------


michaljurecko (approved):
LGTM

PR #1701, opened 2024-04-15
Refactor dataapps/* packages

michaljurecko:
Jira: https://keboola.atlassian.net/browse/PSGO-541

**Changes:**
- Refactored `dataapps/*` packages to make it more in line with the Single Responsibility Principle.
- (Further changes in other PRs).

---------------

Before:
image

---------------

After:
image

-----------

PR #1708, opened 2024-04-18
Data Apps Proxy: Refactor error handling

michaljurecko:
Jira: https://keboola.atlassian.net/browse/PSGO-541

**Changes:**
- Refactored Data Apps Proxy
  - Splitted to more pkgs/files.
  - Improved composition of all components/dependencies.
  - Error handling and pages rendering is in one place.
- Fixed issues/bugs found during refactoring. 
- Added unit tests.

-----------

:information_source: Commented per commit.


cmd/apps-proxy/main.go
```diff
@@ -91,18 +91,14 @@ func run(ctx context.Context, cfg config.Config, _ []string) error {
 	oautproxylogger.SetErrOutput(loggerWriter)
 
 	// Create dependencies
-	scope, err := dependencies.NewServiceScope(ctx, cfg, proc, logger, tel, os.Stdout, os.Stderr) // nolint:forbidigo
+	d, err := dependencies.NewServiceScope(ctx, cfg, proc, logger, tel, os.Stdout, os.Stderr) // nolint:forbidigo
```
Matovidlo, line 94:
```suggestion
	ss, err := dependencies.NewServiceScope(ctx, cfg, proc, logger, tel, os.Stdout, os.Stderr) // nolint:forbidigo
```
This looks more like service scope then `d`

internal/pkg/service/appsproxy/dataapps/api/config.go
```diff
@@ -38,6 +46,22 @@ func (c AppID) String() string {
...
+		return c.Name + "-" + c.ID.String()
+	}
+}
+
+func (c AppConfig) Domain() string {
+	if c.Name == "" {
+		return c.ID.String()
+	} else {
+		return c.Name + "-" + c.ID.String()
+	}
```
Matovidlo, line 62:
```suggestion
	if c.Name == "" {
		return c.ID.String()
	}
	
	return c.Name + "-" + c.ID.String()
```

internal/pkg/service/appsproxy/dataapps/api/rule.go
```diff
@@ -32,15 +32,20 @@ type Rule struct {
 	AuthRequired *bool         `json:"authRequired"`
 }
 
-func (r *Rule) RegisterHandler(mux *http.ServeMux, handler http.Handler) error {
+func (r *Rule) Match(req *http.Request) (bool, error) {
```
Matovidlo, line 35: Do you need `bool`. Just by inspecting i see, `err == nil` is exactly what bool matches.

internal/pkg/service/appsproxy/proxy/apphandler/apphandler.go
```diff
@@ -0,0 +1,145 @@
...
+	defer span.End(&err)
+
+	req = req.WithContext(ctx)
+
+	// Delete all X-KBC-* headers as they're used for internal information.
+	for name := range req.Header {
+		if strings.HasPrefix(strings.ToLower(name), "x-kbc-") {
+			req.Header.Del(name)
+		}
+	}
```
Matovidlo, line 115: This looks like nice helper/util function. Isnt it used multiple times in code?

internal/pkg/service/appsproxy/proxy/apphandler/authproxy/handler.go
```diff
@@ -0,0 +1,128 @@
...
+}
+
+func (h *Handler) ServeHTTPOrError(w http.ResponseWriter, req *http.Request) error {
+	if h.initErr != nil {
+		return h.initErr
+	}
+
+	h.proxyHandler.ServeHTTP(w, req) // errors are handled by the page writer
+	return nil
+}
```
Matovidlo, line 122:
```suggestion

	return h.proxyConfig.Cookie.Expire
}

func (h *Handler) SignInPath() string {
	if h.initErr != nil {
		return "/error"
	}
	
	return h.proxyHandler.SignInPath
}

func (h *Handler) ServeHTTPOrError(w http.ResponseWriter, req *http.Request) error {
	if h.initErr != nil {
		return h.initErr
	}

	h.proxyHandler.ServeHTTP(w, req) // errors are handled by the page writer
	return nil
}
```
code looks consistent with spacings :)

internal/pkg/service/appsproxy/proxy/apphandler/authproxy/pagewriter.go
```diff
@@ -0,0 +1,82 @@
...
+		switch opts.Status {
+		case http.StatusForbidden:
+			messages = []string{"You do not have permission to access this resource."}
+		case http.StatusUnauthorized:
+			messages = []string{"You need to be logged in to access this resource."}
+		case http.StatusInternalServerError:
+			messages = []string{"Internal Server Error Oops! Something went wrong."}
+		default:
+			messages = []string{opts.AppError}
+		}
```
Matovidlo, line 70: Doesn't it makes sense to provide this message as default? It would be included when no `opts.Status` is present, therefore you don't need to do this case and by default there is message. In the docs just include that when no `opts.Status` is defined the messages references to `opts.AppError`.

internal/pkg/service/appsproxy/proxy/apphandler/authproxy/selector.go
```diff
@@ -0,0 +1,181 @@
...
+	// Shortcut, if there is only one provider
+	if len(handlers) == 1 {
+		for _, handler := range handlers {
+			// Set cookie if needed
+			if providerID := s.providerIDFromCookie(req); providerID != handler.Provider().ID() {
+				s.setCookie(w, req, handler)
+			}
+			return handler.ServeHTTPOrError(w, req)
+		}
+	}
```
Matovidlo, line 68: Is this shortcut very performant? If yes just resolve the comment, however the code will get thiner and therefore there is less possibilities to make mistake, because each handler goes trough same code, here is exception that when single handler provided it does different flow.

internal/pkg/service/appsproxy/proxy/apphandler/authproxy/selector.go
```diff
@@ -0,0 +1,181 @@
...
+		Secure:   true,
+		HttpOnly: true,
+		SameSite: http.SameSiteLaxMode,
+	}
+
+	if expires > 0 {
+		v.Expires = s.clock.Now().Add(expires)
+	} else {
+		v.MaxAge = -1
+	}
```
Matovidlo, line 178:
```suggestion
	v := &http.Cookie{
		Name:     providerCookie,
		Value:    value,
		Path:     "/",
		Domain:   requtil.Host(req),
		Secure:   true,
		HttpOnly: true,
		SameSite: http.SameSiteLaxMode,
		MaxAge: -1,
	}

	if expires > 0 {
	        v.MaxAge = 0
		v.Expires = s.clock.Now().Add(expires)
	}
```
Maybe not the best but the else looks weird. Also it could be reverted. (if expires <= 0)

internal/pkg/service/appsproxy/proxy/apphandler/upstream/upstream.go
```diff
@@ -0,0 +1,214 @@
...
+	return upstream, nil
+}
+
+func (u *AppUpstream) ServeHTTPOrError(rw http.ResponseWriter, req *http.Request) error {
+	// Difference between regular and websocket request
+	if strings.EqualFold(req.Header.Get("Connection"), "upgrade") && req.Header.Get("Upgrade") == "websocket" {
+		return u.wsHandler.ServeHTTPOrError(rw, req)
+	} else {
+		return u.handler.ServeHTTPOrError(rw, req)
+	}
```
Matovidlo, line 116:
```suggestion
	}
	
	return u.handler.ServeHTTPOrError(rw, req)
```

internal/pkg/service/appsproxy/proxy/approuter/approuter.go
```diff
@@ -0,0 +1,75 @@
...
+	// Get subdomain
+	subdomain := domain[:strings.IndexByte(domain, '.')]
+
+	// Remove optional app name prefix, if any
+	lastDash := strings.LastIndexByte(subdomain, '-')
+	if lastDash >= 0 {
+		return api.AppID(subdomain[lastDash+1:]), true
+	} else {
+		return api.AppID(subdomain), true
+	}
```
Matovidlo, line 74:
```suggestion
	}
	
	return api.AppID(subdomain), true
```

internal/pkg/service/appsproxy/proxy/pagewriter/error.go
```diff
@@ -0,0 +1,111 @@
...
+
+func (pw *Writer) ProxyErrorHandler(w http.ResponseWriter, req *http.Request, err error) {
+	var dnsError *net.DNSError
+	if errors.As(err, &dnsError) {
+		pw.logger.Info(req.Context(), "app is not running, rendering spinner page")
+		pw.WriteSpinnerPage(w, req)
+	} else {
+		err = svcerrors.NewBadGatewayError(err).WithUserMessage("Request to application failed.")
+		pw.WriteError(w, req, err)
+	}
```
Matovidlo, line 28:
```suggestion
	}
	
	err = svcerrors.NewBadGatewayError(err).WithUserMessage("Request to application failed.")
	pw.WriteError(w, req, err)
```

internal/pkg/service/appsproxy/proxy/pagewriter/error.go
```diff
@@ -0,0 +1,111 @@
...
+
+	// Error name, by default "internal"
+	var errName string
+	var errNameProvider svcerrors.WithName
+	if errors.As(err, &errNameProvider) {
+		errName = errNameProvider.ErrorName()
+	} else {
+		errName = "internal"
+	}
```
Matovidlo, line 46:
```suggestion
	var errName := "internal"
	var errNameProvider svcerrors.WithName
	if errors.As(err, &errNameProvider) {
		errName = errNameProvider.ErrorName()
	}
```

internal/pkg/service/appsproxy/proxy/pagewriter/error.go
```diff
@@ -0,0 +1,111 @@
...
+	var logMsgProvider svcerrors.WithLogMessage
+	if errors.As(err, &logMsgProvider) {
+		logMessage = logMsgProvider.ErrorLogMessage()
+	} else {
+		logMessage = errors.Format(err, errors.FormatWithUnwrap(), errors.FormatWithStack())
+		if errName != "" {
+			logMessage = errName + ": " + logMessage
+		}
+	}
+
```
Matovidlo, line 73:
```suggestion
	}
	
	logMessage = errors.Format(err, errors.FormatWithUnwrap(), errors.FormatWithStack())
	if errName != "" {
		logMessage = errName + ": " + logMessage
	}

```

internal/pkg/service/appsproxy/proxy/proxy_test.go
```diff
@@ -289,22 +292,23 @@ func TestAppProxyRouter(t *testing.T) {
...
+					assert.Equal(t, http.SameSiteLaxMode, cookies[0].SameSite)
+
+					assert.Equal(t, "_oauth2_proxy", cookies[1].Name)
+					assert.Equal(t, "", cookies[1].Value)
+					assert.Equal(t, "/", cookies[1].Path)
+					assert.Equal(t, "oidc.hub.keboola.local", cookies[1].Domain)
+					assert.True(t, cookies[1].HttpOnly)
+					assert.True(t, cookies[1].Secure)
+					assert.Equal(t, http.SameSiteLaxMode, cookies[1].SameSite)
+				}
```
Matovidlo, line 311:
```suggestion
				assert.Len(t, cookies, 2)
				oauthProviders := []string {
				     "_oauth2_provider",
				     "_oauth2_proxy",
				}
				for i, cookie := range cookies {
					assert.Equal(t, oauthProviders[i], cookie.Name)
					assert.Equal(t, "", cookie.Value)
					assert.Equal(t, "/", cookie.Path)
					assert.Equal(t, "oidc.hub.keboola.local", cookie.Domain)
					assert.True(t, cookie.HttpOnly)
					assert.True(t, cookie.Secure)
					assert.Equal(t, http.SameSiteLaxMode, cookie.SameSite)
				}
```
Could be copied also below in another tests. When you add cookie you copy boilerplate. I think this is just improvement for future cookie extension and also prescription how to test multiple cookies in future.

internal/pkg/service/appsproxy/proxy/proxy_test.go
```diff
@@ -493,34 +504,35 @@ func TestAppProxyRouter(t *testing.T) {
...
+					assert.True(t, cookies[0].Secure)
+					assert.Equal(t, http.SameSiteLaxMode, cookies[0].SameSite)
+
+					assert.Equal(t, "_oauth2_proxy_csrf", cookies[1].Name)
+					assert.Equal(t, "/", cookies[1].Path)
+					assert.Equal(t, "oidc.hub.keboola.local", cookies[1].Domain)
+					assert.True(t, cookies[1].HttpOnly)
+					assert.True(t, cookies[1].Secure)
+					assert.Equal(t, http.SameSiteNoneMode, cookies[1].SameSite)
+				}
```
Matovidlo, line 532: same as above ^

internal/pkg/service/appsproxy/proxy/proxy_test.go
```diff
@@ -558,32 +570,34 @@ func TestAppProxyRouter(t *testing.T) {
...
```
Matovidlo, line 591: same as above ^ just adjust the values, not only providers. Or create cookies type object so you deep equal the fields? This does not look good. Reflection also solves possible changes in `cookie` structure etc.

internal/pkg/service/appsproxy/proxy/proxy_test.go
```diff
@@ -615,69 +629,78 @@ func TestAppProxyRouter(t *testing.T) {
...
 				cookies := response.Cookies()
+				if assert.Len(t, cookies, 1) {
+					assert.Equal(t, "_oauth2_provider", cookies[0].Name)
+					assert.Equal(t, "oidc1", cookies[0].Value)
+					assert.Equal(t, "/", cookies[0].Path)
+					assert.Equal(t, "multi.hub.keboola.local", cookies[0].Domain)
+					assert.True(t, cookies[0].HttpOnly)
+					assert.True(t, cookies[0].Secure)
+					assert.Equal(t, http.SameSiteLaxMode, cookies[0].SameSite)
+				}
```
Matovidlo, line 676: same as above ^. Possibly this could be some helper function or deep reflection using assert equal of 2 objects.

internal/pkg/service/appsproxy/proxy/proxy_test.go
```diff
@@ -615,69 +629,78 @@ func TestAppProxyRouter(t *testing.T) {
...
+				assert.Contains(t, location, "/oidc/authorize?client_id=")
+				cookies = response.Cookies()
+				if assert.Len(t, cookies, 1) {
+					assert.Equal(t, "_oauth2_proxy_csrf", cookies[0].Name)
+					assert.Equal(t, "/", cookies[0].Path)
+					assert.Equal(t, "multi.hub.keboola.local", cookies[0].Domain)
+					assert.True(t, cookies[0].HttpOnly)
+					assert.True(t, cookies[0].Secure)
+					assert.Equal(t, http.SameSiteNoneMode, cookies[0].SameSite)
+				}
```
Matovidlo, line 694: same as above ^

internal/pkg/service/appsproxy/proxy/proxy_test.go
```diff
@@ -1585,32 +1728,34 @@ func TestAppProxyRouter(t *testing.T) {
...
+					assert.True(t, cookies[0].Secure)
+					assert.Equal(t, http.SameSiteLaxMode, cookies[0].SameSite)
+
+					assert.Equal(t, "_oauth2_proxy_csrf", cookies[1].Name)
+					assert.Equal(t, "/", cookies[1].Path)
+					assert.Equal(t, "oidc.hub.keboola.local", cookies[1].Domain)
+					assert.True(t, cookies[1].HttpOnly)
+					assert.True(t, cookies[1].Secure)
+					assert.Equal(t, http.SameSiteNoneMode, cookies[1].SameSite)
+				}
```
Matovidlo, line 1749: same as above ^

internal/pkg/service/appsproxy/proxy/server.go
```diff
@@ -21,10 +19,11 @@ const (
 	gracefulShutdownTimeout = 30 * time.Second
 )
 
-func StartServer(ctx context.Context, d dependencies.ServiceScope, router http.Handler) error {
-	logger, tel, cfg := d.Logger(), d.Telemetry(), d.Config()
+func StartServer(ctx context.Context, d dependencies.ServiceScope) error {
```
Matovidlo, line 22:
```suggestion
func StartServer(ctx context.Context, d dependencies.ServiceScope) error {
```
```suggestion
func StartServer(ctx context.Context, ss dependencies.ServiceScope) error {
```

internal/pkg/service/appsproxy/proxy/server.go
```diff
@@ -53,25 +52,33 @@ func StartServer(ctx context.Context, d dependencies.ServiceScope, router http.H
 	return nil
 }
 
-func newHandler(logger log.Logger, tel telemetry.Telemetry, router http.Handler, publicURL *url.URL) http.Handler {
+func NewHandler(d dependencies.ServiceScope) http.Handler {
```
Matovidlo, line 55:
```suggestion
func NewHandler(ss dependencies.ServiceScope) http.Handler {
```

internal/pkg/service/appsproxy/proxy/server_test.go
```diff
@@ -0,0 +1,204 @@
...
+	// Send ignored request
+	rec = httptest.NewRecorder()
+	req = httptest.NewRequest("GET", "https://hub.keboola.local/health-check", nil)
+	req.Header.Set("User-Agent", "my-user-agent")
+	handler.ServeHTTP(rec, req)
+	require.Equal(t, http.StatusOK, rec.Code)
+	assert.Equal(t, "OK\n", rec.Body.String())
+
+	assert.NotNil(t, mocked.DebugLogger().CompareJSONMessages(`{"level":"info","message":"req https://123.hub.keboola.local/health-check %A"}`))
+	mocked.DebugLogger().AssertJSONMessages(t, `{"level":"info","message":"req https://123.hub.keboola.local/path %A","http.request_id":"%s","component":"http"}`)
```
Matovidlo, line 108: This looks like some table tests, 5 test cases, similar code, creating recorder request etc. I dont say that you need to do it now, but It would be fine if it is refactored as table tests. It also prevents writing commentaries which are part of table test `ID`.

Matovidlo (changes requested):
I think some of the things are not part of the refactor, however I would do the renaming and fix the cookie tests. It looks like big boilerplate. Consider both deepReflection of `expectedCookie` vs `actualCookie` or just sample as I provider (test changes one by one, e.g oauthProvider). Overall I did not find many problems.

michaljurecko:
@Matovidlo I haven't asked for a review yet :) 

If someone would send me this for review, in its current state, I would reject it right away.

cmd/apps-proxy/main.go
michaljurecko: There is our convenction to use `d`, as `dependencies`, the scopes (dependency containers) have different names, we can discuss it more.

internal/pkg/service/appsproxy/dataapps/api/config.go
```diff
@@ -38,6 +46,22 @@ func (c AppID) String() string {
 	return string(c)
 }
 
+func (c AppConfig) IdAndName() string {
+	if c.Name == "" {
+		return c.ID.String()
+	} else {
+		return c.Name + "-" + c.ID.String()
+	}
```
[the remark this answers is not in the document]
michaljurecko: :heavy_check_mark:

internal/pkg/service/appsproxy/dataapps/api/config.go
michaljurecko: :heavy_check_mark:

internal/pkg/service/appsproxy/dataapps/api/rule.go
michaljurecko: No, there is a difference, I am adding tests for it. 

Error means, the rule is invalid.

`false` means that the rude doesn't match the path.

internal/pkg/service/appsproxy/proxy/apphandler/apphandler.go
michaljurecko: no, it is used once

internal/pkg/service/appsproxy/proxy/apphandler/authproxy/selector.go
michaljurecko: It is logic shortcut, if there is one handler, no selector page is needed, it is not performance shortcut.

michaljurecko: I will update comment.

internal/pkg/service/appsproxy/proxy/apphandler/authproxy/selector.go
michaljurecko: It better match the logic, ... if there is an expiration ... set it ... else .... clear the cookie. I will add comment.

internal/pkg/service/appsproxy/proxy/approuter/approuter.go
michaljurecko: You can setup linter for it, if you want.

internal/pkg/service/appsproxy/proxy/pagewriter/error.go
michaljurecko: It doesn;t match style of the other blocks, default value may require some logic/code.

internal/pkg/service/appsproxy/proxy/proxy_test.go
michaljurecko: Yes, the test may be improved, I did not make big changes there so that changes in behavior could be seen.

internal/pkg/service/appsproxy/proxy/apphandler/authproxy/handler.go
michaljurecko: I don't think it's necessary, if you want, find a linter that will process it. But whether it is 3 lines or 4 seems unimportant to me, for a short code.

internal/pkg/service/appsproxy/proxy/apphandler/authproxy/pagewriter.go
michaljurecko: Please explain it better, I don't understand.

internal/pkg/service/appsproxy/proxy/pagewriter/error.go
```diff
@@ -0,0 +1,117 @@
...
+	var dnsError *net.DNSError
+	if errors.As(err, &dnsError) {
+		pw.logger.Info(req.Context(), "app is not running, rendering spinner page")
+		pw.WriteSpinnerPage(w, req)
+	}
+
+	pw.WriteError(w, req, svcerrors.NewBadGatewayError(err).WithUserMessage("Request to application failed."))
+}
+
+func (pw *Writer) WriteError(w http.ResponseWriter, req *http.Request, err error) {
```
Matovidlo, line 32: Overall I would split this function into smaller functions if possible. It is quite long, could be separated by error or by status reported.

internal/pkg/service/appsproxy/proxy/apphandler/authproxy/selector.go
```diff
@@ -0,0 +1,183 @@
...
+	// Skip selector page, if there is only one provider
+	if len(handlers) == 1 {
+		for _, handler := range handlers {
+			// Set cookie if needed
+			if providerID := s.providerIDFromCookie(req); providerID != handler.Provider().ID() {
+				s.setCookie(w, req, handler)
+			}
+			return handler.ServeHTTPOrError(w, req)
+		}
+	}
```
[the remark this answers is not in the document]
Matovidlo: Something like `SkipSelector()` Which make sense, when 1 handler has to be present  do this.

internal/pkg/service/appsproxy/proxy/testutil/api.go
```diff
@@ -0,0 +1,95 @@
...
+	ts := httptest.NewUnstartedServer(mux)
+	ts.EnableHTTP2 = true
+	ts.Start()
+
+	service.Server = ts
+
+	return service
+}
+
+func (v *DataAppsAPI) Register(apps []api.AppConfig) {
```
Matovidlo, line 91:
```suggestion
func (service *DataAppsAPI) Register(apps []api.AppConfig) {
```
As in New.

internal/pkg/service/appsproxy/proxy/apphandler/authproxy/selector.go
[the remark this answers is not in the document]
Matovidlo: The `for` loop is weird it actually does 1 iteration.

Matovidlo (changes requested):

internal/pkg/service/appsproxy/dependencies/mocked.go
```diff
@@ -7,20 +7,26 @@ import (
...
-	config config.Config
+	config    config.Config
+	dnsServer *dnsmock.Server
 }
 
 func (v *mocked) TestConfig() config.Config {
 	return v.config
 }
 
+func (v *mocked) TestDNSServer() *dnsmock.Server {
```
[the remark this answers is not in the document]
michaljurecko: :arrow_right: let's discuss it outside this PR, please add link to the comment to the [your document](https://keboola.atlassian.net/wiki/spaces/ENGG/pages/3301507091/TMP+Martin+Va+ko+Onboarding+Suggestions).

internal/pkg/service/appsproxy/proxy/pagewriter/error.go
michaljurecko: :arrow_right: let's discuss it outside this PR, please add link to the comment to the [your document](https://keboola.atlassian.net/wiki/spaces/ENGG/pages/3301507091/TMP+Martin+Va+ko+Onboarding+Suggestions).

internal/pkg/service/appsproxy/proxy/testutil/api.go
michaljurecko: :arrow_right: let's discuss it outside this PR, please add link to the comment to the [your document](https://keboola.atlassian.net/wiki/spaces/ENGG/pages/3301507091/TMP+Martin+Va+ko+Onboarding+Suggestions).

internal/pkg/service/appsproxy/proxy/pagewriter/error.go
michaljurecko: :arrow_right: let's discuss it outside this PR, please add link to the comment to the [your document](https://keboola.atlassian.net/wiki/spaces/ENGG/pages/3301507091/TMP+Martin+Va+ko+Onboarding+Suggestions).

internal/pkg/service/appsproxy/proxy/apphandler/authproxy/selector.go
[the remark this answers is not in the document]
michaljurecko: Ad `for`, ... the `handlers` is a map, you cannot get easily the first item from the map.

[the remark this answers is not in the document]
michaljurecko: Other points: :arrow_right: let's discuss it outside this PR, please add link to the comment to the [your document](https://keboola.atlassian.net/wiki/spaces/ENGG/pages/3301507091/TMP+Martin+Va+ko+Onboarding+Suggestions).

internal/pkg/service/appsproxy/dependencies/dependencies.go
```diff
@@ -137,10 +167,30 @@ func (v *serviceScope) AppsAPI() *api.API {
 	return v.appsAPI
 }
 
+func (v *serviceScope) AppHandlers() *apphandler.Manager {
```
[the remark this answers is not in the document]
michaljurecko: :arrow_right: let's discuss it outside this PR, please add link to the comment to the [your document](https://keboola.atlassian.net/wiki/spaces/ENGG/pages/3301507091/TMP+Martin+Va+ko+Onboarding+Suggestions).

internal/pkg/service/appsproxy/proxy/apphandler/authproxy/selector.go
```diff
@@ -0,0 +1,183 @@
...
+		SameSite: http.SameSiteLaxMode,
+	}
+
+	if expires > 0 {
+		// If there is an expiration, set it
+		v.Expires = s.clock.Now().Add(expires)
+	} else {
+		// Otherwise clear the cookie
+		v.MaxAge = -1
+	}
```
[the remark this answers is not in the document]
michaljurecko: As I wrote (maybe you missed it in other comments)

> It better match the logic, ... if there is an expiration ... set it ... else .... clear the cookie. I will add comment.

-------------

I really don't understand your request. For me, it is important that the code is 1. functional 2. easy to understand. What I think this code accomplishes.

I have the feeling that you prefer rules over purpose: `if possible, always set the default value, don't use "else", and never otherwise`.

In my opinion, you also need to think about a person, what will be work on it after me/you, and he should notice this important part of the the function.

:arrow_right: so please let's discuss it outside this PR, please add link to the comment to the [your document](https://keboola.atlassian.net/wiki/spaces/ENGG/pages/3301507091/TMP+Martin+Va+ko+Onboarding+Suggestions).

internal/pkg/service/appsproxy/proxy/transport/transport.go
```diff
@@ -0,0 +1,138 @@
...
+		// The DNS query is not recursive and not cached, see "dns" package for details.
+		ip, err := dnsClient.Resolve(resolveCtx, host) //nolint:contextcheck
+
+		// Trace DNSDone
+		if trace != nil && trace.DNSDone != nil {
+			trace.DNSDone(httptrace.DNSDoneInfo{
+				Addrs: []net.IPAddr{{IP: net.ParseIP(ip)}},
+				Err:   err,
+			})
+		}
```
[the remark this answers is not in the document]
michaljurecko: This issue maybe was reason of unexpected `502` errors, ... because we detected the connection to the DNS server as sucessful conection to the app, .. and then it failed in Go reverse proxy. (cc @jachym-tousek-keboola @hosekpeter)

But maybe there are some other issues, we will see after manual testing in the testing stack.

internal/pkg/service/appsproxy/proxy/pagewriter/error.go
```diff
@@ -0,0 +1,118 @@
...
+	"net/http"
+	"strings"
+
+	svcerrors "github.com/keboola/keboola-as-code/internal/pkg/service/common/errors"
+	"github.com/keboola/keboola-as-code/internal/pkg/utils/errors"
+)
+
+const ExceptionIDPrefix = "keboola-appsproxy-"
+
+type errorPageData struct {
```
hosekpeter, line 14: I would add a field "AppName", because we will need that for each page (selector,spinner,error)

michaljurecko: :heavy_check_mark: :hourglass:

michaljurecko: @hosekpeter I create two new PRs to address it:
- https://github.com/keboola/keboola-as-code/pull/1709
- https://github.com/keboola/keboola-as-code/pull/1710

I also added `projectId`, so we don't have to adjust it in some other PR

internal/pkg/service/appsproxy/proxy/server_test.go
```diff
@@ -0,0 +1,623 @@
...
+			},
+		},
+	})
+
+	// Create proxy handler
+	handler := proxy.NewHandler(d)
+
+	// Get robots.txt
+	rec := httptest.NewRecorder()
+	req := httptest.NewRequest("GET", "https://hub.keboola.local/robots.txt", nil)
```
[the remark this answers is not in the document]
michaljurecko: Yes, it is test of the file for the `hub` root size.

internal/pkg/service/appsproxy/proxy/server_test.go
```diff
@@ -0,0 +1,623 @@
...
+	handler.ServeHTTP(rec, req)
+	require.Equal(t, http.StatusOK, rec.Code)
+	assert.Contains(t, rec.Body.String(), "Disallow: /")
```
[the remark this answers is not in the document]
michaljurecko: It is response for the `hub` root site, ... there is no `<appID>.hub` in the URL.

internal/pkg/service/appsproxy/proxy/apphandler/authproxy/pagewriter.go
```diff
@@ -0,0 +1,82 @@
...
+			messages = append(messages, str)
+		}
+	}
+
+	if len(messages) == 0 {
+		switch opts.Status {
+		case http.StatusForbidden:
+			messages = []string{"You do not have permission to access this resource."}
+		case http.StatusUnauthorized:
+			messages = []string{"You need to be logged in to access this resource."}
```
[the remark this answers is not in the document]
michaljurecko: Ok, we can sort it by status code.

internal/pkg/service/appsproxy/dataapps/api/rule.go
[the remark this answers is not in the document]
michaljurecko: Yes @jachym-tousek-keboola but `mux.Handle(<path>)` gives to much power to the user, and instead of preventing misuse of other options, it is better to exactly say what we support.

internal/pkg/service/appsproxy/proxy/server_test.go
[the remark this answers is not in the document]
Matovidlo: So it is expected when you contact the `hub` without application that it returns OK and `Disallow: /` in body if I understand it right. I could not find such use case in RFC.

internal/pkg/service/appsproxy/dataapps/api/rule.go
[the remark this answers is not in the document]
michaljurecko: We already talked a little about it in the first PR: https://github.com/keboola/keboola-as-code/pull/1701#discussion_r1565370885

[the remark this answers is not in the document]
michaljurecko: Simplicity is not touched with it:
https://github.com/keboola/keboola-as-code/blob/cfcdc177a4215665f3831bb6fdfecac22dbf6cad/internal/pkg/service/appsproxy/proxy/apphandler/apphandler.go#L122-L129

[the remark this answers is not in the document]
michaljurecko: Sorry, I should have commented it at the beginning.

internal/pkg/service/appsproxy/proxy/server_test.go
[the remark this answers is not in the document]
michaljurecko: Such details are not covered by the RFC.

`robots.txt` serves to prevent various web robots from scanning our proxy, because it would unnecessarily generate logs and warnings in the logs, I don't see any problem there (cc @jachym-tousek-keboola for check).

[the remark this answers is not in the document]
michaljurecko: For context: I noticed that `OAuth2Proxy` does this, and I realized that it is missing in our top-level handler.

internal/pkg/service/appsproxy/dataapps/api/rule.go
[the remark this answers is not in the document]
michaljurecko: And it also allows use to pass errors from lower to upper levels, I forgot to write that.

internal/pkg/service/appsproxy/proxy/server_test.go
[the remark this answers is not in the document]
Matovidlo: Can we also mention such technical detail in some document to always make sure that these handlers are working correcntly ? Only if it is possible :) . I know how `robots.txt` works but never made it in production so maybe I was not doing it correctly.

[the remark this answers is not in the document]
michaljurecko: :man_shrugging: it's a nice to have thing, if it wasn't there we would solve it probably as soon as we saw strange logs in DataDog.

It seems like common knowledge to me, just like we don't document what status code 200 means.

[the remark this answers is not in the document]
Matovidlo: Ok then I resolve this comment.

internal/pkg/service/appsproxy/proxy/apphandler/authproxy/pagewriter.go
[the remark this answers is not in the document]
michaljurecko: :heavy_check_mark: https://github.com/keboola/keboola-as-code/pull/1708/commits/95f37097f1ae3cbdc540d4c239984bd292b1219e

internal/pkg/service/appsproxy/proxy/proxy_test.go
```diff
@@ -90,110 +89,116 @@ func TestAppProxyRouter(t *testing.T) {
...
-			run: func(t *testing.T, client *http.Client, m []*mockoidc.MockOIDC, appServer *appServer, service *sandboxesService, dnsServer *dnsmock.Server) {
+			run: func(t *testing.T, client *http.Client, m []*mockoidc.MockOIDC, appServer *testutil.AppServer, service *testutil.DataAppsAPI, dnsServer *dnsmock.Server) {
 				// Request to broken app
 				request, err := http.NewRequestWithContext(context.Background(), http.MethodGet, "https://broken.hub.keboola.local/", nil)
 				require.NoError(t, err)
 				response, err := client.Do(request)
 				require.NoError(t, err)
-				require.Equal(t, http.StatusForbidden, response.StatusCode)
+				require.Equal(t, http.StatusServiceUnavailable, response.StatusCode)
 				body, err := io.ReadAll(response.Body)
```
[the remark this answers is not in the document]
michaljurecko: 502 bad gateway: problem with the server in the app itself, e.g. it doesn't run
503 service unavailable: problem with the configuration of the app

hosekpeter (approved):
LGTM

internal/pkg/service/appsproxy/dataapps/api/config.go
```diff
@@ -80,6 +112,11 @@ func (a *API) GetAppConfig(appID AppID, eTag string) request.APIRequest[*AppConf
 			// Add MaxAge
 			result.maxAge = maxAge
 
+			// Sort rules, longest value/path first
+			slices.SortStableFunc(result.AuthRules, func(a, b Rule) int {
+				return strings.Compare(b.Value, a.Value)
+			})
```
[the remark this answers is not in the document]
michaljurecko: @jachym-tousek-keboola it match your original implementation :slightly_smiling_face: 

But good point, we should address it.

> Patterns name fixed, rooted paths, like "/favicon.ico", or rooted subtrees, like "/images/" (note the trailing slash). **Longer patterns take precedence over shorter ones**, so that if there are handlers registered for both "/images/" and "/images/thumbnails/", the latter handler will be called for paths beginning "/images/thumbnails/" and the former will receive requests for any other paths in the "/images/" subtree.

https://stackoverflow.com/a/55216191

internal/pkg/service/appsproxy/proxy/proxy_test.go
```diff
@@ -90,110 +89,116 @@ func TestAppProxyRouter(t *testing.T) {
...
+				assert.Contains(t, string(body), pagewriter.ExceptionIDPrefix)
 			},
 			expectedNotifications: map[string]int{},
 			expectedWakeUps:       map[string]int{},
 		},
 		{
-			name: "empty-providers-array-app",
-			run: func(t *testing.T, client *http.Client, m []*mockoidc.MockOIDC, appServer *appServer, service *sandboxesService, dnsServer *dnsmock.Server) {
+			name: "not-empty-providers-auth-required-false",
+			run: func(t *testing.T, client *http.Client, m []*mockoidc.MockOIDC, appServer *testutil.AppServer, service *testutil.DataAppsAPI, dnsServer *dnsmock.Server) {
```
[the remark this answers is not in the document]
michaljurecko: Thanks, I will check it again.

internal/pkg/service/appsproxy/proxy/proxy_test.go
```diff
@@ -289,22 +299,23 @@ func TestAppProxyRouter(t *testing.T) {
...
-				assert.True(t, cookies[0].Secure)
-
-				assert.Equal(t, "_oauth2_proxy", cookies[1].Name)
-				assert.Equal(t, "", cookies[1].Value)
-				assert.Equal(t, "/", cookies[1].Path)
-				assert.Equal(t, "oidc.hub.keboola.local", cookies[1].Domain)
-				assert.True(t, cookies[1].HttpOnly)
-				assert.True(t, cookies[1].Secure)
-				assert.Equal(t, http.SameSiteLaxMode, cookies[1].SameSite)
+				if assert.Len(t, cookies, 2) {
```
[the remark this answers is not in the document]
michaljurecko: It is also good alternative, but now you can see results of the other asserts, an MAYBE it helps you to detect the problem sooner? I thing it is not a "stop now" case.

internal/pkg/service/appsproxy/proxy/proxy_test.go
```diff
@@ -615,69 +636,78 @@ func TestAppProxyRouter(t *testing.T) {
...
+				assert.Contains(t, string(body), html.EscapeString(`https://multi.hub.keboola.local/_proxy/selection?provider=oidc1&rd=%2Fsome%2Fpath`))
+				assert.Contains(t, string(body), html.EscapeString(`https://multi.hub.keboola.local/_proxy/selection?provider=oidc2&rd=%2Fsome%2Fpath`))
 
-				// Provider selection
-				request, err = http.NewRequestWithContext(context.Background(), http.MethodGet, "https://multi.hub.keboola.local/_proxy/selection?provider=oidc1", nil)
+				// Select provider
+				request, err = http.NewRequestWithContext(context.Background(), http.MethodGet, "https://multi.hub.keboola.local/_proxy/selection?provider=oidc1&rd=%2Fsome%2Fpath", nil)
 				require.NoError(t, err)
 				response, err = client.Do(request)
 				require.NoError(t, err)
```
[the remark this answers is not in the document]
michaljurecko: > Except I'm unsure if there is validation that the callback must start with /. 

To prevent external URLs. In code we have control over it, and `req.url.Path` always start with `/`.

But it protects the user from clicking on some link so  he is not redirected to a fraudulent site after logging in. This is a redirection in our proxy (to the first page, that the user visited), not in the auth provider -> to proxy.

internal/pkg/service/appsproxy/proxy/proxy_test.go
```diff
@@ -709,7 +741,7 @@ func TestAppProxyRouter(t *testing.T) {
 		},
 		{
 			name: "multi-app-selection-page-redirect",
```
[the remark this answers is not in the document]
michaljurecko: Ok, thanks, I am going to revert it back.

internal/pkg/service/appsproxy/proxy/proxy_test.go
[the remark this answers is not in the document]
michaljurecko: > > I generally tried to prefer 4xx rather than 5xx since you told me that 5xx cause alerts but these cases are typically caused by user configuration of the data app.
> 
> Sorry if I informed you wrongly, ... from the user's point of view, it doesn't matter if the error is in the proxy or in the app, it's 5xx, ... we have to distinguish it in our telemetry and alerts.

internal/pkg/service/appsproxy/dataapps/api/config.go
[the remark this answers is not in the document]
michaljurecko: It is part of a changes in Go 1.22 (one more reason to not use it in our case), but `Precedence` seems to be same.
The sentences from stackoverflow are no more there.

```
The pattern syntax and matching behavior of ServeMux changed significantly in Go 1.22. 
```

```
Precedence
If two or more patterns match a request, then the most specific pattern takes precedence. A pattern P1 is more specific than P2 if P1 matches a strict subset of P2’s requests; that is, if P2 matches all the requests of P1 and more. If neither is more specific, then the patterns conflict. There is one exception to this rule, for backwards compatibility: if two patterns would otherwise conflict and one has a host while the other does not, then the pattern with the host takes precedence. If a pattern passed [ServeMux.Handle](https://pkg.go.dev/net/http#ServeMux.Handle) or [ServeMux.HandleFunc](https://pkg.go.dev/net/http#ServeMux.HandleFunc) conflicts with another pattern that is already registered, those functions panic.
```

https://pkg.go.dev/net/http#hdr-Precedence

PR #1711, opened 2024-04-22
feat: Add new templates for data apps

hosekpeter:
Jira: [PSGO-515](https://keboola.atlassian.net/browse/PSGO-515)

**Changes:**
- added new templates 

-----------


michaljurecko (approved):
It doesn't look good graphically, but it works.

We need to consult with the desing/UI team.

But we can iterate on it further.

PR #1722, opened 2024-04-23
feat: jsonnet pool

jachym-tousek-keboola:
Jira:
- https://keboola.atlassian.net/browse/PSGO-544
- https://keboola.atlassian.net/browse/PSGO-545
- https://keboola.atlassian.net/browse/PSGO-546

**Changes:**
- Added jsonnet pool
- Added column renderer
-----------

### DIFF
--- a/internal/pkg/encoding/jsonnet/vm.go
+++ b/internal/pkg/encoding/jsonnet/vm.go
@@ -16,16 +16,16 @@ type VM[T any] struct {
 	lock    sync.Mutex
 	vm      *jsonnet.VM
 	err     error
-	payload T
+	payload *T
 }
 
 // Payload returns the payload of the current evaluation.
 // Can be used by custom functions for request specific values.
-func (vm *VM[T]) Payload() T {
+func (vm *VM[T]) Payload() *T {
 	return vm.payload
 }
 
-func (vm *VM[T]) Evaluate(code string, payload T) (jsonOut string, err error) {
+func (vm *VM[T]) Evaluate(code string, payload *T) (jsonOut string, err error) {
 	if vm.err != nil {
 		return "", vm.err
 	}
@@ -34,6 +34,9 @@ func (vm *VM[T]) Evaluate(code string, payload T) (jsonOut string, err error) {
 	defer vm.lock.Unlock()
 
 	vm.payload = payload
+	defer func() {
+		vm.payload = nil
+	}()
 
 	astNode, err := ToAst(code, "")
 	if err != nil {
--- a/internal/pkg/service/stream/api/receive/jsonnet/jsonnet.go
+++ b/internal/pkg/service/stream/api/receive/jsonnet/jsonnet.go
@@ -39,9 +39,9 @@ const (
 	bodyPathFnName      = "_bodyPath"
 )
 
-func NewPool() *jsonnet.VMPool[*receivectx.Context] {
+func NewPool() *jsonnet.VMPool[receivectx.Context] {
 	return jsonnet.NewVMPool(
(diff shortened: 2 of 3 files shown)


internal/pkg/encoding/jsonnet/vm.go
```diff
@@ -0,0 +1,69 @@
...
+
+	"github.com/keboola/keboola-as-code/internal/pkg/utils/errors"
+)
+
+type VM struct {
+	lock    sync.Mutex
+	vm      *jsonnet.VM
+	err     error
+	payload any
+}
```
michaljurecko, line 20: Is the `encoding/jsonnet` right layer for the pool implementation?

--------------

My point of view:
- forked lib: `Freeze` method :heavy_check_mark: :+1: 
- `encoding/jsonnet`: (almost) no changes are necessary
- `receive/jsonnet`: implementation of the `pool`, wrapped `VM`
   - `payload` may be a specific type, for example `receivectx.Context`
   - if we would need another performant implementation of the Jsonnet, we can repeat this way in another pkg, with another payload type.
   
-----------------

I really like how you designed it, I think there is only one step left ^^^. 

This way we can get rid of retyping `payload` in every function <- that is always a last choice, if another approach cannot be used.

michaljurecko: You can also consider a generic implementation, where the payload will be taken from the generic type `[P any]`.

michaljurecko: > pool moved to receive/jsonnet?
no, generic pool can be in `encoding/jsonnet`

michaljurecko (approved):
LGTM

internal/pkg/encoding/jsonnet/vm.go
```diff
@@ -16,13 +16,13 @@ type VM[T any] struct {
...
 	err     error
-	payload T
+	payload *T
 }
 
 // GetPayload returns the payload of the current evaluation.
 // Can be used by custom functions for request specific values.
 func (vm *VM[T]) GetPayload() T {
-	return vm.payload
+	return *vm.payload
```
michaljurecko, line 25: You should return `*T` if you want to keep original logic, and don't copy the payload.

------------

If you keep `VM[*receive.Context]`, the payload type is `**receive.Context` and it is a little bit weird.
So `VM[receive.Context]` seem to me as a better solution.
```
T = receive.Context
*T = *receive.Context
VM[*receive.Context] - > VM[receive.Context]`
Evaluate(T) -> Evaluate(*T)
Payload() T -> Payload() *T
```

go.mod
```diff
@@ -2,7 +2,7 @@ module github.com/keboola/keboola-as-code
 
 go 1.22.2
 
-replace github.com/google/go-jsonnet v0.20.0 => github.com/keboola/go-jsonnet v0.20.1-0.20240415122724-d526145c4cb8
+replace github.com/google/go-jsonnet v0.20.0 => github.com/keboola/go-jsonnet v0.20.1-0.20240418121644-b48711daf3b7
```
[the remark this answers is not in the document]
michaljurecko: https://github.com/keboola/keboola-as-code/pull/1722/commits/6d2b6c9cb5b45c64b61b7ea21d0fcaf81dc89780#r1580679429

internal/pkg/encoding/jsonnet/vm.go
michaljurecko: :heavy_check_mark: :+1:

michaljurecko (approved):
LGTM

PR #1725, opened 2024-04-24
fix: Disk allocation for MacOS

hosekpeter:
Jira: [PSGO-581](https://keboola.atlassian.net/browse/PSGO-581)

**Changes:**
- support disk allocation for MacOS

-----------

[PSGO-581]: https://keboola.atlassian.net/browse/PSGO-581?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/internal/pkg/service/stream/storage/level/local/volume/diskalloc/allocator_linux.go
+++ b/internal/pkg/service/stream/storage/level/local/volume/diskalloc/allocator_linux.go
@@ -31,5 +31,6 @@ func Allocated(path string) (datasize.ByteSize, error) {
 	}
 
 	// Note: Blksize is in bits not bytes
-	return datasize.ByteSize((sysStat.Blksize / 8) * sysStat.Blocks), nil
+	// Note: using type int64 for arm64 (MacOS)
+	return datasize.ByteSize((int64(sysStat.Blksize / 8)) * int64(sysStat.Blocks)), nil // nolint:unconvert
 }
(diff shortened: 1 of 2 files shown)


internal/pkg/service/stream/storage/level/local/volume/diskalloc/allocator_linux.go
```diff
@@ -31,5 +31,6 @@ func Allocated(path string) (datasize.ByteSize, error) {
 	}
 
 	// Note: Blksize is in bits not bytes
-	return datasize.ByteSize((sysStat.Blksize / 8) * sysStat.Blocks), nil
+	// Note: using type int64 for arm64 (MacOS)
+	return datasize.ByteSize((int64(sysStat.Blksize / 8)) * int64(sysStat.Blocks)), nil // nolint:unconvert
```
michaljurecko, line 35: - The original comment should be kept.
- The new comment is not accurate.

My suggestion:
```
// Notes:
// 1. "/ 8": Blksize is in bits not bytes
// 2. The type of fields can vary depending on the architecture (int32/int64), so retyping it to int64 is necessary.
```

michaljurecko: Please, also send a PR with the fixed `.air.toml`, as we discussed.

hosekpeter:
I added default values for starting stream api. I fixed this error after this command `make run-stream-service`
image

michaljurecko (approved):
LGTM

PR #1726, opened 2024-04-25
tests: Improve statsitics providers tests

michaljurecko:
Jira: https://keboola.atlassian.net/browse/PSGO-567

**Changes:**
- Added missing line/assert to the test.
-----------

### DIFF
--- a/internal/pkg/service/stream/storage/statistics/provider_test.go
+++ b/internal/pkg/service/stream/storage/statistics/provider_test.go
@@ -2,7 +2,6 @@ package statistics_test
 
 import (
 	"context"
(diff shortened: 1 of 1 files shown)


internal/pkg/service/stream/storage/statistics/provider_test.go
```diff
@@ -372,6 +373,9 @@ func TestCaches(t *testing.T) {
 			expectedRevision = header.Revision
 		}
 
+		// Test repository
+		tc.Assert(repo)
+
```
michaljurecko, line 378: Missing test for repository (it implements the provider interface).

I moved test one level upper.

PR #1732, opened 2024-04-26
Add a new alias for remote create table command

hosekpeter:
Jira: [PSGO-551](https://keboola.atlassian.net/browse/PSGO-551)

**Changes:**
- added a new alias
- updated test

-----------

[PSGO-551]: https://keboola.atlassian.net/browse/PSGO-551?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/internal/pkg/service/cli/cmd/cmd.go
+++ b/internal/pkg/service/cli/cmd/cmd.go
@@ -243,6 +243,7 @@ func NewRootCommand(stdin io.Reader, stdout io.Writer, stderr io.Writer, osEnvs 
 	root.addAlias(`r`, `template repository`)
 	root.addAlias(`repo`, `template repository`)
 	root.addAlias(`table`, `remote table`)
+	root.addAlias(`create-table`, `remote create table`)
 
 	// Add aliases to usage template
 	root.Cmd.Annotations = map[string]string{`aliases`: root.listAliases()}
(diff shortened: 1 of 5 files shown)


internal/pkg/service/cli/cmd/cmd.go
```diff
@@ -243,6 +243,7 @@ func NewRootCommand(stdin io.Reader, stdout io.Writer, stderr io.Writer, osEnvs
 	root.addAlias(`r`, `template repository`)
 	root.addAlias(`repo`, `template repository`)
 	root.addAlias(`table`, `remote table`)
+	root.addAlias(`create-table`, `remote create table`)
```
hosekpeter, line 246: I can't use `table create` , because we are using alias `table`. 

I can't think of a good name for the alias.

michaljurecko: :thinking: but it doesn't match the naming style of other commands.

@hosekpeter I suggest to move code/command, to `remote/table/create/cmd.go`, so the command `remote table create` will be implemented, and then create an alias for the old command:

```
root.addAlias(`remote create table`, `remote table create`)
```

----------

Sorry, I don't know about this issue.

michaljurecko: The goal is: if someone has `remote create table` in some script, it must remain functional.

PR #1739, opened 2024-04-30
ci: add build and test cache into github actions cache

Matovidlo:
Jira: [PSGO-561](https://keboola.atlassian.net/browse/PSGO-561)

**Changes:**
- Add caching of go build and tests into github actions cache.
- Change mod time of files due to caching issues of test cases, where it is based on mod time and each `git clone` pulls the files and changes mod time to current time.  https://github.com/golang/go/issues/58571
- Remove `-coverprofile` flag that prevents caching of tests.
- Add `-coverprofile` flag for release pipeline. Tagged commits and `main`.
- Create `go tools` cache to reduce size of cache per step.
- Cache of `e2e`  tests is already injected but not evaluated. **Will be done in next PR**.
- Add `go mod download` step to cache also packages in tools cache.
-----------

[PSGO-561]: https://keboola.atlassian.net/browse/PSGO-561?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.github/actions/go-cache/action.yml
+++ b/.github/actions/go-cache/action.yml
@@ -20,10 +20,9 @@ runs:
           ~\go\pkg
           ~/go/bin
           ~\go\bin
-          ~/.cache/golangci-lint
           ~/.cache/go-build
-          ~/Library/Caches/golangci-lint
-          ~\AppData\Local\golangci-lint
+          ~/Library/Caches/go-build
+          ~\AppData\Local\go-build
         key: ${{ runner.os }}-go-${{ env.GO_VERSION }}-v6-${{ inputs.key }}-${{ hashFiles('**/go.sum') }}
         restore-keys: |
           ${{ runner.os }}-go-${{ env.GO_VERSION }}-v6-
--- a/.github/actions/go-cache/action.yml
+++ b/.github/actions/go-cache/action.yml
@@ -21,6 +21,7 @@ runs:
           ~/go/bin
           ~\go\bin
           ~/.cache/golangci-lint
+          ~/.cache/go-build
           ~/Library/Caches/golangci-lint
           ~\AppData\Local\golangci-lint
         key: ${{ runner.os }}-go-${{ env.GO_VERSION }}-v6-${{ inputs.key }}-${{ hashFiles('**/go.sum') }}
--- a/.github/actions/go-cache/action.yml
+++ b/.github/actions/go-cache/action.yml
@@ -13,17 +13,14 @@ runs:
       run: echo 'C:\Program Files\Git\usr\bin' >> $GITHUB_PATH
 
     - name: Load Go cache
+      id: load-go-cache
       uses: actions/cache@v4
       with:
         path: |
-          ~/go/pkg
-          ~\go\pkg
-          ~/go/bin
-          ~\go\bin
           ~/.cache/go-build
           ~/Library/Caches/go-build
           ~\AppData\Local\go-build
         key: ${{ runner.os }}-go-${{ env.GO_VERSION }}-v6-${{ inputs.key }}-${{ hashFiles('**/go.sum') }}
         restore-keys: |
-          ${{ runner.os }}-go-${{ env.GO_VERSION }}-v6-
           ${{ runner.os }}-go-${{ env.GO_VERSION }}-v6-${{ inputs.key }}-
+          ${{ runner.os }}-go-${{ env.GO_VERSION }}-v6-
--- a/.github/actions/go-tools-cache/action.yml
+++ b/.github/actions/go-tools-cache/action.yml
@@ -1,0 +1,27 @@
+name: 'Load Go tools cache'
+description: 'Load Go tools cache'
+runs:
+  using: "composite"
+  steps:
+    - name: Use GNU tar instead BSD tar
+      if: runner.os == 'Windows'
+      shell: bash
+      run: echo 'C:\Program Files\Git\usr\bin' >> $GITHUB_PATH
+
+    - name: Install Go modules
+      shell: bash
+      run: go mod download
+
+    - name: Load Go tools cache
+      id: go-tools-cache
+      uses: actions/cache@v4
+      with:
+        path: |
+          ~/go/pkg
+          ~\go\pkg
+          ~/go/bin
+          ~\go\bin
+        key: ${{ runner.os }}-go-${{ env.GO_VERSION }}-v6-tools-${{ hashFiles('**/go.sum') }}
+        restore-keys: |
+          ${{ runner.os }}-go-${{ env.GO_VERSION }}-v6-tools-
+
--- a/.github/actions/go-tools-cache/action.yml
+++ b/.github/actions/go-tools-cache/action.yml
@@ -8,10 +8,6 @@ runs:
       shell: bash
       run: echo 'C:\Program Files\Git\usr\bin' >> $GITHUB_PATH
 
-    - name: Install Go modules
-      shell: bash
-      run: go mod download
-
     - name: Load Go tools cache
       id: go-tools-cache
       uses: actions/cache@v4
@@ -24,4 +20,12 @@ runs:
         key: ${{ runner.os }}-go-${{ env.GO_VERSION }}-v6-tools-${{ hashFiles('**/go.sum') }}
         restore-keys: |
           ${{ runner.os }}-go-${{ env.GO_VERSION }}-v6-tools-
+
+    - name: Install Go modules
+      shell: bash
+      run: go mod download
+
+    - name: Install tools
+      run: . ./scripts/tools.sh
+      shell: bash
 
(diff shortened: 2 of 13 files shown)


.github/actions/go-cache/action.yml
```diff
@@ -13,17 +13,14 @@ runs:
       run: echo 'C:\Program Files\Git\usr\bin' >> $GITHUB_PATH
 
     - name: Load Go cache
+      id: load-go-cache
```
Matovidlo, line 16: single multi purpose cache per platform. Executed for unit tests and E2E CLI tests.

.github/workflows/test-unit.yml
```diff
@@ -28,19 +29,34 @@ jobs:
...
+      - name: restore timestamps
+        uses: chetan/git-restore-mtime-action@v2
+
+      #- name: Change modtime of files to 1 unix timestamp
+      #  shell: bash
+      #  run: |
+      #    pwd
+      #    for i in ~/go/pkg; do
+      #      find "$i" -type f -exec touch -r {} -d 'Thu Jan 01 1970 00:00:01' {} \; || true
+      #    done
```
Matovidlo, line 54: We still do not know whether fixer of timestamp is needed. Once I tried the CI and it worked also without restoring timestamps.

scripts/tests.sh
```diff
@@ -29,7 +32,7 @@ fi
 # Run tests, sequentially because the API is shared resource
 echo "Running tests ..."
 export KBC_VERSION_CHECK=false # do not check the latest version in the tests
-cmd="gotestsum --no-color=false --format \"$TEST_LOG_FORMAT\" -- -timeout 1200s -p $TEST_PARALLELISM_PKG -parallel $TEST_PARALLELISM -v $TEST_ARGS  "$TEST_PACKAGE" $@"
+cmd="gotestsum --no-color=false --format \"$TEST_LOG_FORMAT\" -- -timeout 1200s -p $TEST_PARALLELISM_PKG -parallel $TEST_PARALLELISM $TEST_ARGS "$TEST_PACKAGE" $@"
```
Matovidlo, line 35: remove verbose by default. Add it only when `$TEST_VERBOSE` is present.

.github/workflows/test-unit.yml
```diff
@@ -7,6 +7,7 @@ env:
   GO_VERSION: 1.22.2
   TEST_PARALLELISM: 12
   TEST_PARALLELISM_PKG: 6
+  TEST_COVERAGE: false
```
Matovidlo, line 10: This has to be set to cache the test results into `go build testcache`. I will finish the CI when and how to execute the `TEST_COVERAGE:true` part.

.github/workflows/test-e2e-service-stream.yml
```diff
@@ -27,12 +27,15 @@ jobs:
           secrets: ${{ toJSON(secrets) }}
           projects: ${{ vars.TEST_KBC_PROJECTS }}
 
-      - name: Load Go cache
+      - name: Load Go build cache
         uses: ./.github/actions/go-cache
         with:
-          key: e2e
+          key: build
```
Matovidlo, line 33: currenctly same key for e2e stream and templates.

.github/actions/go-cache/action.yml
```diff
@@ -13,17 +13,14 @@ runs:
...
       uses: actions/cache@v4
       with:
         path: |
-          ~/go/pkg
-          ~\go\pkg
-          ~/go/bin
-          ~\go\bin
-          ~/.cache/golangci-lint
-          ~/Library/Caches/golangci-lint
-          ~\AppData\Local\golangci-lint
```
Matovidlo, line 25: As the linter ends up quicker even without caching I adjusted it currently by removing the files. The cache is smaller.

Matovidlo (commented):
Need to fix the CI to generate `coverprofile` on `tagged` commits and `main` branch/ release.
Caches currently takes up 4.5 GB instead of 3.9GB which is not bad result as it is approximatelly `~600MB` per `~/.cache/go-build` for each platform. So ~1.8GB was approximate increase in general and ends up only increased by `~600MB`.

.github/actions/go-cache/action.yml
Matovidlo: Can be reworked with another PR where support for Azure Blob storage will be provided.

.github/actions/go-tools-cache/action.yml
```diff
@@ -0,0 +1,23 @@
+name: 'Load Go tools cache'
```
Matovidlo, line 1: Multi purpose cache, should have plenty cache hits even test cache was not triggered yet.

.github/workflows/test-e2e-cli.yml
```diff
@@ -28,7 +30,11 @@ jobs:
     steps:
       - name: Setup line endings
         run: git config --global core.autocrlf false
-      
+
+      - name: Change default TEST_COVERAGE when on main branch
+        if: github.ref == 'refs/heads/main'
+        run: echo "TEST_COVERAGE=true" >> $GITHUB_ENV
```
Matovidlo, line 36: Coverage triggered only on `main` branch as was discussed.

.github/workflows/test-unit.yml
```diff
@@ -26,19 +27,35 @@ jobs:
...
           projects: ${{ vars.TEST_KBC_PROJECTS }}
 
-      - name: Load Go cache
+      - name: Load Go build cache
         uses: ./.github/actions/go-cache
         with:
-          key: unit-${{ matrix.name }}
+          key: unit
+
+      - name: Change modtime of files to 1 unix timestamp
```
Matovidlo, line 48: Currently only `unit` tests are changing timestamps. E2e tests will be solved in another PR

scripts/tools.sh
```diff
@@ -10,7 +10,10 @@ SCRIPT_DIR="$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
 cd "$SCRIPT_DIR"
 
 # go-licenses
-go install github.com/google/go-licenses@latest
+if ! command -v go-licenses &> /dev/null
+then
+  go install github.com/google/go-licenses@latest
+fi
```
Matovidlo, line 16: Speeds up windows build tools

.github/actions/go-tools-cache/action.yml
```diff
@@ -0,0 +1,31 @@
...
+    - name: Load Go tools cache
+      id: go-tools-cache
+      uses: actions/cache@v4
+      with:
+        path: |
+          ~/go/pkg
+          ~\go\pkg
+          ~/go/bin
+          ~\go\bin
+        key: ${{ runner.os }}-go-${{ env.GO_VERSION }}-v6-tools-${{ hashFiles('**/go.sum') }}
```
michaljurecko, line 20: `bin` is enought here, right?

.github/workflows/test-k8s-service-buffer.yml
```diff
@@ -242,16 +239,13 @@ jobs:
         if: always()
         run: |
           set -Eeuo pipefail
-
```
michaljurecko, line 245: Why you are removing these empty lines?

michaljurecko (approved):
:clap: Let's try it.

.github/actions/go-tools-cache/action.yml
Matovidlo: According to the documentation I have read it should be sufficient. https://github.com/actions/cache/blob/main/examples.md#go---modules . I suspect that `bin` is generated next to the `pkg` folder.

.github/workflows/test-k8s-service-buffer.yml
Matovidlo: Jachym was last time complaining about it in PR. I have set up my editor according to `editorconfig` and done this. I can drop the commit if we want, just want to point out that according to editorconfig the trailing spaces are removed.
As I found on this line newline and trailing spaces I just removed the newline.
As I pointed out above, i can drop the commit :slightly_smiling_face:

michaljurecko: :shrug: I understand if it removed trailing spaces but why should it delete empty lines in the middle of the code?

hosekpeter (approved):

.github/workflows/test-k8s-service-buffer.yml
Matovidlo: I understand, maybe my bad, I'll make sure next time leave new lines :slightly_smiling_face: .

.github/actions/go-tools-cache/action.yml
michaljurecko: :thinking: tools are some compiled binaries we use, generic cache documentation is not relevant here, right?

PR #1743, opened 2024-05-02
migration: add buffer migration to stream

hosekpeter:
Jira: [PSGO-593](https://keboola.atlassian.net/browse/PSGO-593)

**Changes:**
- added a receiver(source) migration to a new stream api

-----------
This task deals with getting `receivers` from the `buffer api` and creating them in the `stream api`.

[PSGO-593]: https://keboola.atlassian.net/browse/PSGO-593?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/internal/pkg/service/stream/migrate/migrate.go
+++ b/internal/pkg/service/stream/migrate/migrate.go
@@ -1,0 +1,58 @@
+package main
+
+import (
+	"context"
+	"flag"
+	"os"
+	"sync"
+
+	"github.com/keboola/keboola-as-code/internal/pkg/log"
+	"github.com/keboola/keboola-as-code/internal/pkg/service/stream/migrate/source"
+)
+
+type Flags struct {
+	storageAPIToken string
+	host            string
+}
+
+func NewFlags() Flags {
+	storageAPIToken := flag.String("storage-api-token", "", "storage api token")
+	host := flag.String("host", "", "host")
+
+	flag.Parse()
+
+	return Flags{*storageAPIToken, *host}
+}
+
+func main() {
+	ctx := context.Background()
+	logger := log.NewServiceLogger(os.Stdout, true)
+	wg := sync.WaitGroup{}
+
+	flags := NewFlags()
+	if flags.host == "" && flags.storageAPIToken == "" {
+		logger.Error(ctx, "usage: migrate storage-api-token host")
+		os.Exit(1)
+	}
+
+	// fetch receivers and exports from old API
+	bufferReceivers, err := source.FetchBufferReceivers(ctx, flags.host, flags.storageAPIToken, logger)
+	if err != nil {
+		logger.Error(ctx, err.Error())
+		os.Exit(1)
+	}
+
+	logger.Info(ctx, `Starting migration...`)
+	for _, receiver := range bufferReceivers.Receivers {
+		wg.Add(1)
+		go func() {
+			err = receiver.CreateSource(ctx, flags.storageAPIToken, flags.host, &wg, logger)
+			if err != nil {
+				logger.Error(ctx, err.Error())
+			}
+		}()
+	}
+	wg.Wait()
+
+	logger.Info(ctx, `Migration was finished.`)
+}
--- a/internal/pkg/service/stream/migrate/migrate.go
+++ b/internal/pkg/service/stream/migrate/migrate.go
@@ -4,10 +4,10 @@ import (
 	"context"
 	"flag"
 	"os"
-	"sync"
 
 	"github.com/keboola/keboola-as-code/internal/pkg/log"
 	"github.com/keboola/keboola-as-code/internal/pkg/service/stream/migrate/source"
+	"github.com/keboola/keboola-as-code/internal/pkg/utils/errors"
 )
 
 type Flags struct {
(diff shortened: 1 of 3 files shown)


internal/pkg/service/stream/migrate/migrate.go
```diff
@@ -0,0 +1,58 @@
...
+
+	logger.Info(ctx, `Starting migration...`)
+	for _, receiver := range bufferReceivers.Receivers {
+		wg.Add(1)
+		go func() {
+			err = receiver.CreateSource(ctx, flags.storageAPIToken, flags.host, &wg, logger)
+			if err != nil {
+				logger.Error(ctx, err.Error())
+			}
+		}()
```
Matovidlo, line 53: Why the `defer wg.Done()` is not here but in function.  Since you use it as prerequisite step as first statement in `CreateSource` I would move it here as first step.

internal/pkg/service/stream/migrate/source/source.go
```diff
@@ -0,0 +1,190 @@
...
+	err = json.NewDecoder(resp.Body).Decode(&result)
+	if err != nil {
+		return nil, err
+	}
+
+	return result, nil
+}
+
+func (r *Receiver) CreateSource(ctx context.Context, token string, host string, wg *sync.WaitGroup, l log.Logger) error {
+	defer wg.Done()
```
Matovidlo, line 119:
```suggestion
func (r *Receiver) CreateSource(ctx context.Context, token string, host string, l log.Logger) error {
```

internal/pkg/service/stream/migrate/source/source.go
```diff
@@ -0,0 +1,190 @@
...
+
+	// Set a payload to create source
+	body, err := r.createSourcePayload()
+	if err != nil {
+		return err
+	}
+
+	streamHost := strings.Replace(host, bufferPrefix, streamPrefix, 1)
```
Matovidlo, line 127: This looks weird, could we get some example how `bufferPrefix` or `streamPrefix` looks like? Or maybe create `substituteHost` function so it is more convinient that the `Replace` has to be used when communicating.
Something like
```
func substituteHost(host, bufferPrefix, streamPrefix string) string {
  return "https://" + strings.Replace(host, bufferPrefix, streamPrefix, 1)
}
```

internal/pkg/service/stream/migrate/source/source.go
```diff
@@ -0,0 +1,190 @@
...
+	case http.StatusUnprocessableEntity:
+		l.Infof(ctx, `Maximum number of sources per project is 100.`)
+	default:
+		l.Infof(ctx, "Unexpected status code: %v for source %s.", resp.StatusCode, r.ID)
+	}
+	return nil
+}
+
+func newHTTPRequest(ctx context.Context, c RequestConfig) (*http.Response, error) {
+	url := fmt.Sprintf("%s%s", c.host, c.path)
```
Matovidlo, line 158: We should enable linter with reworking these `fmt.Sprintf` statements to construct URL. 
More convinient way to do it is to actually use `url` package, Then you use the constructor `url.URL: {Host: c.host, Path: c.path}` Instead of `fmt.Sptrintf`. :sweat_smile:

internal/pkg/service/stream/migrate/source/source.go
```diff
@@ -0,0 +1,190 @@
...
+	Mapping    Mapping    `json:"mapping"`
+	Conditions Conditions `json:"conditions"`
+}
+type Receiver struct {
+	ID          string    `json:"id"`
+	URL         string    `json:"url"`
+	Name        string    `json:"name"`
+	Description string    `json:"description"`
+	Exports     []Exports `json:"exports"`
+}
```
Matovidlo, line 63: Try to move definitions together. Maybe split `source` into more files?

Matovidlo (changes requested):

internal/pkg/service/stream/migrate/migrate.go
hosekpeter: After discussion with Michal the goroutine was removed,because there is not so much data.

internal/pkg/service/stream/migrate/migrate.go
```diff
@@ -0,0 +1,58 @@
...
+	storageAPIToken := flag.String("storage-api-token", "", "storage api token")
+	host := flag.String("host", "", "host")
+
+	flag.Parse()
+
+	return Flags{*storageAPIToken, *host}
+}
+
+func main() {
+	ctx := context.Background()
```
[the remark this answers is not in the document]
hosekpeter: [909458d](https://github.com/keboola/keboola-as-code/pull/1743/commits/909458d63ae43159c11dd8a067cab9016a6f24bb)

internal/pkg/service/stream/migrate/migrate.go
hosekpeter: [909458d](https://github.com/keboola/keboola-as-code/pull/1743/commits/909458d63ae43159c11dd8a067cab9016a6f24bb)

internal/pkg/service/stream/migrate/source/source.go
hosekpeter: [475667f](https://github.com/keboola/keboola-as-code/pull/1743/commits/475667ff4e3ce6df6dc3190038050551733e2e19)

Matovidlo (approved):
LGTM

PR #1753, opened 2024-05-07
fix: Remove sinks list from the Source detail endpoint

michaljurecko:

**Changes:**
- Removed `sinks` field from the Source detail endpoint.
  - It now match REST API conventions.
  - There will be an `aggregated` endpoint in the UI, with all nested objects/statistics.

-----------


api/stream/design.go
```diff
@@ -643,8 +643,7 @@ var Source = Type("Source", func() {
 	Attribute("version", EntityVersion)
 	Attribute("deleted", DeletedEntity)
 	Attribute("disabled", DisabledEntity)
-	Attribute("sinks", Sinks)
-	Required("version", "type", "name", "description", "type", "sinks")
+	Required("version", "type", "name", "description", "type")
 })
```
michaljurecko, line 647: Removed `sinks` field, list of nested sinks, it it went against REST principles.

PR #1755, opened 2024-05-09
ci: add support for s3 storage as cache.

Matovidlo:
Jira: [PSGO-579](https://keboola.atlassian.net/browse/PSGO-579)

**Changes:**
- Add support of s3 storage as cache instead of github cache.
- Remove usage of previous github cache.
- Rework both tools and go-build cache.

-----------

[PSGO-579]: https://keboola.atlassian.net/browse/PSGO-579?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.github/actions/go-cache-push-s3/action.yml
+++ b/.github/actions/go-cache-push-s3/action.yml
@@ -1,0 +1,93 @@
+name: 'Save Go essentials into S3'
+description: 'Push the Go essentials into S3'
+inputs:
+  key:
+    description: Cache key
+    required: true
+  paths:
+    description: Where files/folders is located
+    required: true
+  aws-s3-bucket:
+    description: Bucket to save cache
+    required: true
+  aws-access-key-id:
+    description: Access key to AWS
+    required: true
+  aws-secret-access-key:
+    description: Secret access key to AWS
+    required: true
+  override:
+    description: Always override the cache
+    required: false
+    default: "true"
+runs:
+  using: "composite"
+  steps:
+    - name: Do not push cache when exists
+      shell: bash
+      run: |
+        if '${{ inputs.override }}' == 'true'; then
+          echo "SKIP=false" >>  $GITHUB_ENV
+          exit 0
+        fi
+
+        key=$(aws s3 ls s3://${{ inputs.aws-s3-bucket }} | tr -s ' ' | cut -d ' ' -f 4 | grep '${{ runner.os }}-go-${{ env.GO_VERSION }}-v6-${{ inputs.key }}-${{ hashFiles('**/go.sum') }}') || true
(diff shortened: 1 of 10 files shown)


.github/actions/go-cache-push-s3/action.yml
```diff
@@ -0,0 +1,93 @@
...
+
+        key=$(aws s3 ls s3://${{ inputs.aws-s3-bucket }} | tr -s ' ' | cut -d ' ' -f 4 | grep '${{ runner.os }}-go-${{ env.GO_VERSION }}-v6-${{ inputs.key }}-${{ hashFiles('**/go.sum') }}') || true
+        echo "SKIP=false" >> $GITHUB_ENV
+
+      env:
+        AWS_ACCESS_KEY_ID: ${{ inputs.aws-access-key-id }}
+        AWS_SECRET_ACCESS_KEY: ${{ inputs.aws-secret-access-key }}
+        AWS_DEFAULT_REGION: 'us-east-1'
+
+    - name: Convert multiline string to singleline
```
Matovidlo, line 42: converting multi variable escaped by `\n` into signle string escaped with ` `.

.github/actions/go-cache/action.yml
```diff
@@ -4,6 +4,22 @@ inputs:
...
+  aws-secret-access-key:
+    description: Secret access key to AWS
+    required: true
+  windows_go_cache_folder:
+    description: Where windows build cache is located
+    required: true
+  with-variant-keys:
+    description: Check more generic cache keys
+    required: false
+    default: "true"
```
Matovidlo, line 22: Instead of `restore-keys` as the workflow before, I decided to do `boolean` indication of variant keys. when non present by default it tries to restore also less scoped cache keys.

.github/actions/go-cache-push-s3/action.yml
```diff
@@ -0,0 +1,93 @@
...
+      shell: python
```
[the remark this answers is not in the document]
Matovidlo: I understand, The task is to check whether the given `paths` exists in direcotry structure. At first I wanted to use `stat` however it ends up with an error when the `directory/file` does not exists. And in CI it is very not good solution. When shell ends up with status code different than `0`, it stops. Therefore I have to use already for `aws cli` everywhere `key=$(cmd 2&>1) || true` to suppress the status code and actually check the stdout/stderr.
I came accross this solution which I found more convinient due to no need of suppressing error and just plain iteration over paths that exists, rest of paths are concatenated into final path list that is compressed.
If I would want to convert everyting into python It would be more boilerplate than this nested statements in `shell`.

PR #1765, opened 2024-05-14
feat: Add a new field `appSlug`

hosekpeter:
Jira: [PSGO-611](https://keboola.atlassian.net/browse/PSGO-611)

**Changes:**
- added a new field `appSlug`
- updated tests

-----------
SLACK: https://keboolaglobal.slack.com/archives/C04L6J6CWJJ/p1715092291103389?thread_ts=1715085124.992289&cid=C04L6J6CWJJ

[PSGO-611]: https://keboola.atlassian.net/browse/PSGO-611?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/internal/pkg/service/appsproxy/dataapps/api/config.go
+++ b/internal/pkg/service/appsproxy/dataapps/api/config.go
@@ -29,6 +29,7 @@ type AppID string
 type AppConfig struct {
 	ID             AppID              `json:"appId"`
 	Name           string             `json:"appName"`
+	AppSlug        *string            `json:"appSlug,omitempty"`
 	ProjectID      string             `json:"projectId"`
 	UpstreamAppURL string             `json:"upstreamAppUrl"`
 	AuthProviders  provider.Providers `json:"authProviders"`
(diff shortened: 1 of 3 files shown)


internal/pkg/service/appsproxy/proxy/proxy_test.go
```diff
@@ -2105,6 +2105,7 @@ func testDataApps(upstream *url.URL, m []*mockoidc.MockOIDC) []api.AppConfig {
 			ID:             "12345",
 			ProjectID:      "123",
 			Name:           "LOWERCASE",
+			AppSlug:        pointer("LOWERCASE"),
```
michaljurecko, line 2108: Please, in tests, use some different app name as the `slug`, so it will be clear that the `Name` is no more used in URL.

All apps can have for example name `My App`.

internal/pkg/service/appsproxy/dataapps/api/config.go
```diff
@@ -56,11 +57,15 @@ func (c AppConfig) IdAndName() string {
 	return c.Name + "-" + c.ID.String()
 }
 
+func (c AppConfig) Slug() string {
+	return *c.AppSlug
+}
+
 func (c AppConfig) Domain() string {
-	if c.Name == "" {
+	if c.AppSlug == nil {
```
michaljurecko, line 65: `if c.AppSlug == nil || *c.AppSlug == ""`, to prevent `-12345` prefix if the value is empty string

PR #1759, opened 2024-05-13
PSGO-596 - TO SPLIT

michaljurecko:
Temporary: This PR is divided into smaller parts.

-----------

PR #1773, opened 2024-05-17
feat: Improve etcd helpers

michaljurecko:

--------------

**Changes:**
- Added function `etcdhelper.AssertKVsFromFile`.
- Added function `etcdlogger.AssertFromFile`.
- Updated linter.
- Added deepcopy support for `utctime.UTCTime`.
- Fixed `utctime.UTCTime.Time.Equal(now)`.

-----------

Commend per commits.

-----------

image

image


internal/pkg/utils/etcdhelper/assert.go
```diff
@@ -68,14 +81,26 @@ func AssertKeys(t assert.TestingT, client etcd.KV, expectedKeys []string, ops ..
 	return assert.Equal(t, expectedKeys, actualKeys)
 }
 
+// AssertKVsFromFile dumps all KVs from an etcd database and compares them with content of the file.
+// In the file, a wildcards can be used, see the wildcards package.
+func AssertKVsFromFile(t assert.TestingT, client etcd.KV, path string, ops ...AssertOption) bool {
```
michaljurecko, line 86: Added function to compare etcd state with a file, because `expected` state can have a lot of lines.

internal/pkg/utils/etcdhelper/assert_test.go
```diff
@@ -199,6 +199,66 @@ value2
...
+	            	-----
+	            	valueA
+	            	-----
+
+%AValue of the actual key
+	            	"key2"
+	            	doesn't match the expected key
+	            	"key2"
+	            	defined in the file
+	            	"fixtures/expected-001.txt"
```
michaljurecko, line 244: Information, where the expected file is located.

internal/pkg/utils/etcdhelper/assert.go
```diff
@@ -113,6 +138,14 @@ func AssertKVs(t assert.TestingT, client etcd.KV, expectedKVs []KV, ops ...Asser
...
 	}
 
+	// Dump actual state
+	if c.expectedStateFromFile != "" {
+		outDir := filepath.Join(filepath.Dir(c.expectedStateFromFile), ".out")              //nolint:forbidigo // no virtual FS
+		filePath := filepath.Join(outDir, filepath.Base(c.expectedStateFromFile)+".actual") //nolint:forbidigo // no virtual FS
+		assert.NoError(t, os.MkdirAll(outDir, 0o750))                                       //nolint:forbidigo // no virtual FS
+		assert.NoError(t, os.WriteFile(filePath, []byte(KVsToString(actualKVs)), 0o600))    //nolint:forbidigo // no virtual FS
+	}
+
```
michaljurecko, line 148: Actual state is saved to the `.out` directory, to be checked or copied to the expected file.

internal/pkg/utils/etcdlogger/assert.go
```diff
@@ -27,3 +34,22 @@ func Assert(t assert.TestingT, expected, actual string) bool {
 
 	return wildcards.Assert(t, expected, actual)
 }
+
+// AssertFromFile compares logs captured by the KVLogWrapper.
+// Comments "// ..." and empty lines are ignored.
+func AssertFromFile(t assert.TestingT, path, actual string) bool {
+	result := false
```
michaljurecko, line 41: Another new function, to check log of etcd operations.

build/ci/golangci.yml
```diff
@@ -23,6 +23,13 @@ linters-settings:
     # minimal code complexity to report, 30 by default (but we recommend 10-20)
     min-complexity: 10
 
+  gosec:
+    excludes:
+      - G601 # Implicit memory aliasing of items from a range statement (Go 1.22)
+
```
michaljurecko, line 29: It is no more issues in Go 1.22, but the linter is not updated yet.

build/ci/golangci.yml
```diff
@@ -23,6 +23,13 @@ linters-settings:
...
+  dupl:
+    threshold: 500
```
michaljurecko, line 31: Increased threshold for duplication detections, to prevent multiple false positives.

build/ci/golangci.yml
```diff
@@ -105,7 +112,7 @@ linters:
     - forbidigo
     - gci
     - gochecknoglobals
-    - gochecknoinits
+    #- gochecknoinits - temporary disabled https://github.com/golangci/golangci-lint/issues/4697
```
michaljurecko, line 115: Broken in new GoLangLint version.

build/ci/golangci.yml
```diff
@@ -144,6 +150,7 @@ linters:
     #- gofmt # replaced with gofumpt
     #- nolintlint # strange behavior
     #- gomoddirectives # allow replace directive in go.mod
+    #- misspell - broken, rewrites code
```
michaljurecko, line 153: Broken, renames variables, rewriting cannot be disabled.

scripts/tools.sh
```diff
@@ -30,7 +30,7 @@ fi
 # golangci-lint
 if ! command -v golangci-lint &> /dev/null
 then
-  curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b $(go env GOPATH)/bin v1.57.2
+  curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b $(go env GOPATH)/bin v1.58.0
 fi
```
michaljurecko, line 34: Updated `golangci-lint`.

internal/pkg/service/common/utctime/utctime.go
```diff
@@ -13,6 +15,10 @@ type (
 	UTCTime time.Time
 )
 
+func (v UTCTime) HandleDeepCopy(_ deepcopy.TranslateFn, _ deepcopy.Path, _ deepcopy.VisitedPtrMap) (UTCTime, deepcopy.CloneFn) {
+	return v, nil
+}
+
```
michaljurecko, line 21: Unrelated small change, `deepcopy` now can copy `UTCTime` time.

This method is required for all types with a private field.

michaljurecko: In this case, the value is simple, and we only copy it.

internal/pkg/service/common/utctime/utctime.go
```diff
@@ -61,7 +61,7 @@ func FormatTime(t time.Time) string {
 }
 
 func From(t time.Time) UTCTime {
-	return UTCTime(t.UTC().Truncate(time.Millisecond))
+	return UTCTime(t.UTC())
 }
```
michaljurecko, line 65: Small unrelated fix:

This broke `UTCTime.Time().Equal(now)` comparison.

On marshalling, microseconds (all parts smaller than millisecond) are truncated by the specified format.

The format ends with 3 `.000` millisecond digits.

build/ci/golangci.yml
Matovidlo: This is quite usefull plugin which prevents misspelling. I understand that rewriting is annoying however I would rather adopt than drop. 
@jachym-tousek-keboola @hosekpeter what do you think?

internal/pkg/service/common/utctime/utctime.go
Matovidlo: Isnt the problem of comparison side instead of function side? You can truncate `time.Now`. I would expect that the test uses truncate as well.

internal/pkg/utils/etcdlogger/assert.go
```diff
@@ -27,3 +34,22 @@ func Assert(t assert.TestingT, expected, actual string) bool {
...
+	data, err := os.ReadFile(path) //nolint:forbidigo // no virtual FS
+	if assert.NoError(t, err) || errors.Is(err, os.ErrNotExist) {
+		expected := string(data)
+		result = Assert(t, expected, actual)
+	}
```
Matovidlo, line 46: I could not find usage within this PR.  The `if` condition  asserts actual vs. expected, when file within given `path` was possible to read or when it does not exists? Otherwise it creates state dump into `.out` directory. 

Is the dumping section usefull? It covers edge case like `permission denied`.

Matovidlo (approved):
LGTM

build/ci/golangci.yml
michaljurecko: It is not annoying, it is not working, illogically it overwrites valid names (after the update).

image

internal/pkg/utils/etcdlogger/assert.go
michaljurecko: > I could not find usage within this PR. 

Yes, you should search for usage in [the final code](https://github.com/keboola/keboola-as-code/pull/1759), (or do you want to review all in one PR? :slightly_smiling_face:  )

--------------------

>  The if condition ...

The if condition contains `errors.Is(err, os.ErrNotExist)` to print actual keys, even if the file doesn't exists.
It also dumps the actual state to the `.out` dir, so you can easily create the expected file, by copying the actual file.
The "file not exist" error is printed at first.

-------------------

> Is the dumping section usefull? It covers edge case like permission denied.

I don't understand.

build/ci/golangci.yml
Matovidlo: I understand, what about using `corrections` file/ `-i` to prevent update `Statuser`.
I think we have some common dictionary that is not part of the linter and it expects that it is a mistake. I can easily create a Jira task to go trough new `misspell` corrections, ignore them and add them to our `Keboola` dictionary.
It seems common issue also for others not just us :)

internal/pkg/utils/etcdlogger/assert.go
Matovidlo: I'll try to adjust it in another thread

hosekpeter (approved):
LGTM

internal/pkg/utils/etcdlogger/assert.go
```diff
@@ -27,3 +34,22 @@ func Assert(t assert.TestingT, expected, actual string) bool {
...
+
+	// Dump actual state to the .out dir
+	outDir := filepath.Join(filepath.Dir(path), ".out")              //nolint:forbidigo // no virtual FS
+	filePath := filepath.Join(outDir, filepath.Base(path)+".actual") //nolint:forbidigo // no virtual FS
+	assert.NoError(t, os.MkdirAll(outDir, 0o750))                    //nolint:forbidigo // no virtual FS
+	assert.NoError(t, os.WriteFile(filePath, []byte(actual), 0o600)) //nolint:forbidigo // no virtual FS
+
+	return result
```
Matovidlo, line 54:
```suggestion
	if err != nil && !errors.Is(err, os.ErrNotExists) {
		return
	}
	
	expected := string(data)
	result = Assert(t, expected, actual)
	// Dump actual state to the .out dir
	outDir := filepath.Join(filepath.Dir(path), ".out")              //nolint:forbidigo // no virtual FS
	filePath := filepath.Join(outDir, filepath.Base(path)+".actual") //nolint:forbidigo // no virtual FS
	assert.NoError(t, os.MkdirAll(outDir, 0o750))                    //nolint:forbidigo // no virtual FS
	assert.NoError(t, os.WriteFile(filePath, []byte(actual), 0o600)) //nolint:forbidigo // no virtual FS

	return result
```
I think this more fits into logic. So all errors except `os.ErrNotExists` are resolved as `false`. Otherwise the error occurrs or not, does not matter. Do the logic.
I expected that the `return` is inside the if statement, but it was not so I get confused.
Also you would be able to get rid of `result := false` at start of this function so the code is thinner.
Convence in Go is to catch error and return so I would assume that here the logic is the same even though it is a test :slightly_smiling_face:
If we would like always to dump the file, could be done as `defer` function.

build/ci/golangci.yml
hosekpeter: If it overwrites to nonsensical names, I certainly wouldn't want it there because I don't want to have to go through it again to see if something was renamed wrong or something.

michaljurecko: @Matovidlo please create a task to fix it

PR #1786, opened 2024-05-23
refactor: Simplify SinkStatisticsFiles

jachym-tousek-keboola:
I realized I was doing the same logic in two different ways in two different places so I tried to simplify it here.


hosekpeter (approved):

michaljurecko:
@jachym-tousek-keboola it is compatible with the latest branch? If not, I prefer to include it in an PR above `TO SPLIT` branch.

PR #1767, opened 2024-05-15
feat: Improve etcdop framework

michaljurecko:
Jira: https://keboola.atlassian.net/browse/PSGO-596

All tests will pass in the final [PR](https://github.com/keboola/keboola-as-code/pull/1770).

[Large set of fixes and E2e tests](https://github.com/keboola/keboola-as-code/pull/1759) is divided into smaller PRs to do a review.

--------------

**Changes:**
- Various improvements and fixes in `etcdop` framework, see comments.

-----------

Commend per commits.

-----------

image

---------

At first, please read `AtomicOp` docs:

https://github.com/keboola/keboola-as-code/blob/a8001e970c4b56222dfdc7e907b1d2fce18fde9a/internal/pkg/service/common/etcdop/op/atomic.go#L14-L34

PR #1789, opened 2024-05-27
feat: Fix enable/disable by rollback

michaljurecko:
Jira: XXXXX

**Changes:**
- ...

-----------


internal/pkg/service/common/etcdop/op/atomic.go
```diff
@@ -56,8 +56,13 @@ type Mutex interface {
...
 	v := &AtomicOp[R]{AtomicOpCore: newAtomicCore(client), result: result}
 	v.setProcessorFactory(func() func(ctx context.Context, r *TxnResult[NoResult]) {
+		// Invoke processors only if the transaction succeeded and there is no collision.
+		// Otherwise, succeeded == false, the processors are skipped,
+		// because the atomic operation is retried due to a collision or an etcd error, see AtomicOp.Do method.
 		return func(ctx context.Context, r *TxnResult[NoResult]) {
-			v.processors.invoke(ctx, newResult(r.Response(), v.result))
+			if r.Err() != nil || r.Succeeded() {
+				v.processors.invoke(ctx, newResult(r.Response(), v.result))
+			}
```
michaljurecko, line 65: Reverted/fixed :man_facepalming: 

I didn't remember why it was there and I deleted the condition and didn't run all the tests.

Fixed and I added the comment.

See comment: 
https://github.com/keboola/keboola-as-code/pull/1767#discussion_r1604499784

AtomicOp is retried here:
https://github.com/keboola/keboola-as-code/blob/c7f5e796aa1178d25c7c3afb25e72ceefc612329/internal/pkg/service/common/etcdop/op/atomic_do.go#L25

------------

It is questionable whether there should/shouldn't be `r.Err != nil`. it means, does it make sense for the processor to solve some network and similar errors.

internal/pkg/service/stream/definition/repository/sink/sink_enable_test.go
```diff
@@ -92,6 +92,27 @@ func TestSinkRepository_Enable(t *testing.T) {
...
 		assert.True(t, sink.IsEnabled())
 		assert.True(t, sink.IsEnabledAt(now))
+		assert.Equal(t, definition.VersionNumber(3), sink.VersionNumber())
+	}
+
+	// Rollback to the disabled state
+	// -----------------------------------------------------------------------------------------------------------------
+	{
+		now = now.Add(time.Hour)
+		sink, err := repo.RollbackVersion(sinkKey, now, by, 2).Do(ctx).ResultOrErr()
```
Matovidlo, line 102: i see that we rollback within the test to state 2 which I supposed is Create - OK, Get -OK and Disable - OK <- here.
Why the definition.VersionNumber is 4? It is just autoincrement of versions? I see 3 version or 5 but not 4. Could you explain why the VersionNumber is 4? Is it 3 versions + 1 rollback?

internal/pkg/service/stream/definition/repository/sink/sink_enable_test.go
```diff
@@ -259,4 +280,30 @@ func TestSinkRepository_EnableSinksOnSourceEnable(t *testing.T) {
 		require.NoError(t, err)
 		assert.True(t, sink3.IsEnabled())
 	}
+
+	// Rollback Source to the disabled state
+	// -----------------------------------------------------------------------------------------------------------------
+	{
+		now = now.Add(time.Hour)
+		source, err := d.DefinitionRepository().Source().RollbackVersion(sourceKey, now, by, 2).Do(ctx).ResultOrErr()
```
Matovidlo, line 288: The version 1 is when we create the `Source`? Then version 2 is 1 change performed on previously created `Source` for example.

internal/pkg/service/common/etcdop/op/atomic.go
michaljurecko: cc @jachym-tousek-keboola @hosekpeter 

-----

Etcd client itself has retry mechanism, but only for operations whose retried execution will certainly not cause a problem, e.g. `GET RANGE`. Retries are not made for `PUT`, because it can lead to the fact that the key will have 2 revisions, two historical versions instead of one. If the first attempt was successful, but the report did not reach the client.

-----

For all operations, we have additional retry mechanism for etcd/network errors, see:
https://github.com/keboola/keboola-as-code/blob/02d3bde2cd901e0c375008edbc41f8bdfa3395ee/internal/pkg/service/common/etcdop/op/op.go#L84-L88

https://github.com/keboola/keboola-as-code/blob/c1adb0608219fe3fb7610e145b9bd733b1a9ffb8/internal/pkg/service/common/etcdop/op/retry.go#L35-L72

We assume following errors as safe for retry, for our use-cases.
https://github.com/keboola/keboola-as-code/blob/c1adb0608219fe3fb7610e145b9bd733b1a9ffb8/internal/pkg/service/common/etcdop/op/retry.go#L16-L28

-------------------

AND! we have additional retry mechansim for `AtomicOp`, primarily for a logical :exclamation:  error when a collision occurred and the top-level IF conditions of `AtomicOp` were not met.

See `!ok`:
https://github.com/keboola/keboola-as-code/blob/8fd94c901089f8c17d58385111a7f0924c78bde0/internal/pkg/service/common/etcdop/op/atomic_do.go#L21-L35

hosekpeter (approved):

internal/pkg/service/stream/definition/repository/sink/sink_enable_test.go
michaljurecko: [tests: Add more version asserts](https://github.com/keboola/keboola-as-code/pull/1789/commits/8af8087e1bd94548fe22e0c30ca38661ce6b97d7)

Matovidlo: Thank you, so read operations are not counted as versionNumber obviously. Rollback is to exact version.

Matovidlo (approved):
LGTM

PR #1810, opened 2024-06-04
lint: Add linter for named parameters in interfaces.

Matovidlo:

**Changes:**
- Add linter for named parameters in interfaces.

-----------

### DIFF
--- a/internal/pkg/model/manifest.go
+++ b/internal/pkg/model/manifest.go
@@ -21,7 +21,7 @@ type RecordPaths interface {
 	// GetRelativePath - for example path of the object inside parent object/path.
 	GetRelativePath() string
 	// SetRelativePath - for example path of the object inside parent object/path.
-	SetRelativePath(p string)
+	SetRelativePath(path string)
 	// GetParentPath - for example path of the parent object.
 	GetParentPath() string
 	// SetParentPath - for example path of the parent object.
--- a/internal/pkg/model/manifest.go
+++ b/internal/pkg/model/manifest.go
@@ -21,11 +21,11 @@ type RecordPaths interface {
 	// GetRelativePath - for example path of the object inside parent object/path.
 	GetRelativePath() string
 	// SetRelativePath - for example path of the object inside parent object/path.
-	SetRelativePath(string)
+	SetRelativePath(p string)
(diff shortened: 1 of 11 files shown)


internal/pkg/model/manifest.go
```diff
@@ -21,11 +21,11 @@ type RecordPaths interface {
 	// GetRelativePath - for example path of the object inside parent object/path.
 	GetRelativePath() string
 	// SetRelativePath - for example path of the object inside parent object/path.
-	SetRelativePath(string)
+	SetRelativePath(p string)
 	// GetParentPath - for example path of the parent object.
 	GetParentPath() string
```
michaljurecko, line 26: I don't see any benefit in it at all.

Matovidlo: If you call the names properly you know what it accepts (`string` or `int`) is quite generic. In this case it is `p` which means path. It could also prevent some naming issue of method in interface against it's parameters. But that's my opinion

internal/pkg/model/manifest.go
```diff
@@ -21,11 +21,11 @@ type RecordPaths interface {
 	// GetRelativePath - for example path of the object inside parent object/path.
 	GetRelativePath() string
 	// SetRelativePath - for example path of the object inside parent object/path.
-	SetRelativePath(string)
+	SetRelativePath(p string)
```
jachym-tousek-keboola, line 24: I'd rather use actual words than abbreviations to be honest... like `path` here.

Matovidlo: 06c930872aa366ca9896ba7c9a5b26ef4293aefa

hosekpeter (approved):
LGTM

internal/pkg/model/manifest.go
jachym-tousek-keboola: You still left `p` here... :sweat_smile:

jachym-tousek-keboola: You still left `p` here... :sweat_smile:

```suggestion
	SetRelativePath(path string)
```

jachym-tousek-keboola (approved):

PR #1813, opened 2024-06-04
lint: Add testifylint to prevent unnecessary test code issues

Matovidlo:

**Changes:**
- Add testifylint to prevent unnecessary test code issues (require/assert).

-----------


internal/pkg/diff/diff_test.go
```diff
@@ -22,11 +23,11 @@ func TestDiffOnlyInLocal(t *testing.T) {
...
-	assert.NoError(t, projectState.Set(branchState))
+	require.NoError(t, projectState.Set(branchState))
 
 	d := NewDiffer(projectState)
 	results, err := d.Diff()
-	assert.NoError(t, err)
+	require.NoError(t, err)
 	assert.Len(t, results.Results, 1)
 	result := results.Results[0]
 	assert.Equal(t, ResultOnlyInLocal, result.State)
```
michaljurecko, line 33: Sometimes it makes sense for the test to continue, so you get the overall picture and you don't have to run the test repeatedly.

I prefer to use assert and require only when there is no point in continuing.

michaljurecko (changes requested):

internal/pkg/diff/diff_test.go
Matovidlo: Should you continue when `err` is not nil? If yes then use `if assert.NoError(t, err); err != nil { }` Otherwise it makes sense. I don't see anything else just a plain help to prevent unnecessary panics in test code. When you do not care about error, just omit it using `_`

jachym-tousek-keboola (approved):

PR #1815, opened 2024-06-04
lint: Add musttag linter.

Matovidlo:
Jira: XXXXX

**Changes:**
- ...

-----------

### DIFF
--- a/internal/pkg/service/common/etcdop/op/op_test.go
+++ b/internal/pkg/service/common/etcdop/op/op_test.go
@@ -21,7 +21,7 @@ type testOp struct {
 }
 
 type testValue struct {
-	Foo string
+	Foo string `json:"foo"`
 }
 
 type opTestCase[R any] struct {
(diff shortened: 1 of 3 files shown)


internal/pkg/service/common/etcdop/op/op_test.go
```diff
@@ -21,7 +21,7 @@ type testOp struct {
 }
 
 type testValue struct {
-	Foo string
+	Foo string `json:"foo"`
 }
```
michaljurecko, line 26: :man_shrugging: apparently it is not necessary here

michaljurecko (changes requested):

internal/pkg/service/common/etcdop/op/op_test.go
Matovidlo: Should we ommit the `test` files from this linter?

jachym-tousek-keboola (approved):

PR #1816, opened 2024-06-04
lint: Add new linters without code change. Add new disabled linters.

Matovidlo:

**Changes:**
- Add new litners without code changed.

-----------

### DIFF
--- a/build/ci/golangci.yml
+++ b/build/ci/golangci.yml
@@ -112,21 +112,27 @@ linters:
   disable-all: true
   enable:
     - asciicheck
+    - bidichk
     - bodyclose
     - contextcheck
+    - decorder
     - depguard
     - dogsled
     - dupl
     - durationcheck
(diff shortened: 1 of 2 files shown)


michaljurecko (commented):
:man_shrugging: I don't know which which are useful.

Matovidlo:
@michaljurecko About 3 linters are checking the style of code. Most of the linters should prevent unnecessary bugs from using wrong `//go:` declaratives or checks types passed to the json encoding functions etc. In the comment section I added others. I dont see any reason why we should not use new linters when we use new `go` version. It should help us and each linter has it's description here https://golangci-lint.run/usage/linters/
So the same reason as you update your libs you update your linters (better code/less code less bugs).

jachym-tousek-keboola (approved):

hosekpeter (approved):
LGTM

PR #1819, opened 2024-06-05
feat: Implement aggregation endpoint

jachym-tousek-keboola:
Jira: [keboola.atlassian.net/browse/PSGO-623](https://keboola.atlassian.net/browse/PSGO-623)

**Changes:**
- Implement aggregation endpoint
- Add e2e tests

-----------

### DIFF
--- a/internal/pkg/service/stream/aggregation/repository/repository.go
+++ b/internal/pkg/service/stream/aggregation/repository/repository.go
@@ -68,7 +68,7 @@ func New(d dependencies) *Repository {
 	}
 }
 
-func (r *Repository) SourcesWithSinksAndStatistics(ctx context.Context, sourceKeys []key.SourceKey) ([]*SourceWithSinks, error) {
+func (r *Repository) SourcesWithSinksAndStatistics(ctx context.Context, sourceKeys []key.SourceKey) (map[key.SourceKey]*SourceWithSinks, error) {
 	sourcesWithSinks, err := r.sinksForSources(ctx, sourceKeys)
 	if err != nil {
 		return nil, err
@@ -84,7 +84,7 @@ func (r *Repository) SourcesWithSinksAndStatistics(ctx context.Context, sourceKe
 		return nil, err
 	}
 
-	return maps.Values(sourcesWithSinks), nil
+	return sourcesWithSinks, nil
 }
 
 func (r *Repository) sinksForSources(ctx context.Context, sourceKeys []key.SourceKey) (map[key.SourceKey]*SourceWithSinks, error) {
--- a/internal/pkg/service/stream/api/service/aggregation.go
+++ b/internal/pkg/service/stream/api/service/aggregation.go
@@ -31,14 +31,12 @@ func (s *service) AggregateSources(ctx context.Context, d dependencies.BranchReq
 func (s *service) addSinksToAggregationResponse(ctx context.Context, d dependencies.BranchRequestScope, response *stream.AggregatedSourcesResult) error {
 	// Collect source IDs
 	var sourceKeys []key.SourceKey
-	sourceKeyToIndex := make(map[key.SourceKey]int)
-	for i, source := range response.Sources {
+	for _, source := range response.Sources {
 		sourceKey := key.SourceKey{
 			BranchKey: d.BranchKey(),
 			SourceID:  source.SourceID,
 		}
 		sourceKeys = append(sourceKeys, sourceKey)
-		sourceKeyToIndex[sourceKey] = i
 	}
 
 	// Get sinks for all the sources
@@ -47,9 +45,17 @@ func (s *service) addSinksToAggregationResponse(ctx context.Context, d dependenc
 		return err
 	}
 
-	for _, sourceWithSinks := range sourcesWithSinks {
-		// Find index of the source in response
-		sourceIndex := sourceKeyToIndex[sourceWithSinks.SourceKey]
+	// Add sinks to response
+	for _, source := range response.Sources {
+		sourceKey := key.SourceKey{
+			BranchKey: d.BranchKey(),
+			SourceID:  source.SourceID,
+		}
+
+		sourceWithSinks, ok := sourcesWithSinks[sourceKey]
+		if !ok {
+			continue
+		}
 
 		for _, sink := range sourceWithSinks.Sinks {
 			if sink == nil {
@@ -61,7 +67,6 @@ func (s *service) addSinksToAggregationResponse(ctx context.Context, d dependenc
 				return err
 			}
 
-			source := response.Sources[sourceIndex]
 			source.Sinks = append(source.Sinks, sinkResponse)
 		}
 	}


internal/pkg/service/stream/api/mapper/sink.go
```diff
@@ -17,25 +19,44 @@ func (m *Mapper) NewSinkStatisticsTotalResponse(result statistics.Aggregated) (r
...
 		UncompressedSize: uint64(value.UncompressedSize),
 	}
 }
 
-func timeValueToPointer(time utctime.UTCTime) *string {
-	var result *string
-	if !time.IsZero() {
-		value := time.String()
-		result = &value
+func timeToString(time *utctime.UTCTime) *string {
```
Matovidlo, line 64: `timeToStringPointer` Someone can use it and expects that it returns `""` as default.

internal/pkg/service/stream/aggregation/repository/repository.go
```diff
@@ -0,0 +1,198 @@
...
+		logger:     d.Logger().WithComponent("aggregation.repository"),
+		client:     d.EtcdClient(),
+		definition: d.DefinitionRepository(),
+		statistics: d.StatisticsRepository(),
+		storage:    d.StorageRepository(),
+	}
+}
+
+func (r *Repository) GetSourcesWithSinksAndStatistics(ctx context.Context, sourceKeys []key.SourceKey) ([]*SourceWithSinks, error) {
+	res, err := r.findSinksForSources(ctx, sourceKeys)
```
Matovidlo, line 72: Why do we call it `res` when it should be `sourcesWithSinks` for example? Looks better then. Also `map.Values(sourcesWithSinks)` is the result of `GetSourcesWithSinksAndStatistics`.

internal/pkg/service/stream/aggregation/repository/repository.go
```diff
@@ -0,0 +1,198 @@
...
+	return &Repository{
+		logger:     d.Logger().WithComponent("aggregation.repository"),
+		client:     d.EtcdClient(),
+		definition: d.DefinitionRepository(),
+		statistics: d.StatisticsRepository(),
+		storage:    d.StorageRepository(),
+	}
+}
+
+func (r *Repository) GetSourcesWithSinksAndStatistics(ctx context.Context, sourceKeys []key.SourceKey) ([]*SourceWithSinks, error) {
```
Matovidlo, line 71: In go we have `getter` using leading character uppercase. Therefore I would expect `SourcesWithSinksAndStatistics` to be a getter. No `Get` needed. 
Also return value is called the same so it looks convinient. :slightly_smiling_face:

internal/pkg/service/stream/aggregation/repository/repository.go
```diff
@@ -0,0 +1,198 @@
...
+
+	err = r.addFileStatisticsToAggregationResponse(ctx, res)
+	if err != nil {
+		return nil, err
+	}
+
+	return maps.Values(res), nil
+}
+
+func (r *Repository) findSinksForSources(ctx context.Context, sourceKeys []key.SourceKey) (map[key.SourceKey]*SourceWithSinks, error) {
```
Matovidlo, line 95: `sinksForSources`

internal/pkg/service/stream/aggregation/repository/repository.go
```diff
@@ -0,0 +1,198 @@
...
+			sinkKey := key.SinkKey{
+				SourceKey: sourceKey,
+				SinkID:    sink.SinkID,
+			}
+
+			txn.Merge(r.statistics.AggregateIn(sinkKey).OnResult(func(result *op.TxnResult[statistics.Aggregated]) {
+				sink.Statistics = &SinkStatistics{
+					Total: ptr.Ptr(result.Result()),
+				}
+			}))
```
Matovidlo, line 133: Cannot we do double `Merge` to prevent having 2 functions doing very simillar loop? It seems to be unoptimized
```
txn.Merge(r.statistics.AggregateIn(sinkKey).OnResult(func(result *op.TxnResult[statistics.Aggregated]) {
sink.Statistics = &SinkStatistics{
	Total: ptr.Ptr(result.Result()),
}
}))
txn.Merge(r.storage.File().ListRecentIn(sinkKey).ForEach(func(value model.File, header *iterator.Header) error {
sink.Statistics.Files = append(sink.Statistics.Files, &FileWithStatistics{
	File: ptr.Ptr(value),
})
return nil
}))
```

internal/pkg/service/stream/api/mapper/aggregation.go
```diff
@@ -0,0 +1,109 @@
...
+		Type:        entity.Type,
+		Name:        entity.Name,
+		Description: entity.Description,
+		Created:     m.NewCreatedResponse(entity.Created),
+		Version:     m.NewVersionResponse(entity.Version),
+		Deleted:     m.NewDeletedResponse(entity.SoftDeletable),
+		Disabled:    m.NewDisabledResponse(entity.Switchable),
+	}
+
+	// Type
```
Matovidlo, line 51: Remove? I see switch `out.Type`.

internal/pkg/service/stream/api/mapper/aggregation.go
```diff
@@ -0,0 +1,109 @@
...
+		Version:     m.NewVersionResponse(entity.Version),
+		Deleted:     m.NewDeletedResponse(entity.SoftDeletable),
+		Disabled:    m.NewDisabledResponse(entity.Switchable),
+	}
+
+	if entity.Statistics.Total != nil {
+		totals := m.NewSinkStatisticsTotalResponse(*entity.Statistics.Total)
+		files := api.SinkFiles{}
+		for _, file := range entity.Statistics.Files {
+			sinkFile := m.NewSinkFile(*file.File)
```
Matovidlo, line 82: Possible panic when `file` = `nil`?

internal/pkg/service/stream/api/mapper/aggregation.go
```diff
@@ -0,0 +1,109 @@
...
+			files = append(files, sinkFile)
+		}
+		out.Statistics = &api.AggregationStatistics{
+			Total:  totals.Total,
+			Levels: totals.Levels,
+			Files:  files,
+		}
+	}
+
+	// Type
```
Matovidlo, line 95: Remove commentary

internal/pkg/service/stream/api/service/aggregation.go
```diff
@@ -3,11 +3,59 @@ package service
...
+
+func (s *service) addSinksToAggregationResponse(ctx context.Context, d dependencies.BranchRequestScope, response *stream.AggregationSourcesResult) error {
+	sourceKeys := make(map[key.SourceKey]int)
+	for i, source := range response.Sources {
+		sourceKey := key.SourceKey{
+			BranchKey: d.BranchKey(),
+			SourceID:  source.SourceID,
+		}
+		sourceKeys[sourceKey] = i
+	}
```
Matovidlo, line 41: So we create map of `sourceKeys` to deduplicate it and then we cast it into slice `[]key.SourceKey`. 
Two things

1. Is it possible to write somewhere that we need to deduplicate the entries? Or maybe move it to the top level layer so we work only with deduplicated entries.
2. Do we need to cast `map` into `slice` ? It need to loop over fields to actually create slice. Cannot we use the keys in the map? The loop in `GetSourcesWithSinksAndStatistics` will be change into `for key, _ := range sourceKeys`
https://github.com/keboola/keboola-as-code/pull/1819/commits/266712c97ba90386318011029620a74eea733223#diff-8c5b3a594c9443cdd704ef50193f2fd6f6c9bb148d52fe716c85a6db31da8df6R99

we will reduce numerous casting `map -> slice -> loop` into `map -> loop`. This would have effect when we would like to preserve the values and the `GetSorucesWithSinksAndStatistics` would have side effect on `sourceKeys`

internal/pkg/service/stream/api/service/aggregation.go
```diff
@@ -3,11 +3,59 @@ package service
...
+
+	sourcesWithSinks, err := d.AggregationRepository().GetSourcesWithSinksAndStatistics(ctx, maps.Keys(sourceKeys))
+	if err != nil {
+		return err
+	}
+
+	for _, sourceWithSinks := range sourcesWithSinks {
+		for _, sink := range sourceWithSinks.Sinks {
+			sink, err := s.mapper.NewAggregationSinkResponse(*sink)
```
Matovidlo, line 50: Possible panic?  + shadowing. Try to rename the `sink` argument or return value `sink` into `sinkResponse` or so.

internal/pkg/service/stream/api/service/aggregation.go
```diff
@@ -3,11 +3,59 @@ package service
...
+			if err != nil {
+				return err
+			}
+
+			source := response.Sources[sourceKeys[sourceWithSinks.SourceKey]]
```
Matovidlo, line 55: `sourceKeys[sourceWithSinks.SourceKey]` could be extracted into outer loop. Therefore instead of `sourceWithSinks.Sinks` access into map it will be reduced into `single` access.
It may be possible that the compiler would optimize it however I would be rather explicit than implicit.

Matovidlo (commented):
Some styling issues and questions regarding code effectivity.

internal/pkg/service/stream/api/service/aggregation.go
```diff
@@ -31,29 +29,40 @@ func (s *service) AggregateSources(ctx context.Context, d dependencies.BranchReq
...
 			SourceID:  source.SourceID,
 		}
-		sourceKeys[sourceKey] = i
+		sourceKeys = append(sourceKeys, sourceKey)
+		sourceKeyToIndex[sourceKey] = i
 	}
 
-	sourcesWithSinks, err := d.AggregationRepository().GetSourcesWithSinksAndStatistics(ctx, maps.Keys(sourceKeys))
+	// Get sinks for all the sources
+	sourcesWithSinks, err := d.AggregationRepository().SourcesWithSinksAndStatistics(ctx, sourceKeys)
```
Matovidlo, line 45: Why can't we work with the map? Do we have to preserve order of input `response.Sources`?

internal/pkg/service/stream/aggregation/repository/repository.go
```diff
@@ -0,0 +1,198 @@
...
+	}
+
+	return maps.Values(res), nil
+}
+
+func (r *Repository) findSinksForSources(ctx context.Context, sourceKeys []key.SourceKey) (map[key.SourceKey]*SourceWithSinks, error) {
+	res := make(map[key.SourceKey]*SourceWithSinks)
+
+	txn := op.Txn(r.client)
+	for _, sourceKey := range sourceKeys {
```
hosekpeter, line 94: Shouldn't we check if source exists? 🤔

internal/pkg/service/stream/api/service/aggregation.go
Matovidlo: It looks better.

Matovidlo (approved):
LGTM

hosekpeter (approved):

PR #1824, opened 2024-06-07
refactor: Use module instead of 3 intialization of variables.

Matovidlo:
**Changes:**
- refactor code to be simpler.

-----------


michaljurecko (approved):
LGTM

PR #1826, opened 2024-06-07
feat: Improve etcdlogger, log request options

michaljurecko:

--------------

**Changes:**
- Details (`count only`, `keys only`, `serializable`) of etcd operation are logged.
- Semantic of etcd `Get*` has been improved, you know if the missing key is `nil` result or error.

-----------

Commend per commits.


internal/pkg/utils/etcdlogger/etcdlogger.go
```diff
@@ -169,6 +169,15 @@ func (v *kvWrapper) start(op *etcd.Op, opName, key, value string) string {
...
 		}
+		if op.IsCountOnly() {
+			out.WriteString(" | count only")
+		}
+		if op.IsKeysOnly() {
+			out.WriteString(" | keys only")
+		}
+		if op.IsSerializable() {
+			out.WriteString(" | serializable")
+		}
```
michaljurecko, line 180: etcd query can have these options set, I add them to logs, you can find the meaning of individual options in the etcd documentation.

internal/pkg/service/common/etcdop/iterator/iterator_test.go
```diff
@@ -87,7 +87,7 @@ func TestIterator(t *testing.T) {
   ➡️  THEN:
   001 ➡️  GET ["some/prefix/", "some/prefix0")
 ✔️  TXN | succeeded: true | rev: %d
-➡️  GET ["some/prefix/foo004", "some/prefix0") | rev: %d
+➡️  GET ["some/prefix/foo004", "some/prefix0") | rev: %d | serializable
 ✔️  GET ["some/prefix/foo004", "some/prefix0") | rev: %d | count: 2
```
michaljurecko, line 91: Updated iterator tests.

`2+` page is loaded with locked revision and `serializable` flag. We don't need cluster consensus to read the specific revision.

From docs:
> By default, Range is linearizable; it reflects the current consensus of the cluster. For better performance and availability, in exchange for possible stale reads, a serializable range request is served locally without needing to reach consensus with other nodes in the cluster.

internal/pkg/service/common/etcdop/key.go
```diff
@@ -56,8 +56,8 @@ func (v Key) Exists(client etcd.KV, opts ...etcd.OpOption) op.BoolOp {
 	)
 }
 
-func (v Key) Get(client etcd.KV, opts ...etcd.OpOption) op.GetOneOp {
-	return op.NewGetOneOp(
+func (v Key) Get(client etcd.KV, opts ...etcd.OpOption) op.WithResult[*op.KeyValue] {
+	return op.NewForType(
 		client,
 		func(_ context.Context) (etcd.Op, error) {
 			return etcd.OpGet(v.Key(), opts...), nil
```
michaljurecko, line 63: I removed old `op.GetOneOp` types, we have now the `WithResult` generic type for it. 

This is `Get` operation on untyped key, there is only one `Get` method.
Bellow are `Get*` opeations on TYPED key.

internal/pkg/service/common/etcdop/key.go
```diff
@@ -171,7 +173,36 @@ func (v KeyT[T]) GetKV(client etcd.KV, opts ...etcd.OpOption) op.WithResult[*op.
 	)
 }
 
-func (v KeyT[T]) Get(client etcd.KV, opts ...etcd.OpOption) op.WithResult[T] {
+// GetOrNil gets a decoded value of the key as pointer.
+// If the key is missing, result is nil pointer.
+func (v KeyT[T]) GetOrNil(client etcd.KV, opts ...etcd.OpOption) op.WithResult[*T] {
+	return op.NewForType(
+		client,
```
michaljurecko, line 180: After the refactoring, we have following get methods:
- `GetOrNil` - `op.WithResult[*T]`
- `GetOrErr` - `op.WithResult[T]`
- `GetOrEmpty` - `op.WithResult[T]`
- `GetKV` - `op.WithResult[*op.KeyValueT[T]]`

So you immediately know how it works from the code. 

For example, `GetOrErr` is used in the definition repository, because the missing entity is a problem.

For statistics, `GetOrEmpty` makes sense.

internal/pkg/service/stream/definition/repository/branch/branch_create.go
```diff
@@ -18,29 +18,29 @@ import (
 func (r *Repository) Create(input *definition.Branch, now time.Time, by definition.By) *op.AtomicOp[definition.Branch] {
 	k := input.BranchKey
 	var created definition.Branch
-	var deleted *op.KeyValueT[definition.Branch]
+	var deleted *definition.Branch
```
michaljurecko, line 21: Now, we can use `GetOrNil`, instead of more complicated `GetKV` (we don't need KV metadata).

internal/pkg/service/common/etcdop/key_test.go
```diff
@@ -116,8 +116,18 @@ func TestTypedKeyOperations(t *testing.T) {
...
 
-	// Get - not found
-	result, err := k.Get(client).Do(ctx).ResultOrErr()
+	// GetOrNil - not found
+	resultPtr, err := k.GetOrNil(client).Do(ctx).ResultOrErr()
+	require.NoError(t, err)
+	assert.Nil(t, resultPtr)
+
+	// GetOrEmpty - not found
+	result, err := k.GetOrEmpty(client).Do(ctx).ResultOrErr()
```
Matovidlo, line 125: One would expect that `GetOrEmpty` never returns error. I understand that there are possibilities to return error, I was just wandering about better naming.

Matovidlo (approved):
LGTM

internal/pkg/service/common/etcdop/key_test.go
michaljurecko: There can always be a network error, etc., if you have a better idea, tell me.

hosekpeter (approved):

internal/pkg/service/common/etcdop/iterator/iterator_test.go
michaljurecko: :telephone_receiver:

PR #1830, opened 2024-06-10
ci: Fix windows tools installation

Matovidlo:
Jira: [PSGO-665](https://keboola.atlassian.net/browse/PSGO-665)

**Changes:**
- ci: Fix windows tools installation by not using `curl` but `go install`.
- ci: Remove restore-keys part so the tools are cached only by one key. REASON: when changing tool in `tools.sh` we want to use the new tool, not the old ones.

-----------


michaljurecko:
@jachym-tousek-keboola can you please finalize it, @Matovidlo is OOO.

PR #1832, opened 2024-06-10
Create new tracer wrapper to not propagate data apps errors into DD

Matovidlo:
Jira: [PSGO-637](https://keboola.atlassian.net/browse/PSGO-637)

**Changes:**
- Modify request info middleware to contain request. 
- Spans of proxied requests should be always set to Ok instead of error when app is broken (waked up).
- Create tests for checking the `Status` of Spans.

-----------

[PSGO-637]: https://keboola.atlassian.net/browse/PSGO-637?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/internal/pkg/service/appsproxy/proxy/proxy_test.go
+++ b/internal/pkg/service/appsproxy/proxy/proxy_test.go
@@ -25,6 +25,11 @@ import (
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 	"github.com/umisama/go-regexpcache"
+	"go.opentelemetry.io/otel/attribute"
+	"go.opentelemetry.io/otel/codes"
+	tracesdk "go.opentelemetry.io/otel/sdk/trace"
+	"go.opentelemetry.io/otel/sdk/trace/tracetest"
+	"go.opentelemetry.io/otel/trace"
 	"go.uber.org/atomic"
 	"nhooyr.io/websocket"
 	"nhooyr.io/websocket/wsjson"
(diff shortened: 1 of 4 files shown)


internal/pkg/service/appsproxy/proxy/proxy_test.go
```diff
@@ -119,6 +126,94 @@ func TestAppProxyRouter(t *testing.T) {
 			},
 			expectedNotifications: map[string]int{},
 			expectedWakeUps:       map[string]int{},
+			expectedSpans: []tracetest.SpanStub{
+				{
+					Name:     "keboola.go.common.dependencies.NewBaseScope",
+					SpanKind: 1,
```
michaljurecko, line 132: There are options to filter out unwanted spans in tests, do we need all these spans in the asset?

internal/pkg/service/appsproxy/proxy/server.go
```diff
@@ -86,8 +135,10 @@ func NewHandler(d dependencies.ServiceScope) http.Handler {
 		middleware.RequestInfo(),
 		middleware.Filter(middlewareCfg),
 		middleware.Logger(d.Logger()),
+		// Mandatory middleware when used in combination with newTracerProviderWrapper
+		middleware.Request(),
```
michaljurecko, line 139: There is already the `RequestInfo` middleware, I would move the code there, and not create a new middleware.

internal/pkg/service/appsproxy/proxy/proxy_test.go
Matovidlo: I would rather check all spans in order to know `name` etc. No more than 1 test for this case is needed. It is always possible to extend it with filter in future.

michaljurecko: ok

internal/pkg/service/appsproxy/proxy/server.go
Matovidlo: Fixed.

michaljurecko (approved):
LGTM

hosekpeter (approved):
LGTM

PR #1833, opened 2024-06-12
lint: Add prealoc linter to reduce realocation

Matovidlo:
Jira: XXXXX

**Changes:**
- ...

-----------


internal/pkg/env/map.go
```diff
@@ -67,7 +67,7 @@ func (m *Map) ToString() (string, error) {
 }
 
 func (m *Map) ToSlice() []string {
-	var out []string
+	out := make([]string, 0, len(m.Keys()))
 	for _, k := range m.Keys() {
```
michaljurecko, line 71: It has no measurable impact, but it can be here.

pkg/lib/operation/project/remote/table/preview/operation.go
```diff
@@ -77,7 +77,7 @@ func Run(ctx context.Context, o Options, d dependencies) (err error) {
 }
 
 func getPreviewOptions(o *Options) []keboola.PreviewOption {
-	var opts []keboola.PreviewOption
+	opts := make([]keboola.PreviewOption, 0, 4+len(o.WhereFilters)+len(o.Order))
 	if o.Limit > 0 {
 		opts = append(opts, keboola.WithLimitRows(o.Limit))
```
michaljurecko, line 82: `4+len(o.WhereFilters)+len(o.Order)` ?

Are you sure you want to have this here, what is the benefit?

Matovidlo: You have 4 options, then filters and then order. If you pre alocate the slice for expected amount of entries, it does not need to realocate and it is faster. Ofc this is just optimization of CLI, does not have very huge effect but also why don't you allocate exact number of the entries that you are expected to insert using `append` ?

jachym-tousek-keboola (approved):

PR #1835, opened 2024-06-12
CLI: Fix ignoring branch name for diff

hosekpeter:
Jira: [PSGO-664](https://keboola.atlassian.net/browse/PSGO-664)

**Changes:**
- ignored branch name (if allowTargetEnv = true)
- added tests for this issue

-----------

[PSGO-664]: https://keboola.atlassian.net/browse/PSGO-664?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


internal/pkg/diff/diff.go
```diff
@@ -18,13 +18,26 @@ import (
...
 
 type Differ struct {
-	objects   model.ObjectStates
-	results   []*Result                                 // diff results
-	typeCache map[typeName][]*reflecthelper.StructField // reflection cache
-	errors    errors.MultiError
+	objects        model.ObjectStates
+	results        []*Result                                 // diff results
+	typeCache      map[typeName][]*reflecthelper.StructField // reflection cache
+	allowTargetEnv bool
```
Matovidlo, line 39: Put it as last in struct, so it is obvious that it is an options for `Differ`. Maybe add commentary `// option` ?

pkg/lib/operation/template/sync/diff/create/operation.go
```diff
@@ -1,29 +1,21 @@
 package create
 
-import (
-	"context"
+// type Options struct {
```
Matovidlo, line 3: Why is this commented?

test/cli/allow-target-env/push-changed-config-values/args
```diff
@@ -0,0 +1 @@
+push "my change description" --force --storage-api-token %%TEST_KBC_STORAGE_API_TOKEN%%
```
Matovidlo, line 1: Do we have to use `--force`? If yes change name of  test to `push-force-changed-config-values` and create another test without `--force`.

Matovidlo (changes requested):

pkg/lib/operation/template/sync/diff/create/operation.go
hosekpeter: This part of the code was unused. I don't know if it's ready for the future.

Matovidlo: Hm I would rather remove it instead of commenting it. If it is unsed probably does not need to be there at all. Put this part into your 1:1 with Michal to go through the `lib/operation` module what should be there and what not.

Matovidlo (approved):
LGTM. Just check with Michal the commented code, even after merging

internal/pkg/diff/diff.go
```diff
@@ -18,13 +18,26 @@ import (
...
+type Config struct {
+	allowTargetEnv bool
+}
+
+func WithAllowTargetEnv(allowTargetEnv bool) Option {
+	return func(c *Config) {
+		c.allowTargetEnv = allowTargetEnv
+	}
+}
+
```
michaljurecko, line 32: Here should be `WithIgnoreBranchName`, this part of KAC has nothing directly to do with "allowTargetEnv".

hosekpeter: [8855283](https://github.com/keboola/keboola-as-code/pull/1835/commits/8855283ad3bcc9f6c1507a6b8e768a6c2472cb82)

PR #1836, opened 2024-06-12
feat: Writers chain optimalization in: CSV writer, disk sync wait mechanism, compression, etc.

michaljurecko:
Jira: XXXXX

**Changes:**
- ...

-----------


internal/pkg/service/stream/storage/level/local/writer/writer.go
```diff
@@ -1,42 +1,360 @@
...
+	}
+
+	// Add a buffer before the file
+	{
+		if cfg.FileBuffer > 0 {
+			w.chain.PrependWriter(func(w io.Writer) io.Writer {
+				return limitbuffer.New(w, int(cfg.FileBuffer.Bytes()))
+			})
+		}
+	}
```
michaljurecko, line 158: Important buffer before the file, ... so when we are waiting for the sync to the file - new CSV rows can be generated, compresed and prepared in the buffer.

internal/pkg/service/stream/storage/level/local/writer/writer.go
```diff
@@ -1,42 +1,360 @@
...
+	{
+		backupPath := filepath.Join(dirPath, CompressedSizeFile)
+		_, err = w.chain.PrependWriterOrErr(func(writer io.Writer) (out io.Writer, err error) {
+			w.compressedMeter, err = size.NewMeterWithBackupFile(ctx, clk, logger, writer, backupPath, cfg.Statistics.DiskSyncInterval)
+			return w.compressedMeter, err
+		})
+		if err != nil {
+			return nil, err
+		}
+	}
```
michaljurecko, line 170: Size of the compressed data is measured before the buffer, so we know how much is waiting for sync to disk.

internal/pkg/service/stream/storage/level/local/writer/writer.go
michaljurecko: (chain in composed from the end, from the output file)

internal/pkg/service/stream/storage/level/local/writer/writer.go
```diff
@@ -1,42 +1,360 @@
...
+			if err != nil {
+				return nil, err
+			}
+
+			// Add a buffer before compression writer, if it is not included in the writer itself
+			if cfg.InputBuffer > 0 && !compConfig.HasWriterInputBuffer() {
+				w.chain.PrependWriter(func(w io.Writer) io.Writer {
+					return limitbuffer.New(w, int(cfg.InputBuffer.Bytes()))
+				})
+			}
```
michaljurecko, line 188: Important buffer before the compression, BUT! only, if there is no buffer inside the compression writer.

CSV rows can be pre-generated to the buffer before the compression writer.

internal/pkg/service/stream/storage/level/local/writer/writer.go
```diff
@@ -1,42 +1,360 @@
...
+	// It is entrypoint of the writers chain.
+	{
+		w.formatWriter, err = formatWriterFactory(cfg, w.chain, w.slice)
+		if err != nil {
+			return nil, err
+		}
+
+		// Flush/Close the file format writer at first
+		w.chain.PrependFlusherCloser(w.formatWriter)
+	}
```
michaljurecko, line 220: The start of the chain is `FormarWriter = csv.Writer`.

internal/pkg/service/stream/storage/level/local/writer/writer.go
```diff
@@ -1,42 +1,360 @@
...
+
+	// Dispatch "open" event
+	if err = volumeEvents.dispatchOnWriterOpen(w); err != nil {
+		return nil, err
+	}
```
michaljurecko, line 225: `EventsWriter` from the old code was merged with the `writer` struct.

internal/pkg/service/stream/storage/level/local/writer/writer.go
```diff
@@ -1,42 +1,360 @@
...
+
+	notifier := w.syncer.Notifier()
+
+	// Increments number of high-level writes in progress
+	w.acceptedWrites.Add(timestamp, 1)
+
+	// Wait for sync and return sync error, if any
+	if err := notifier.Wait(); err != nil {
+		return err
+	}
```
michaljurecko, line 258: If enabled, the `WriteRecord` method waits until the record is written to the disk, so we provide strong guarantees that data will not be lost. It would be even better if we also had replication on several disks.

michaljurecko:
Moved :arrow_right: https://github.com/keboola/keboola-as-code/pull/1842

PR #1841, opened 2024-06-13
fix: Upload command

hosekpeter:
Jira: [PSGO-667](https://keboola.atlassian.net/browse/PSGO-667)

**Changes:**
- added more tests
- changed value for empty string `[] -> nil`

-----------
Slack: 
https://keboolaglobal.slack.com/archives/C055WKPHYSD/p1718257832123069?thread_ts=1718257769.496969&cid=C055WKPHYSD

https://keboolaglobal.slack.com/archives/C055WKPHYSD/p1718179324954169?thread_ts=1718178052.148319&cid=C055WKPHYSD

[PSGO-667]: https://keboola.atlassian.net/browse/PSGO-667?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Matovidlo:
@hosekpeter Missing these cases:

1. table exists = false, --columns
2. table exists = false, --file-without-headers
3. table exists = false, --columns, --file-without-headers
4. table exists = true, CSV columns - subset of columns of the existing table + different order (always included --file-without-header)
5. table exists = true, --columns - subset of columns of the existing table + different order (always included --file-without-header)

Try to go once through the test cases. I would expect same amount of tests for both `--columns` flag and `CSV file` without `--columns` flag with combination of `--file-without-headers` + `table` exists/does not exists.

Matovidlo (commented):
Missing these cases:

1. table exists = false, --columns
2. table exists = false, --file-without-headers
3. table exists = false, --columns, --file-without-headers
4. table exists = true, CSV columns - subset of columns of the existing table + different order (always included --file-without-header)
5. table exists = true, --columns - subset of columns of the existing table + different order (always included --file-without-header)

Try to go once through the test cases. I would expect same amount of tests for both `--columns` flag and `CSV file` without `--columns` flag with combination of `--file-without-headers` + `table` exists/does not exists.

hosekpeter:
I know about that. @Matovidlo

pkg/lib/operation/project/remote/table/import/operation.go
```diff
@@ -43,36 +46,42 @@ func Run(ctx context.Context, o Options, d dependencies) (err error) {
 			return err
 		}
 
-		_, err = d.KeboolaProjectAPI().CreateTableFromFileRequest(o.TableKey, o.FileKey, getCreateOptions(&o)...).Send(ctx)
-		if err != nil {
-			rb.Invoke(ctx)
-			return err
+		if (o.Columns != nil && !o.WithoutHeaders) || (o.WithoutHeaders && o.Columns == nil) {
```
Matovidlo, line 49: Why did we connect `columns` and `headers` flags? I can have csv file without headers and specified `columns` flag. Now it says that I cannot use because of missing `without headers` flag. Do we want to return error or just ignore the `without headers` as before?
I would say that only `2nd` condition is correct. First one is too restrictive, the csv can or does not need to have header.

pkg/lib/operation/project/remote/table/import/operation.go
```diff
@@ -115,3 +124,59 @@ func checkTableExists(ctx context.Context, d dependencies, tableKey keboola.Tabl
...
+	// Read the first line (headers)
+	headers, err := reader.Read()
+	if err != nil {
+		return nil, err
+	}
+
+	return convertHeadersToColumns(headers), nil
+}
+
+func printInfo(ctx context.Context, o Options, d dependencies, isCreated bool) {
```
Matovidlo, line 176: This is antipattern function. Based on `isCreated` it prints something, or something else, just extract it into code. Also it has single usage.

test/cli/remote-table-upload/table-create-columns-flag/args
```diff
@@ -0,0 +1 @@
+remote table upload --storage-api-host %%TEST_KBC_STORAGE_API_HOST%% --storage-api-token %%TEST_KBC_STORAGE_API_TOKEN%% --primary-key Id --file-name data.csv --file-tags test in.c-my-bucket.data data.csv --columns Id,Name,Region,Status,First_Order
```
Matovidlo, line 1: This usage is ok. I would not restrict it. You enforce user's to adjust the `.csv` file to remove `header` section.

test/cli/remote-table-upload/table-create-file-without-headers/in/data.csv
```diff
@@ -0,0 +1,4 @@
+"f030ed64cbc8babbe50901a26675a2ee","CSK Auto","US West","Active","2015-01-23"
```
Matovidlo, line 1: Add header

test/cli/remote-table-upload/table-exist-columns-flag-diff-order/args
```diff
@@ -0,0 +1 @@
+remote table upload --storage-api-host %%TEST_KBC_STORAGE_API_HOST%% --storage-api-token %%TEST_KBC_STORAGE_API_TOKEN%% --primary-key Id --file-name data.csv --file-tags test in.c-my-bucket.data data.csv --columns Status,Region,Id,Name,First_Order
```
Matovidlo, line 1: You still miss the test. Here is full set of columns, but we miss test where we have subset of columns in different order.

test/cli/remote-table-upload/table-create-file-without-headers/in/data.csv
hosekpeter: Why? I think there is no reason to add headers because this is a test that the csv file does not have column names.

test/cli/remote-table-upload/table-exist-columns-flag-diff-order/args
hosekpeter: [7dbf765](https://github.com/keboola/keboola-as-code/pull/1841/commits/7dbf765a29ed8b01a93b947b5744f7fdaa44536f)

test/cli/remote-table-upload/table-create-columns-flag/args
hosekpeter: [7dbf765](https://github.com/keboola/keboola-as-code/pull/1841/commits/7dbf765a29ed8b01a93b947b5744f7fdaa44536f)

pkg/lib/operation/project/remote/table/import/operation.go
hosekpeter: [7916294](https://github.com/keboola/keboola-as-code/pull/1841/commits/7916294c1dee38c2b2739525ce8e4098de12a529)

Matovidlo (approved):
LGTM

PR #1842, opened 2024-06-13
feat: Improve CSV/compression writers performance

michaljurecko:

--------------

**Changes:**
- Improved CSV/compression writers throughput.
- Reduced allocated memory size.
- Reduced number of allocations per write.
- **Even before, it was quite fast, but I completed some WIP optimizations from the past.**

-----------

-------------

**[FULL BENCHMARKS](https://gist.github.com/michaljurecko/5ec5cfdf8af247aaf7c384a660670649)**

------------------------------

**Benchmark Summary**

Input Data Write Speed: `+60-300%`

image

- 🔵  `compression=None,sync=None`
  - The most simple case, no compression, no waiting for the sync.
  - `1,4M/s` x `1kB` record
- 🟢 `compression=GZIP_Parallel_DefaultCompression,sync=ToDisk,wait=true`
  - Default configuration, parallel GZIP compression, waiting for the sync to the disk.
  - `500M/s` x `1kB` record

---------------------------------

Memory allocations count per write: `=0`, 

image

(the number of allocations does not increase with the number of writes)

---------------------------------

AVG allocated memory per write: `-80%`

image

(the allocated memory does not increase with the number of writes)


internal/pkg/service/stream/storage/level/local/writer/format/csv/fastcsv/fastcsv.go
```diff
@@ -0,0 +1,54 @@
...
+	// The sync.Pool is internally implemented using per-processor local pools.
+	// When goroutine is scheduled to run on a specific thread associated with specific processor
+	// and will try to retrieve an object from the pool, sync.Pool will first look in the current processor local pool.
+	// Read more: https://victoriametrics.com/blog/tsdb-performance-techniques-sync-pool/
+	<-p.sem
+	w := p.pool.Get().(*writer)
+	err := w.WriteRow(cols)
+	p.pool.Put(w)
+	p.sem <- struct{}{}
+	return err
```
michaljurecko, line 53: `p.sem` is used to limit number of parallel writers (to number of CPUs, by default).

`struct{}` is a token, number of the tokens in the channel buffer match number of free writers

internal/pkg/service/stream/storage/level/local/writer/base.go
```diff
@@ -1,105 +0,0 @@
...
-
-type syncer = disksync.Syncer
-
-// BaseWriter implements common logic for all file types.
-type BaseWriter struct {
-	*chain
-	*syncer
-	events   *Events
-	logger   log.Logger
-	slice    *model.Slice
```
michaljurecko, line 26: Originally, there was `BaseWriter` and `CSVWriter` was using it.

Now, there is `Writer`, and its dependency is `FormatWriterFactory`.

The `Writer` creates on initialization the `FormatWriter` (CSV writer) and the uses it.

internal/pkg/service/stream/storage/level/local/writer/writer.go
```diff
@@ -1,42 +1,360 @@
...
+			if w.completedWrites != nil {
+				_ = w.completedWrites.Close()
+			}
+		}
+	}()
+
+	// Create counters
+	{
+		// In progress writes counter
+		w.acceptedWrites = count.NewCounter()
```
michaljurecko, line 135: Following code is composition of the writers chain, each part of the chain is imported and also their order.

Small change in parameters/logic can reduce the throughput x times.

internal/pkg/service/stream/storage/level/local/writer/writer.go
```diff
@@ -1,42 +1,360 @@
...
+	}
+
+	// Add a buffer before the file
+	{
+		if cfg.FileBuffer > 0 {
+			w.chain.PrependWriter(func(w io.Writer) io.Writer {
+				return limitbuffer.New(w, int(cfg.FileBuffer.Bytes()))
+			})
+		}
+	}
```
michaljurecko, line 158: Important buffer before the file, ... so when we are waiting for the sync to the file - new CSV rows can be generated, compresed and prepared in the buffer.

(chain in composed from the end, from the output file, to the CSV writer on the start)

internal/pkg/service/stream/storage/level/local/writer/writer.go
```diff
@@ -1,42 +1,360 @@
...
+	{
+		backupPath := filepath.Join(dirPath, CompressedSizeFile)
+		_, err = w.chain.PrependWriterOrErr(func(writer io.Writer) (out io.Writer, err error) {
+			w.compressedMeter, err = size.NewMeterWithBackupFile(ctx, clk, logger, writer, backupPath, cfg.Statistics.DiskSyncInterval)
+			return w.compressedMeter, err
+		})
+		if err != nil {
+			return nil, err
+		}
+	}
```
michaljurecko, line 170: Size of the compressed data is measured before the buffer, so we know how much is waiting for sync to disk.

internal/pkg/service/stream/storage/level/local/writer/writer.go
```diff
@@ -1,42 +1,360 @@
...
+			if err != nil {
+				return nil, err
+			}
+
+			// Add a buffer before compression writer, if it is not included in the writer itself
+			if cfg.InputBuffer > 0 && !compConfig.HasWriterInputBuffer() {
+				w.chain.PrependWriter(func(w io.Writer) io.Writer {
+					return limitbuffer.New(w, int(cfg.InputBuffer.Bytes()))
+				})
+			}
```
michaljurecko, line 188: Important buffer before the compression, BUT! only, if there is no buffer inside the compression writer.

CSV rows can be pre-generated to the buffer before the compression writer.

internal/pkg/service/stream/storage/level/local/writer/writer.go
```diff
@@ -1,42 +1,360 @@
...
+	// It is entrypoint of the writers chain.
+	{
+		w.formatWriter, err = formatWriterFactory(cfg, w.chain, w.slice)
+		if err != nil {
+			return nil, err
+		}
+
+		// Flush/Close the file format writer at first
+		w.chain.PrependFlusherCloser(w.formatWriter)
+	}
```
michaljurecko, line 220: The start of the chain is `FormarWriter = csv.Writer`.

internal/pkg/service/stream/storage/level/local/writer/writer.go
```diff
@@ -1,42 +1,360 @@
...
+
+	// Dispatch "open" event
+	if err = volumeEvents.dispatchOnWriterOpen(w); err != nil {
+		return nil, err
+	}
+
+	return w, nil
```
michaljurecko, line 227: `EventsWriter` from the old code was merged with the `writer` struct.

internal/pkg/service/stream/storage/level/local/writer/writer.go
```diff
@@ -1,42 +1,360 @@
...
+
+	notifier := w.syncer.Notifier()
+
+	// Increments number of high-level writes in progress
+	w.acceptedWrites.Add(timestamp, 1)
+
+	// Wait for sync and return sync error, if any
+	if err := notifier.Wait(); err != nil {
+		return err
+	}
```
michaljurecko, line 258: If enabled, the `WriteRecord` method waits until the record is written to the disk, so we provide strong guarantees that data will not be lost. It would be even better if we also had replication on several disks.

internal/pkg/service/stream/storage/level/local/writer/event.go
```diff
@@ -1,62 +1,20 @@
...
 )
 
-type writer = Writer
-
-// EventWriter adds dispatching of the Events for the underlying writer.
-type EventWriter struct {
-	writer
-	events *Events
-}
-
```
michaljurecko, line 16: `EventsWriter` from the old code was merged with the `writer` struct.

internal/pkg/service/common/task/orchestrator/orchestrator.go
```diff
@@ -34,6 +34,9 @@ func (o orchestrator[T]) start() <-chan error {
 		defer o.logger.Info(o.node.ctx, "stopped")
 		b := newRetryBackoff()
 
+		// Copy variable, it is set to nil later, to signalize successful initialization
+		initDone := initDone
+
 		for {
```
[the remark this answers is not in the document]
Matovidlo: I think this will be reported by linter in future so we will have to fix it. Probably it will be enough to set it into another variable .

scripts/tools.sh
```diff
@@ -70,6 +70,12 @@ then
...
 fi
 
+# benchstat
+if ! command -v benchstat &> /dev/null
+then
+  go install golang.org/x/perf/cmd/benchstat@latest
+fi
+
 # go-mod-upgrade
 if ! command -v go-mod-upgrade &> /dev/null
```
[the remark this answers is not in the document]
Matovidlo: Maybe add it into confluence doc? Also when I see the benchmarks, I would expect that they would run in CI so we know whether the performance is downgraded due to future additions into function etc. Ofc this PR does not need to address it but I would like to open a discussion on this topic.

internal/pkg/service/stream/config/config_test.go
```diff
@@ -132,9 +132,11 @@ storage:
...
                     checkInterval: 5ms
                     # Written records count to trigger sync. Validation rules: min=0,max=1000000,required_if=Mode disk,required_if=Mode cache
-                    countTrigger: 500
-                    # Written size to trigger sync. Validation rules: maxBytes=100MB,required_if=Mode disk,required_if=Mode cache
-                    bytesTrigger: 1MB
+                    countTrigger: 10000
+                    # Size of buffered uncompressed data to trigger sync. Validation rules: maxBytes=500MB,required_if=Mode disk,required_if=Mode cache
+                    uncompressedBytesTrigger: 1MB
+                    # Size of buffered compressed data to trigger sync. Validation rules: maxBytes=100MB,required_if=Mode disk,required_if=Mode cache
+                    compressedBytesTrigger: 256KB
```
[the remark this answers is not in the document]
Matovidlo: Could you explain it in more detail? How do we know how to setup correct compression ratios based on various disks? `Compressed` should be in size small, therefore easier for disk to write. However also disk works with efficient sizes of files such as it is easier to save 10 files with 2MB that are compressed than 1 file with 40MB uncompressed. However we move the complexity from disk to CPU which is faster than disk (due to uncompression of 10 files for example).

internal/pkg/service/stream/storage/level/local/writer/format/csv/fastcsv/fastcsv.go
```diff
@@ -0,0 +1,47 @@
+// Package fastcsv provides parallel writer/formatter for CSV records.
+//   - There is WritersPool with fixed number of writers.
+//   - The io.Writer.Write method of the out writer is always called with the whole CSV row.
+//   - There is no built-in synchronization of the output, so the io.Writer.Write method must be protected externally.
+//   - The values are always quoted, so it is possible to generate row in single pass, without scanning the column value.
+//   - The resulting CSV is therefore a bit larger, but we rely on the compression that may follow in the writers chain.
+package fastcsv
```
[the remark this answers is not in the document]
Matovidlo: Why is row by row writing of CSV more perfomant than stdlib `csv` package?

internal/pkg/service/stream/storage/level/local/writer/format/csv/fastcsv/fastcsv.go
Matovidlo:
```
Using sync.Pool to reuse objects in I/O-bound tasks is much less efficient than re-using objects for CPU-bound tasks.
```
We actually use it in writer. So I would expect that this has more `I/O` operations than CPU ones.

internal/pkg/service/stream/storage/level/local/volume/disksync/disksync.go
```diff
@@ -16,50 +15,77 @@ import (
...
+	logger log.Logger,
+	clock clock.Clock,
+	config Config,
+	chain Chain,
+	statistics StatisticsProvider,
+) *Syncer
+
+// Chain is a resource that should be synchronized.
+type Chain interface {
 	// Flush data from memory to OS disk cache. Used if Config.Mode=ModeToCache.
```
Matovidlo, line 60: How this works? I dont understand how we are able to write into disk cache. I though it is abstraction and OS says whether it goes into cache or right into disk itself.

internal/pkg/service/stream/storage/level/local/volume/disksync/disksync.go
```diff
@@ -16,50 +15,77 @@ import (
...
 
-// Syncer writes data to the underlying writer and according to the Config it calls the chain.Flush() or chain.Sync().
-// Regarding waiting for sync, see Notifier and DoWithNotifier methods.
+// Syncer according to the Config calls the Chain.Flush() or Chain.Sync().
+// Regarding waiting for sync, see Notifier methods.
 type Syncer struct {
 	logger log.Logger
 	clock  clock.Clock
 	config Config
-	chain  chain
```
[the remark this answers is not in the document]
Matovidlo: I see `.svg` attached to this module. Shouldnt it be updated as well?

internal/pkg/service/stream/storage/level/local/writer/writer.go
Matovidlo: If we would like to replicate this on several disks, would it look like the writer is attached to multiple volumes/disks. So we would have `[]writer` or it would be `writer` with some replication configuration that does that?

internal/pkg/service/stream/storage/level/local/writer/format/csv/csv.go
```diff
@@ -0,0 +1,48 @@
...
+	"github.com/keboola/keboola-as-code/internal/pkg/utils/errors"
+)
+
+type Writer struct {
+	columns column.Columns
+	pool    *fastcsv.WritersPool
+}
+
+// NewWriter creates CSV writers pool and implements format.FormatWriter
+// The order of the lines is not preserved, because we use the writers pool,
```
Matovidlo, line 19: Isn't the problem that we do not preserve the order? However as it is streamed, it is expected that the messages could be delayed therefore the order does not matter. Am I right?

michaljurecko:
The updated diagram for writers chain, after the review, I will commit it to the repo.

image

internal/pkg/service/stream/storage/level/local/writer/format/csv/fastcsv/fastcsv.go
michaljurecko: I don't understand, there is no I/O in `fastcsv` Go package at all. 

It needs CPU and memory buffer per row (which is re-used) - so it's almost only about the CPU.

scripts/tools.sh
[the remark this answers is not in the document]
michaljurecko: To run benchmarks in CI, you need a dedicated machine on which nothing else is running, and which has stable parameters.

For E2E benchmarks, you need an empty Kubernetes cluster for each run.

In addition, you need some reporting, e.g. k6cloud, ideal if it sends the statistics to PR as a comment.

So yes, you should create tasks for it and implement it in cooperation with SRE.

internal/pkg/service/stream/storage/level/local/volume/disksync/disksync.go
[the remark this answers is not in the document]
michaljurecko: :heavy_check_mark: See comment at the end, I will add the diagram to repo after reviews.

internal/pkg/service/stream/storage/level/local/writer/format/csv/fastcsv/fastcsv.go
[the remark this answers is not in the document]
michaljurecko: Std CSV writer is not parallel, that's the main difference.

Here individual rows are prepared in parallel + and row buffer is reused.

internal/pkg/service/stream/storage/level/local/writer/writer.go
michaljurecko: - Written bytes should be replicated as some larger blocks.
- Write should wait until the data is saved on all disks.
- What we have now is the primary node, and a replica node would be created.
- The goal is to have the discs the same.


-----

Questions:

- What if the replica does not run, should the write pass or not?
- What if the data is written to the primary disk, but the write to the replica fails?
- ,,,


^^^ These are questions that are often solved in distributed systems/databases, so there will definitely be enough resources and examples on the Internet. For example I would see how you handle MongoDB replication, or etcd.

internal/pkg/service/stream/storage/level/local/volume/disksync/disksync.go
michaljurecko: Try to look at the resulting code, the diff is quite large.

internal/pkg/service/stream/storage/level/local/writer/format/csv/fastcsv/fastcsv.go
```diff
@@ -0,0 +1,54 @@
+// Package fastcsv provides parallel writer/formatter for CSV records.
+//   - There is WritersPool with fixed number of writers.
+//   - The io.Writer.Write method of the out writer is always called with the whole CSV row.
+//   - There is no built-in synchronization of the output, so the io.Writer.Write method must be protected externally.
```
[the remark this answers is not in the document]
michaljurecko: Writers in Go generally don't support parallel calls of `Write` method from different goroutines. It's up to the programmer to solve it, usually you don't need to write to the writer from several goroutines.

In our case, the `writerchain.Chain` handles it, each writer is wrapped in` writechain.safeWriter`, so the `Write` and `Flush` methods are thus protected by a lock.

internal/pkg/service/stream/storage/level/local/writer/format/csv/fastcsv/fastcsv.go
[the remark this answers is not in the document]
Matovidlo: Maybe I don't understand it correctly. I have looked into code of `encoding/csv` and the code is simillar except of 2 things:

1. We use `row` buffer, that we fill with row information, and insert `row` into `out`. Why is this more performant than writing into `out` itself? Ofc, we can reuse buffer but we need sequentially copy buffer into out. 
2. The only thing I found working more optimized way is that we do this in go routines where we have pool of workers, so instead of writing sequential row by row, we do multiple rows instead.

internal/pkg/service/stream/storage/level/local/writer/format/csv/fastcsv/fastcsv.go
Matovidlo: Let's have discussion about this.

scripts/tools.sh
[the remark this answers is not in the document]
michaljurecko: @hosekpeter please add comment above/or link to csv benchmark code.

internal/pkg/service/stream/storage/level/local/writer/format/csv/fastcsv/fastcsv.go
michaljurecko: @Matovidlo please add comments, rows order, etc.

PR #1843, opened 2024-06-14
fix: Two small fixes in background tasks

michaljurecko:

--------------

**Changes:**
- Task prepare phase is now connected to the caller context - fixed telemetry.
- The Stream API is using `StartTask`, instead of `RunTask`.
  - There was a bug, tasks were running in the foreground and were blocking the completion of the request.

-----------

-------------

### DIFF
--- a/internal/pkg/service/common/task/node.go
+++ b/internal/pkg/service/common/task/node.go
@@ -48,8 +48,7 @@ type Node struct {
 	logger log.Logger
 	client *etcd.Client
 
-	tasksCtx context.Context
-	tasksWg  *sync.WaitGroup
+	tasksWg *sync.WaitGroup
 
 	session *etcdop.Session
 
@@ -102,9 +101,7 @@ func NewNode(nodeID string, d dependencies, opts ...NodeOption) (*Node, error) {
 
 	// Graceful shutdown
 	bgContext := ctxattr.ContextWith(context.Background(), attribute.String("node", n.nodeID)) // nolint: contextcheck
-	var cancelTasks context.CancelFunc
 	n.tasksWg = &sync.WaitGroup{}
-	n.tasksCtx, cancelTasks = context.WithCancel(bgContext)
 	sessionWg := &sync.WaitGroup{}
 	sessionCtx, cancelSession := context.WithCancel(bgContext)
 	proc.OnShutdown(func(ctx context.Context) {
@@ -113,7 +110,6 @@ func NewNode(nodeID string, d dependencies, opts ...NodeOption) (*Node, error) {
 		if c := n.tasksCount.Load(); c > 0 {
 			n.logger.Infof(ctx, `waiting for "%d" tasks to be finished`, c)
 		}
-		cancelTasks()
 		n.tasksWg.Wait()
 		cancelSession()
 		sessionWg.Wait()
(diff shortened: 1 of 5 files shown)


internal/pkg/service/common/task/node.go
```diff
@@ -168,15 +164,15 @@ func (n *Node) StartTask(ctx context.Context, cfg Config) (t Task, err error) {
...
 }
 
 // RunTask in foreground, the task run at most once, it is provided by local lock and etcd transaction.
-func (n *Node) RunTask(cfg Config) (t Task, err error) {
+func (n *Node) RunTask(ctx context.Context, cfg Config) (t Task, err error) {
 	// Prepare task, acquire lock, handle error during prepare phase
-	task, fn, err := n.prepareTask(n.tasksCtx, cfg)
+	task, fn, err := n.prepareTask(ctx, cfg)
 	if err != nil {
 		return Task{}, err
```
michaljurecko, line 177: The prepare task phase is now using caller context, so it is now correctly reported by the telemetry.

michaljurecko: Task run phase has its own background context, see next commit.

internal/pkg/service/stream/api/service/task.go
```diff
@@ -43,13 +45,19 @@ func (s *service) startTask(cfg taskConfig) (task.Task, error) {
 	}
 
 	taskID := task.ID(cfg.Type + "/" + objectKey)
-	return s.tasks.RunTask(task.Config{
+	return s.tasks.StartTask(ctx, task.Config{
 		Type: cfg.Type,
 		Key:  task.Key{ProjectID: cfg.ProjectID, TaskID: taskID},
 		Lock: cfg.ObjectKey.String(),
```
michaljurecko, line 51: Fixed bug in API, `RunTask` -> `StartTask`.

internal/pkg/service/stream/api/service/task.go
```diff
@@ -43,13 +45,19 @@ func (s *service) startTask(cfg taskConfig) (task.Task, error) {
...
 		Context: func() (context.Context, context.CancelFunc) {
-			return context.WithTimeout(context.Background(), cfg.Timeout)
+			return context.WithTimeout(context.WithoutCancel(ctx), cfg.Timeout)
+		},
```
michaljurecko, line 54: Here is factory of the task background context, we can use relatively new `context.WithoutCancel` call, I will fix it also in the other places.

internal/pkg/service/stream/api/service/task.go
```diff
@@ -43,13 +45,19 @@ func (s *service) startTask(cfg taskConfig) (task.Task, error) {
...
+		Operation: func(ctx context.Context, logger log.Logger) task.Result {
+			rb := rollback.New(logger)
+			ctx = rollback.ContextWith(ctx, rb)
+			result := cfg.Operation(ctx, logger)
+			rb.InvokeIfErr(ctx, &result.Error)
+			return result
```
michaljurecko, line 60: Keboola table sink bridge operations need `rollback` instance in the context.

hosekpeter (approved):
LGTM

Matovidlo (approved):
LGTM

PR #1845, opened 2024-06-14
tests: Fix Stream API E2E tests, register an empty volume

michaljurecko:

--------------

**Changes:**
- An dummy disk writer component is started in API E2E test, so an empty volume is registered, and sink create/update API calls work.

-----------

-------------

### DIFF
--- a/internal/pkg/service/stream/components.go
+++ b/internal/pkg/service/stream/components.go
@@ -8,11 +8,13 @@ import (
 	"github.com/keboola/keboola-as-code/internal/pkg/service/stream/api"
 	"github.com/keboola/keboola-as-code/internal/pkg/service/stream/config"
 	"github.com/keboola/keboola-as-code/internal/pkg/service/stream/dependencies"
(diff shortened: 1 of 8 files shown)


internal/pkg/service/stream/components.go
```diff
@@ -21,6 +23,13 @@ type Component string
 
 func StartComponents(ctx context.Context, d dependencies.ServiceScope, cfg config.Config, components map[Component]bool) error {
 	// Start components, always in the same order
+
+	if components[ComponentDiskWriter] {
+		if err := writernode.Start(ctx, d, cfg); err != nil {
+			return err
+		}
+	}
+
```
michaljurecko, line 32: To create a sink, we need to create at least one file, with at least one slice.

To create slice, we need at least one active disk writer volume.

So in API E2E test, we have to start minimal disk writer component, which registers the volume.

Matovidlo (approved):
LGTM

hosekpeter (approved):

PR #1847, opened 2024-06-14
tests: Add Sink API E2E tests

michaljurecko:
Jira: https://keboola.atlassian.net/browse/PSGO-596

All tests will pass in the final [PR](https://github.com/keboola/keboola-as-code/pull/1770).

[Large set of fixes and E2e tests](https://github.com/keboola/keboola-as-code/pull/1759) is divided into smaller PRs to do a review.

--------------

**Changes:**
- Added Sink API E2E tests.

-----------

Commend per commits. Please, explore [the final version](https://github.com/keboola/keboola-as-code/pull/1759) in the IDE.

-------------

PR #1855, opened 2024-06-18
refactor: Rename parameter

jachym-tousek-keboola:
It was faster to just do it immediately. Only took 5 minutes.

https://github.com/keboola/keboola-as-code/pull/1847#discussion_r1642546824


Matovidlo (approved):
LGTM

PR #1869, opened 2024-06-25
feat: Change k6 to communicate with streamAPI instead of bufferAPI

Matovidlo:
Jira: https://keboola.atlassian.net/browse/PSGO-703

**Changes:**
- Updated K6 Benchmark to use Stream API

-----------

### DIFF
--- a/scripts/k6/stream-api/common.js
+++ b/scripts/k6/stream-api/common.js
@@ -37,60 +37,100 @@ export const options = {
   },
 };
 
-export function setupReceiver(exports) {
+export function setupSource() {
   if (!TOKEN) throw new Error("Please set the `API_TOKEN` env var.");
 
   const sourceId = "stream-" + randomString(8)
-  let res = post("v1/branches/main/sources", {
-    id: sourceId,
+  let res = post("v1/branches/default/sources", {
+    sourceId: sourceId,
     name: "Stream API Static Benchmark",
-    exports: exports,
+    type: "http",
   });
   if (res.status !== 202) {
     console.error(res);
-    throw new Error("failed to create receiver task");
+    throw new Error("failed to create source task");
   }
 
-  const createReceiverTimeoutSec = 60
+  const createSourceTimeoutSec = 60
   const taskUrl = stripUrlHost(res.json().url)
-  for (let retries = createReceiverTimeoutSec; retries > 0; retries--) {
+  for (let retries = createSourceTimeoutSec; retries > 0; retries--) {
     res = get(taskUrl)
     if (res.status !== 200) {
       console.error(res);
-      throw new Error("failed to get receiver task");
+      throw new Error("failed to get source task");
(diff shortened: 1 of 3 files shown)


scripts/k6/stream-api/static.js
```diff
@@ -9,13 +9,13 @@ export function setup() {
...
       mapping: {
-        tableId: "in.c-buffer-static.data",
+        tableId: "in.c-stream-static.data",
         columns: [
-          {type: "id", name: "id"},
-          {type: "datetime", name: "datetime"},
-          {type: "ip", name: "ip"},
-          {type: "body", name: "body"},
-          {type: "headers", name: "headers"},
+          { type: "id", name: "id" },
```
hosekpeter: I think, we should use uuid-v7 type, because have changed it in the migrate script.

https://github.com/keboola/keboola-as-code/blob/39848ff0a2107ac192517a787677ab05d61f0fb1/internal/pkg/service/stream/migrate/core/sink.go#L86

scripts/k6/stream-api/template.js
```diff
@@ -9,7 +9,7 @@ export function setup() {
       exportId: "test-export",
       name: "test-export",
       mapping: {
-        tableId: "in.c-buffer-template.data",
+        tableId: "in.c-stream-template.data",
         columns: [
           { type: "id", name: "id" },
           {
```
hosekpeter: the same

scripts/k6/stream-api/static.js
hosekpeter: ✔️

scripts/k6/stream-api/template.js
hosekpeter: ✔️

hosekpeter (approved):
LGTM

scripts/k6/stream-api/common.js
```diff
@@ -0,0 +1,170 @@
...
+
+  console.log("Sink id: " + sinkId)
+  return { id: sinkId }
+}
+
+export function stripUrlHost(url) {
+  return (new URL(url)).pathname
+}
+
+export function teardownSource(sourceId) {
```
Matovidlo, line 129: Missing teardown of Sink

scripts/k6/stream-api/static.js
```diff
@@ -0,0 +1,46 @@
...
+  const payload = { a: "b", c: { d: "e", f: { g: "h" } } };
+  const headers = {
+    "My-Custom-Header": "custom header value abcd",
+  };
+
+  return { source, sink, payload, headers };
+}
+
+export function teardown(data) {
+  common.teardownSource(data.source.id)
```
Matovidlo, line 40: Missing teardown of sink. I checked that it is possible to leave teardown of source it disallows the sink. Do we want to leave it like that?

PR #1872, opened 2024-06-28
fix: Rename Buffer -> Stream. Also in spans

Matovidlo:
**Changes:**
- Fix old Buffer API related names to Stream API

-----------


michaljurecko:
@Matovidlo can we merge it?

Matovidlo:
@michaljurecko Yes, I was just not sure whether it is everything but I suppose I went through all occurences of the `Buffer/BUFFER/buffer` occurences.

michaljurecko (approved):
LGTM

PR #1875, opened 2024-07-01
feat: Difference slice statistics by the source node ID

michaljurecko:

**Changes:**
- Statistics are collected by `<sliceID>/<sourceNodeID>`, not by `<sliceID>`.

-----------

### HTTP Source PRs Intro

We are now solving communication between a source node (HTTP source) and a writer node.
- Reason: The HTTP source could be scaled independently of the storage.
- Originally, these parts were connected directly, it was one pod.
- So we can have `M` source nodes and `N` writer nodes.

image

---------------------

#### Architectural decisions

- **Compression** is the responsibility of the source node, not the writer node.
  - The source node typically performs CPU intensive tasks (generation of CSV lines).
  - Compression is also CPU intensive task, so it makes sense to group these jobs, for scaling purposes.
  - The write node is limited by disk I/O, if we had to scale it for the CPU, we would have to unnecessarily scale the number of disks (stateful set).
  - It reduces the size of bytes transferred over the local network.
- **Sink statistics** are collected by the source node.
  - Writer node receives compressed data, so there would be a problem with 1. count and 2. uncompressed size metrics.
    - They have to be sent via a network side channel, so we can rather store them directly in the DB.
- So in the end there are 2 changes:
  - In the source node/writers chain, we don't write to a physical OS file, but to a network file abstraction.
  - There is `M:N` - `source nodes:writer nodes` situation:
    - We need to modify statistics `Collector` (this PR).
    - Source node is stateless, so statistcs cannot be backed up to a file - but only directly to the DB (next PR).
    - We need a routing algorithm `source -> writer` node with retries.

##### WIP diagram

image

### DIFF
--- a/internal/pkg/service/common/utctime/utctime.go
+++ b/internal/pkg/service/common/utctime/utctime.go
@@ -40,6 +40,9 @@ func (v UTCTime) After(target UTCTime) bool {
 }
 
 func (v UTCTime) MarshalJSON() ([]byte, error) {
+	if v.IsZero() {
+		return jsonLib.Marshal("")
+	}
 	return jsonLib.Marshal(v.String())
 }
 
@@ -47,6 +50,10 @@ func (v *UTCTime) UnmarshalJSON(b []byte) error {
 	var str string
 	if err := jsonLib.Unmarshal(b, &str); err != nil {
 		return err
+	}
+	if str == "" {
+		*v = UTCTime{}
+		return nil
 	}
 	out, err := time.Parse(TimeFormat, str)
 	if err != nil {
--- a/internal/pkg/service/stream/storage/node/writernode/writernode.go
+++ b/internal/pkg/service/stream/storage/node/writernode/writernode.go
@@ -46,7 +46,7 @@ func Start(ctx context.Context, d dependencies, cfg config.Config) error {
 
 	// Setup statistics collector
 	syncCfg := cfg.Storage.Statistics.Collector
-	collector.Start(d, volumes.Events(), syncCfg)
+	collector.Start(d, volumes.Events(), syncCfg, cfg.NodeID)
 
 	return nil
 }
--- a/internal/pkg/service/stream/storage/repository/slice/fixtures/slice_open_ops_001.txt
+++ b/internal/pkg/service/stream/storage/repository/slice/fixtures/slice_open_ops_001.txt
@@ -10,18 +10,16 @@
   001 ➡️  GET ["definition/sink/active/123/456/my-source/", "definition/sink/active/123/456/my-source0")
 ✔️  TXN | succeeded: true
 
-// READ - sinks from the source
+// READ - get previous slices in the sink and their statistics to calculate pre-allocated disk space
 ➡️  TXN
   ➡️  THEN:
-  001 ➡️  GET ["storage/stats/staging/123/456/my-source/my-sink/", "storage/stats/staging/123/456/my-source/my-sink0")
-  002 ➡️  GET ["storage/stats/target/123/456/my-source/my-sink/", "storage/stats/target/123/456/my-source/my-sink0")
+  001 ➡️  GET ["storage/slice/level/staging/123/456/my-source/my-sink/", "storage/slice/level/staging/123/456/my-source/my-sink0")
+  002 ➡️  GET ["storage/slice/level/target/123/456/my-source/my-sink/", "storage/slice/level/target/123/456/my-source/my-sink0")
 ✔️  TXN | succeeded: true
-
-// READ - get statistics of the sink - to calculate pre-allocated disk space
 ➡️  TXN
   ➡️  THEN:
-  001 ➡️  GET ["storage/stats/staging/123/456/my-source/my-sink/", "storage/stats/staging/123/456/my-source/my-sink0")
-  002 ➡️  GET ["storage/stats/target/123/456/my-source/my-sink/", "storage/stats/target/123/456/my-source/my-sink0")
+  001 ➡️  GET ["storage/slice/level/staging/123/456/my-source/my-sink/", "storage/slice/level/staging/123/456/my-source/my-sink0")
+  002 ➡️  GET ["storage/slice/level/target/123/456/my-source/my-sink/", "storage/slice/level/target/123/456/my-source/my-sink0")
 ✔️  TXN | succeeded: true
 
 // WRITE
--- a/internal/pkg/service/stream/storage/statistics/collector/collector.go
+++ b/internal/pkg/service/stream/storage/statistics/collector/collector.go
@@ -23,6 +23,7 @@ type Collector struct {
 	logger     log.Logger
 	repository *repository.Repository
 	config     statistics.SyncConfig
+	nodeID     string
 	wg         *sync.WaitGroup
 
 	syncLock    *sync.Mutex
@@ -49,11 +50,12 @@ type dependencies interface {
 	StatisticsRepository() *repository.Repository
 }
 
-func Start(d dependencies, events WriterEvents, config statistics.SyncConfig) {
+func Start(d dependencies, events WriterEvents, config statistics.SyncConfig, nodeID string) {
 	c := &Collector{
 		logger:      d.Logger().WithComponent("statistics.collector"),
 		repository:  d.StatisticsRepository(),
 		config:      config,
+		nodeID:      nodeID,
 		wg:          &sync.WaitGroup{},
 		syncLock:    &sync.Mutex{},
 		writersLock: &sync.Mutex{},
@@ -161,7 +163,7 @@ func (c *Collector) sync(filter *model.SliceKey) error {
 
 	// Update values in the database
 	if len(forSync) > 0 {
-		if err := c.repository.Put(ctx, forSync); err != nil {
+		if err := c.repository.Put(ctx, c.nodeID, forSync); err != nil {
 			err = errors.Errorf("cannot save the storage statistics to the database: %w", err)
 			c.logger.Error(ctx, err.Error())
 			return err
--- a/internal/pkg/service/stream/storage/statistics/collector/collector.go
+++ b/internal/pkg/service/stream/storage/statistics/collector/collector.go
@@ -39,8 +39,10 @@ type WriterEvents interface {
 // writerSnapshot contains collected statistics from a writer.Writer.
 // It is used to determine whether the statistics have changed and should be saved to the database or not.
 type writerSnapshot struct {
-	stats  statistics.PerSlice
-	writer writer.Writer
+	writer       writer.Writer
+	sliceKey     model.SliceKey
+	initialValue statistics.Value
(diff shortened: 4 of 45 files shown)


michaljurecko:
(it's a dead-end road)

michaljurecko:
(in the end it will probably be the best solution)

internal/pkg/service/common/utctime/utctime.go
```diff
@@ -40,6 +40,9 @@ func (v UTCTime) After(target UTCTime) bool {
 }
 
 func (v UTCTime) MarshalJSON() ([]byte, error) {
+	if v.IsZero() {
+		return jsonLib.Marshal("")
+	}
 	return jsonLib.Marshal(v.String())
```
michaljurecko, line 46: Empty UTCTime is marshaled as empty string - the change is BC compatible.

internal/pkg/service/stream/storage/node/writernode/writernode.go
```diff
@@ -46,7 +46,7 @@ func Start(ctx context.Context, d dependencies, cfg config.Config) error {
 
 	// Setup statistics collector
 	syncCfg := cfg.Storage.Statistics.Collector
-	collector.Start(d, volumes.Events(), syncCfg)
+	collector.Start(d, volumes.Events(), syncCfg, cfg.NodeID)
 
 	return nil
```
michaljurecko, line 51: Statistics Collector knows the node ID.

internal/pkg/service/stream/storage/statistics/collector/collector.go
```diff
@@ -161,7 +163,7 @@ func (c *Collector) sync(filter *model.SliceKey) error {
 
 	// Update values in the database
 	if len(forSync) > 0 {
-		if err := c.repository.Put(ctx, forSync); err != nil {
+		if err := c.repository.Put(ctx, c.nodeID, forSync); err != nil {
 			err = errors.Errorf("cannot save the storage statistics to the database: %w", err)
```
michaljurecko, line 187: The node ID is passed to the statistcs repository `Put` method.

internal/pkg/service/stream/storage/statistics/repository/stats_move.go
```diff
@@ -41,27 +41,12 @@ func (r *Repository) moveAll(parentKey fmt.Stringer, from, to model.Level, trans
...
-				}
-			})
-		})
-	} else {
-		// List
-		atomicOp.Read(func(ctx context.Context) op.Op {
-			return r.schema.InLevel(from).InObject(parentKey).GetAll(r.client).WithAllKVsTo(&valueKVs)
-		})
-	}
-
```
michaljurecko, line 63: We could optimize this even without other changes.

internal/pkg/service/stream/storage/statistics/repository/stats_put.go
```diff
@@ -31,7 +31,7 @@ func (r *Repository) Put(ctx context.Context, stats []statistics.PerSlice) (err
 			i = 0
 			addTxn()
 		}
-		currentTxn.Then(r.schema.InLevel(model.LevelLocal).InSlice(v.SliceKey).Put(r.client, v.Value))
+		currentTxn.Then(r.schema.InLevel(model.LevelLocal).InSourceNodeSlice(v.SliceKey, nodeID).Put(r.client, v.Value))
 		i++
```
michaljurecko: Statistcs are collected/stored par slice and node ID.

internal/pkg/service/stream/storage/statistics/repository/stats_open.go
```diff
@@ -0,0 +1,30 @@
...
+// so the Collector can continue where it left off.
+// A non-empty value is returned only if the source node was already writing to the slice, but there was a crash/restart.
+func (r *Repository) OpenSlice(k model.SliceKey, nodeID string) *op.TxnOp[statistics.Value] {
+	openKey := r.schema.InLevel(model.LevelLocal).InSlice(k).Key(sliceOpenKey)
+	nodeKey := r.schema.InLevel(model.LevelLocal).InSliceSourceNode(k, nodeID)
+	var statsValue statistics.Value
+	return op.TxnWithResult(r.client, &statsValue).
+		ThenTxn(openKey.PutIfNotExists(r.client, statistics.Value{SlicesCount: 1})).
+		Then(nodeKey.GetOrEmpty(r.client).WithResultTo(&statsValue))
+}
```
michaljurecko, line 30: There are 2 new problems there.

----------------------------

1. The `SlicesCount` field.
- If every source node stores statistics with `SlicesCount=1`, the final sum would not match.
- So I added a special key `<sliceID>/_open`, where the only `SlicesCount=1` is stored, by the first source node/writer.
- The statistics mechanism thus works correctly without further changes.

-----------------------------------------

2. The source node is stateless, there cannot be a backup to a local file.
- So the Collector needs to know initial statistics state, after a rare crash/restart of the pod.
- So the `OpenSlice` method returns initial statistics value, if any.

internal/pkg/service/stream/storage/statistics/repository/fixtures/stats_open_snapshot_001.txt
```diff
@@ -0,0 +1,9 @@
+<<<<<
+storage/stats/local/123/456/my-source/my-sink/2000-01-01T19:00:00.000Z/my-volume/2000-01-01T20:00:00.000Z/_open
+-----
+{
+  "slicesCount": 1,
+  "firstRecordAt": "",
+  "lastRecordAt": ""
+}
```
michaljurecko, line 8: The UTCTime/time.Time is a struct, so the `omitempty` cannot be used.

It could be solved, but I'm not going to waste time on it.

internal/pkg/service/stream/storage/statistics/statistics.go
```diff
@@ -40,7 +40,14 @@ type Value struct {
...
 	SliceKey model.SliceKey
-	Value    Value
+
+	// The following fields are part of the Value structure, which are applicable when collecting statistics.
+
+	FirstRecordAt    utctime.UTCTime
+	LastRecordAt     utctime.UTCTime
+	RecordsCount     uint64
+	UncompressedSize datasize.ByteSize
+	CompressedSize   datasize.ByteSize
```
michaljurecko, line 50: To prevent misuse, only relevant fields are here, no `SlicesCount` and `StagingSize`.

internal/pkg/service/stream/storage/statistics/statistics.go
```diff
@@ -21,7 +21,7 @@ import (
 
 // Value contains statistics for a slice or summarized statistics for a parent object.
 type Value struct {
-	SlicesCount int `json:"slicesCount,omitempty" validate:"required_with=RecordsCount"`
+	SlicesCount int `json:"slicesCount,omitempty"`
```
michaljurecko, line 24: Slices count reported by the source nodes is `0`, it is `1` in the `_open` key.

internal/pkg/service/stream/storage/statistics/collector/collector.go
```diff
@@ -39,8 +39,10 @@ type WriterEvents interface {
 // writerSnapshot contains collected statistics from a writer.Writer.
 // It is used to determine whether the statistics have changed and should be saved to the database or not.
 type writerSnapshot struct {
-	stats  statistics.PerSlice
-	writer writer.Writer
+	writer       writer.Writer
+	sliceKey     model.SliceKey
+	initialValue statistics.Value
+	value        statistics.Value
 }
```
michaljurecko, line 46: collector.writerSnapshot: added `initialValue`.

internal/pkg/service/stream/storage/statistics/repository/stats_move.go
```diff
@@ -64,10 +64,8 @@ func (r *Repository) moveAll(parentKey fmt.Stringer, from, to model.Level, trans
 				oldKey := etcdop.NewTypedKey[statistics.Value](kv.Key(), r.schema.Serde())
 
 				// Save value to the new destination
-				if kv.Value.RecordsCount > 0 {
-					newKey := oldKey.ReplacePrefix(r.schema.InLevel(from).Prefix(), r.schema.InLevel(to).Prefix())
-					txn.Then(newKey.Put(r.client, kv.Value))
-				}
+				newKey := oldKey.ReplacePrefix(r.schema.InLevel(from).Prefix(), r.schema.InLevel(to).Prefix())
+				txn.Then(newKey.Put(r.client, kv.Value))
```
michaljurecko, line 68: We need to include also the `_open` key, `{SlicesCount:1, RecordsCount:0}`.

internal/pkg/service/stream/storage/statistics/repository/stats_allocation.go
```diff
@@ -27,41 +28,62 @@ func (r *Repository) estimateSliceSizeOnSliceOpen() {
...
+	err := op.Txn(r.client).
+		Then(r.storage.Slice().ListInLevel(parentKey, model.LevelStaging, listSlicesOpts...).WithAllTo(&lastStagingSlices)).
+		Then(r.storage.Slice().ListInLevel(parentKey, model.LevelTarget, listSlicesOpts...).WithAllTo(&lastTargetSlices)).
+		Do(ctx).
+		Err()
+	if err != nil {
+		return 0, err
+	}
+
+	// Load and process statistics for each slice from the previous step
```
michaljurecko, line 67: The main issue was, wit the `estimateSliceSize/maxUsedDiskSizeBySliceIn`.
- Individual records in the etcd prefix are no more full slice statistics, but only partial statistics from a source node.
- So I cannot get easily the last `10` slice statistics.
- I must load the last `10` slices from the storage repository, and then sum statistics for there slices.

internal/pkg/service/stream/storage/statistics/repository/stats_get.go
```diff
@@ -80,6 +80,20 @@ func (r *Repository) AggregateIn(objectKey fmt.Stringer) *op.TxnOp[statistics.Ag
 	return txn
 }
 
+func (r *Repository) AggregateInLevel(objectKey fmt.Stringer, level model.Level) *op.TxnOp[statistics.Aggregated] {
```
michaljurecko, line 83: We loaded the slice entity, so we know the slice state/level.

internal/pkg/service/stream/storage/repository/slice/fixtures/slice_open_ops_001.txt
```diff
@@ -10,18 +10,16 @@
   001 ➡️  GET ["definition/sink/active/123/456/my-source/", "definition/sink/active/123/456/my-source0")
 ✔️  TXN | succeeded: true
 
-// READ - sinks from the source
+// READ - get previous slices in the sink and their statistics to calculate pre-allocated disk space
```
michaljurecko, line 13: Here was typo in the comment.

hosekpeter (approved):

PR #1892, opened 2024-07-08
feat: Add flag `header` for download table command

hosekpeter:
Jira: [PSGO-295](https://keboola.atlassian.net/browse/PSGO-295)

**Changes:**
- added new flag `header`
- added more tests for this command

-----------
Added a new field `header` for download table command. This flag adds column names to csv file result.

[PSGO-295]: https://keboola.atlassian.net/browse/PSGO-295?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


internal/pkg/service/cli/cmd/remote/table/download/cmd.go
```diff
@@ -125,3 +134,16 @@ func Command(p dependencies.Provider) *cobra.Command {
...
+	if f.Header.IsSet() && len(f.Columns.Value) == 0 {
+		table, err := d.KeboolaProjectAPI().GetTableRequest(key).Send(ctx)
+		if err != nil {
+			return nil, err
+		}
+
+		return table.Columns, nil
+	}
+
+	return f.Columns.Value, nil
```
Matovidlo, line 148:
```suggestion
	if !f.Header.IsSet() || len(f.Columns.Value) != 0 {
		return f.Columns.Value, nil
	}
	
	table, err := d.KeboolaProjectAPI().GetTableRequest(key).Send(ctx)
	if err != nil {
		return nil, err
	}
	
	return table.Columns, nil
```

Matovidlo: This way it is less nested and we know that when we do not have flag `headers` set, it returns directly.

Matovidlo (approved):
LGTM

internal/pkg/service/cli/cmd/remote/table/download/cmd.go
hosekpeter: fixed [d827bf2](https://github.com/keboola/keboola-as-code/pull/1892/commits/d827bf2d077fd714171b52cb659415d1f123fabc)

Matovidlo (approved):
LGTM

PR #1894, opened 2024-07-09
feat: Add disk writer network transport layer (kcp-go/yamux)

michaljurecko:

**Changes:**
- Added network transport layer between source and disk writer nodes.
- Based on Reliable UDP Protocol - KCP, and yamux streams multiplexer.

-----------

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -32,6 +32,7 @@ require (
 	github.com/google/go-cmp v0.6.0
 	github.com/google/go-jsonnet v0.20.0
 	github.com/google/shlex v0.0.0-20191202100458-e7afc7fbc510
+	github.com/hashicorp/yamux v0.1.1
 	github.com/jarcoal/httpmock v1.3.1
 	github.com/joho/godotenv v1.5.1
 	github.com/jpillora/longestcommon v0.0.0-20161227235612-adb9d91ee629
@@ -65,6 +66,7 @@ require (
 	github.com/urfave/negroni v1.0.0
 	github.com/valyala/fasthttp v1.55.0
 	github.com/writeas/go-strip-markdown v2.0.1+incompatible
+	github.com/xtaci/kcp-go/v5 v5.6.8
 	go.etcd.io/etcd/api/v3 v3.5.14
 	go.etcd.io/etcd/client/v3 v3.5.14
 	go.etcd.io/etcd/tests/v3 v3.5.14
@@ -229,6 +231,8 @@ require (
 	github.com/jonboulle/clockwork v0.2.2 // indirect
 	github.com/json-iterator/go v1.1.12 // indirect
 	github.com/justinas/alice v1.2.0 // indirect
+	github.com/klauspost/cpuid/v2 v2.2.6 // indirect
+	github.com/klauspost/reedsolomon v1.12.0 // indirect
 	github.com/kyokomi/emoji/v2 v2.2.10 // indirect
 	github.com/lucasb-eyer/go-colorful v1.2.0 // indirect
 	github.com/mattn/go-runewidth v0.0.15 // indirect
@@ -255,6 +259,9 @@ require (
 	github.com/secure-systems-lab/go-securesystemslib v0.7.0 // indirect
 	github.com/soheilhy/cmux v0.1.5 // indirect
 	github.com/sourcegraph/conc v0.3.0 // indirect
+	github.com/templexxx/cpu v0.1.0 // indirect
+	github.com/templexxx/xorsimd v0.4.2 // indirect
+	github.com/tjfoc/gmsm v1.4.1 // indirect
 	github.com/tmc/grpc-websocket-proxy v0.0.0-20201229170055-e5319fda7802 // indirect
 	github.com/valyala/bytebufferpool v1.0.0 // indirect
 	github.com/vmihailenco/msgpack/v5 v5.4.1 // indirect
--- a/internal/pkg/service/stream/storage/level/local/diskwriter/network/transport/transport.go
+++ b/internal/pkg/service/stream/storage/level/local/diskwriter/network/transport/transport.go
@@ -1,0 +1,37 @@
+// Package transport provides transport layer for communication between source and disk writer nodes.
+//
+// Communication is based on KCP Reliable UDP protocol and yamux streams multiplexer.
+// TCP is an old protocol with several issues regarding latency.
+// It is not possible to make major changes to TCP, as it is a widely used standard.
+// Therefore, R-UDP protocols are used to address issues.
+package transport
+
+import (
+	"fmt"
+
+	"github.com/hashicorp/yamux"
+
+	"github.com/keboola/keboola-as-code/internal/pkg/log"
+	"github.com/keboola/keboola-as-code/internal/pkg/service/stream/storage/level/local/diskwriter/network"
+)
+
+func sessionKey(session *yamux.Session) string {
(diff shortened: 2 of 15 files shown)


internal/pkg/service/stream/storage/level/local/diskwriter/network/transport/transport_test.go
```diff
@@ -0,0 +1,108 @@
...
+	"github.com/keboola/keboola-as-code/internal/pkg/service/stream/storage/level/local/diskwriter/network"
+	"github.com/keboola/keboola-as-code/internal/pkg/service/stream/storage/level/local/diskwriter/network/transport"
+	"github.com/keboola/keboola-as-code/internal/pkg/utils/errors"
+)
+
+func TestServerAndClient(t *testing.T) {
+	t.Parallel()
+
+	ctx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
+	defer cancel()
```
michaljurecko, line 26: Implementing a graceful shutdown and a stable test was quite difficult.

In the future, it would be good to add more tests, using `toxiproxy` and test e.g. keep alive mechanism.

We are already using the `toxiproxy` for some critical etcd tests, to simulate network outage.
https://github.com/keboola/keboola-as-code/blob/484f05099bf157c0cf92330da5c21c5d547b382e/internal/pkg/service/common/etcdop/session_test.go#L10

internal/pkg/service/stream/storage/level/local/diskwriter/network/transport/transport_test.go
```diff
@@ -0,0 +1,108 @@
...
+{"level":"info","message":"exited"}
+`)
+
+	// Check server logs
+	srvMock.DebugLogger().AssertJSONMessages(t, `
+{"level":"info","message":"disk writer listening on \"127.0.0.1:%d\"","component":"storage.node.writer.network.server"}
+{"level":"info","message":"accepted connection from \"127.0.0.1:%d\"","component":"storage.node.writer.network.server"}
+{"level":"info","message":"exiting (bye bye)"}
+{"level":"info","message":"closing disk writer server","component":"storage.node.writer.network.server"}
+{"level":"info","message":"waiting 5s for %d streams","component":"storage.node.writer.network.server"}
```
michaljurecko, line 101: If we close the connection on the client, it takes some time for the server to react.

Therefore, it is not defined whether the stream/connection is completed before the shutdown or during the shutdown, the important thing is that no data is lost.

internal/pkg/service/stream/storage/level/local/diskwriter/network/transport/transport.go
```diff
@@ -0,0 +1,37 @@
+// Package transport provides transport layer for communication between source and disk writer nodes.
+//
+// Communication is based on KCP Reliable UDP protocol and yamux streams multiplexer.
+// TCP is an old protocol with several issues regarding latency.
+// It is not possible to make major changes to TCP, as it is a widely used standard.
+// Therefore, R-UDP protocols are used to address issues.
+package transport
+
```
michaljurecko, line 8: Is it clear, or do you have any questions?

michaljurecko: `kcp-go` is used also by the popular https://github.com/syncthing/syncthing

hosekpeter (approved):
LGTM

go.mod
```diff
@@ -32,6 +32,7 @@ require (
 	github.com/google/go-cmp v0.6.0
 	github.com/google/go-jsonnet v0.20.0
 	github.com/google/shlex v0.0.0-20191202100458-e7afc7fbc510
```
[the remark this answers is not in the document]
michaljurecko: @Matovidlo (cc @jachym-tousek-keboola ) you are mixing together different network stack layers :apple: :pear: 

`udp/tcp/kcp` - are low-level protocols to transfer `bytes`

`rpc/grcp/http` - are serialization/API layers

You can run `grpc` over `kcp`, example: https://github.com/Lynnworld/grpc-kcp-transport

I selected `net/rpc` + `gob` (it will be a next PR) as serialization/API layer.

https://github.com/cockroachdb/rpc-bench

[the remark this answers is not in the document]
Matovidlo: 1. I understand that they are `L4` layer. However do we really need such a complication? 
2. You did not respond on streaming part, yet we can optimize it that much but do we need that ? We can have the best implementation vs good enough to fit the project needs.
3. Benchmark is 4 years old, so not relevant now.
4. I don't like the choice as gRPC has over 20k stars and is very relevant these days for internal communication.
I understand that it is not worth to analyse/rewrite it now. I just wanted to point out that this is quite a decision and I am not happy about that.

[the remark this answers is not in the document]
michaljurecko: > 1. I understand that they are L4 layer. However do we really need such a complication?

:one: TCP/KCP
- Tunning of TCP protocol may be complicated, and it must be done on the OS level, [read more](https://blog.cloudflare.com/optimizing-tcp-for-high-throughput-and-low-latency).
- This also applies to the UDP protocol, but there are much fewer settings at the OS level.
- KCP can be well tunned in app level.
- On my `localhost`, TCP is 3x-4x faster than my disk, so it can actually be a bottleneck.
  - Will we run into it in a week, a month, never? I don't know.
  - Well, I'm sure you'll be happy to have already something to experiment with there.
  - Replacing these low-level things later, if you don't think about it id advance, can be quite a problem and time-consuming.
- In the [next PR](https://github.com/keboola/keboola-as-code/pull/1895), I set TCP as default protocol for initial testing.

:two: yamux - connections multiplexing
- Establishing a connection has a a overhead and delay.
- A separate connection for each slice would consume more resources.
- You could simply run into a connections count limit.

> 2. You did not respond on streaming part, yet we can optimize it that much but do we need that ? We can have the best implementation vs good enough to fit the project needs.

I would like to answer, but I don't know how and what exactly you mean. Our entire solution is based on streaming, the protocol has no effect on it.

gRPC supports streaming of entities, it is great, but we are sending chunks of bytes, it is possible in each protocol.

> 3. Benchmark is 4 years old, so not relevant now.

The link to the benchmark was there to show that they are similar solutions, also looked at various other benchmarks, depending on the use-case one or the other will be faster, but there is no significant difference.

> 4. I don't like the choice as gRPC has over 20k stars and is very relevant these days for internal communication.
I understand that it is not worth to analyse/rewrite it now. I just wanted to point out that this is quite a decision and I am not happy about that.

`net/rpc`  +`encoding/gob` are parts of the standard library, but yes, gRPC is great (for API/Server2Server comm.).

I chosed `net/rpc` because gRPC seemed overkill for our simple use-case.

internal/pkg/service/stream/storage/level/local/diskwriter/network/transport/transport.go
michaljurecko: In the [next PR](https://github.com/keboola/keboola-as-code/pull/1895) I slight improved code readability.

- [TCP version](https://github.com/keboola/keboola-as-code/blob/michaljurecko-PSGO-591-network-transport-2/internal/pkg/service/stream/storage/level/local/diskwriter/network/transport/transport_tcp.go)
- [KCP version](https://github.com/keboola/keboola-as-code/blob/michaljurecko-PSGO-591-network-transport-2/internal/pkg/service/stream/storage/level/local/diskwriter/network/transport/transport_kcp.go)

PR #1895, opened 2024-07-09
feat: Network transport (2) - TCP/KCP protocol can be switched by cfg

michaljurecko:

See https://github.com/keboola/keboola-as-code/pull/1894

**Changes:**
- `TCP/KCP` protocol can be switched by configuration.
  - Default value is `TCP`.
  - It will be good enough for start and save time.
  - [UPD Buffer Size](https://github.com/quic-go/quic-go/wiki/UDP-Buffer-Sizes) is necessary to achieve high performance, but it is not easy to set up in Kubernetes (I don't know which defaults value are in our stacks).
- Improved tests.
- Related to:
  - https://github.com/keboola/keboola-as-code/pull/1894#discussion_r1670096248
  - https://github.com/keboola/keboola-as-code/pull/1894#discussion_r1669818027
-----------

KCP is on `localhost` 2-3x slower than TCP, but it is expected.
- TCP runs in kernel space and has various optimalization for localhost.
- KCP runs in user space and its benefits can be seen on real networks (see projects like: [kcptun](https://github.com/xtaci/kcptun), [frp](https://github.com/fatedier/frp), [syncthing](https://github.com/syncthing/syncthing), ...)

`dataSize = 100MB` - both protocols are fast enough up to `500MB/s` on localhost.

image

Read more (`quic` is another R-UDP protocol, the principle is the same):
- https://github.com/quic-go/quic-go/issues/2586#issuecomment-642312892
- https://github.com/quic-go/quic-go/issues/2586#issuecomment-876488346

### DIFF
--- a/.github/workflows/test-lint.yml
+++ b/.github/workflows/test-lint.yml
@@ -16,6 +16,16 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v4
 
+      - name: Free Disk Space (Ubuntu)
+        uses: jlumbroso/free-disk-space@main
+        with:
+          android: true
+          dotnet: true
+          haskell: true
+          large-packages: false
+          docker-images: false
+          swap-storage: false
+
       - name: Setup Go, tools and caching
         uses: ./.github/actions/go-setup
         with:
--- a/internal/pkg/service/stream/storage/level/local/diskwriter/network/transport/transport_kcp.go
+++ b/internal/pkg/service/stream/storage/level/local/diskwriter/network/transport/transport_kcp.go
@@ -1,34 +1,33 @@
 package transport
 
 import (
+	"net"
+
+	"github.com/xtaci/kcp-go/v5"
+
 	"github.com/keboola/keboola-as-code/internal/pkg/service/stream/storage/level/local/diskwriter/network"
 	"github.com/keboola/keboola-as-code/internal/pkg/utils/errors"
-	"github.com/xtaci/kcp-go/v5"
-	"net"
 )
 
-type Transport interface {
-	// Listen for connections by the server.
-	Listen(config network.Config) (net.Listener, error)
-	// Accept new connection by the server.
-	Accept(listener net.Listener) (net.Conn, error)
-	// Dial a new connection from the client.
-	Dial(addr string, config network.Config) (net.Conn, error)
+type kcpTransport struct {
+	config network.Config
 }
 
-type kcpTransport struct{}
+func newKCPTransport(config network.Config) Transport {
+	return &kcpTransport{config: config}
+}
 
-func (*kcpTransport) Listen(config network.Config) (net.Listener, error) {
-	listener, err := kcp.ListenWithOptions(config.Listen, nil, 0, 0)
+func (t *kcpTransport) Listen() (net.Listener, error) {
+	listener, err := kcp.ListenWithOptions(t.config.Listen, nil, 0, 0)
 	if err != nil {
 		return nil, errors.PrefixError(err, "cannot create listener")
 	}
 
 	// Setup buffer sizes (reversed as on the client side)
-	if err := listener.SetReadBuffer(int(config.InputBuffer.Bytes())); err != nil {
+	if err := listener.SetReadBuffer(int(t.config.KCPInputBuffer.Bytes())); err != nil {
 		return nil, errors.PrefixError(err, "cannot set read buffer size")
 	}
-	if err := listener.SetWriteBuffer(int(config.ResponseBuffer.Bytes())); err != nil {
+	if err := listener.SetWriteBuffer(int(t.config.KCPResponseBuffer.Bytes())); err != nil {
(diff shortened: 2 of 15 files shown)


.github/workflows/test-lint.yml
```diff
@@ -16,6 +16,16 @@ jobs:
...
+        uses: jlumbroso/free-disk-space@main
+        with:
+          android: false
+          dotnet: true
+          haskell: false
+          large-packages: false
+          docker-images: false
+          swap-storage: false
+
       - name: Setup Go, tools and caching
```
michaljurecko, line 29: image

internal/pkg/service/stream/storage/level/local/diskwriter/network/transport/transport_kcp.go
```diff
@@ -0,0 +1,72 @@
...
+
+	t.setupConnection(conn)
+	return conn, nil
+}
+
+func (t *kcpTransport) setupConnection(conn net.Conn) {
+	c := conn.(*kcp.UDPSession)
+	c.SetWindowSize(512, 512)
+	c.SetWriteDelay(false)
+	c.SetStreamMode(false)
```
Matovidlo, line 69: Why did we change stream mode from `true` to `false`?

internal/pkg/service/stream/storage/level/local/diskwriter/network/transport/util.go
```diff
@@ -0,0 +1,33 @@
...
+func multiplexerConfig(_ log.Logger, config network.Config) *yamux.Config {
+	return &yamux.Config{
+		AcceptBacklog:          config.MaxWaitingStreams,
+		EnableKeepAlive:        true,
+		KeepAliveInterval:      config.KeepAliveInterval,
+		ConnectionWriteTimeout: config.StreamWriteTimeout,
+		MaxStreamWindowSize:    uint32(config.StreamMaxWindow.Bytes()),
+		StreamOpenTimeout:      config.StreamOpenTimeout,
+		StreamCloseTimeout:     config.StreamCloseTimeout,
+		// Logger: log.NewStdErrorLogger(logger.WithComponent("mux")),
```
Matovidlo, line 30: Commented logger.

michaljurecko: Yes. It logs errors duplicated to those we log ourselves. I will add a comment.

internal/pkg/service/stream/storage/level/local/diskwriter/network/transport/transport_kcp.go
michaljurecko: It is ~ 2-3% faster, it does not affect the functionality. It is an internal setting that is not reflected in the public interface.

internal/pkg/service/stream/storage/level/local/diskwriter/network/transport/transport_kcp.go
```diff
@@ -0,0 +1,72 @@
...
+	c.SetACKNoDelay(true) // send data immediately, needed for bigger payloads
+	c.SetNoDelay(1, 10, 2, 1)
```
michaljurecko, line 71: For context, why?

During testing of upper layers, I found that writing `100MB` takes `5s` to write. 

It didn't fit in any way with the previous experiments that I did when I was choosing the protocol.

That's why I added the TCP protocol there, for comparison.

Finally, after a few hours, I realized that I forgot to set `SetACKNoDelay(true)', which makes the difference.

Otherwise, it waits for the next sending window, triggered by a timer.

It makes sense for small messages that are smaller than a UDP packet (~1KB), but for our 10KB+ records it slows down the record (we have buffers in the pipeline).

internal/pkg/service/stream/storage/level/local/diskwriter/network/transport/util.go
michaljurecko: https://github.com/keboola/keboola-as-code/pull/1895/commits/582ec368d7cedd05846b1bcfe4368f8b1d9106ce

Matovidlo (approved):
LGTM

.github/workflows/test-lint.yml
michaljurecko: (maybe it needs some better fix, It looks like, the linter cache is growing and growing)

Matovidlo: @michaljurecko  We are encoutering this issue in other PRs. Should we disable caching if the linter cache is growing? Currently I do not have much space to investigate it.

michaljurecko: @jachym-tousek-keboola Petr needs the fix, I am mergin it, you comment will be addresed in next PRs, if any.

michaljurecko:
> Looks good but how did you compare TCP and KCP? Did you write some benchmarks?

There is test, which logs write duration, see code.

If you want, write a benchmark, but as I wrote, it will not be sufficiently descriptive on localhost. 

I only needed such a simple comparison to find the issue, so the test was enough.

PR #1903, opened 2024-07-11
feat: Route records from HTTP source to sinks

michaljurecko:

**At first, please see the last commit [tests: Update HTTP source tests](https://github.com/keboola/keboola-as-code/pull/1903/commits/7f440142904dffdfb412f2cd1e0c316a922014a2).**

**Changes:**
- Create HTTP source records dispatcher:
  - Search for specified source.
  - Checks if the source is enabled.
  - Validate the `secret`.
  - Calls sink router.
- Created sink router.
  - Opens/closes sinks pipelines.
  - Check if the sink is enabled.
  - Routes requests from HTTP source to sinks pipelines.
- `recordctx`
  - Renamed `receivectx` to `recordctx`.
    - Added support for `fasthttp`.
    - Added laziness and lock, sinks can be mapped/encoded in parallel.

-----------

It would need a nice diagram, but we don't have one yet

`httpsource -> httpsource/dispatcher -> sink/router -> plugins.OpenSinkPipeline`


internal/pkg/service/stream/definition/key/branch.go
```diff
@@ -16,7 +15,7 @@ type BranchKey struct {
 }
 
 func (v BranchKey) String() string {
-	return fmt.Sprintf("%s/%s", v.ProjectID.String(), v.BranchID.String())
+	return strconv.Itoa(int(v.ProjectID)) + "/" + v.BranchID.String()
 }
```
michaljurecko, line 19: It has measurable impact on http source performance.

go.mod
```diff
@@ -36,6 +36,7 @@ require (
 	github.com/jarcoal/httpmock v1.3.1
 	github.com/joho/godotenv v1.5.1
 	github.com/jpillora/longestcommon v0.0.0-20161227235612-adb9d91ee629
+	github.com/json-iterator/go v1.1.12
 	github.com/keboola/go-client v1.26.5
```
michaljurecko, line 40: HTTP source is using faster JSON implementation,
same as in the  [go-client](https://github.com/keboola/go-client/blob/main/pkg/client/json.go).

internal/pkg/encoding/jsonnet/vm.go
```diff
@@ -16,16 +16,16 @@ type VM[T any] struct {
...
 	vm      *jsonnet.VM
 	err     error
-	payload *T
+	payload T
 }
 
 // Payload returns the payload of the current evaluation.
 // Can be used by custom functions for request specific values.
-func (vm *VM[T]) Payload() *T {
+func (vm *VM[T]) Payload() T {
```
michaljurecko, line 24: I need here INTERFACE `recordctx.Context` (there is HTTP and FastHTTP implemetation).

internal/pkg/service/stream/mapping/recordctx/fasthttp.go
```diff
@@ -0,0 +1,145 @@
+package recordctx
+
+import (
```
michaljurecko, line 3: recordctx provides information about such as: headers, body, ...
- It is lazy.
- It is thread/goroutine free.

internal/pkg/service/stream/plugin/sink.go
```diff
@@ -0,0 +1,26 @@
...
+
+func (p *Plugins) OpenSinkPipeline(ctx context.Context, sink definition.Sink) (pipeline.Pipeline, error) {
+	for _, fn := range p.sinkPipelineOpeners {
+		p, err := fn(ctx, sink)
+		if err != nil {
+			return nil, err
+		}
+		if p != nil {
+			return p, nil
+		}
```
Matovidlo, line 23: Go standard says, taht when `err != nil` the `p` is `nil`. When `err == nil` the `p != nil`. Therefore the if is not needed, so it is not according the standard flow check. Why the function can return `p and err both nil`?

internal/pkg/service/stream/sink/router/pipeline.go
```diff
@@ -0,0 +1,131 @@
...
+	// it is initialized when the first record is received.
+	pipeline pipeline.Pipeline
+
+	openError      error
+	openBackoff    backoff.BackOff
+	openRetryAfter time.Time
+}
+
+// pipeline gets or creates sink pipeline.
+func (r *router) pipeline(ctx context.Context, timestamp time.Time, sinkKey key.SinkKey) (pipeline.Pipeline, error) {
```
Matovidlo, line 31: Why is router method in `pipeline` file? This should be in my opinion in `router.go`. The definition is elsewhere and first function after `type` definition is definiton of another file.

internal/pkg/service/stream/sink/router/pipeline.go
```diff
@@ -0,0 +1,131 @@
...
+	if p, found := r.pipelines[sinkKey]; found {
+		p.lock.Lock()
+		delete(r.pipelines, sinkKey)
+		p.lock.Unlock()
+
+		r.wg.Add(1)
+		go func() {
+			defer r.wg.Done()
+			p.close(ctx, reason)
+		}()
```
Matovidlo, line 98: This should be invoked once, why do we use waitgroup as it is called once? I would expect it is called directly without `go` routine.

internal/pkg/encoding/jsonnet/vm.go
Matovidlo: Maybe if it is not working well, we just switch the implementation to `HTTP` one.

michaljurecko: @jachym-tousek-keboola 
- `TestSource` endpoint is working with Goa/standard HTTP library, so both versions are needed.
- If some new type of Source will be implemented in the future, then another implementation will also be added.

internal/pkg/service/stream/definition/repository/sink/sink_watch.go
```diff
@@ -0,0 +1,14 @@
+package sink
+
+import (
+	"context"
+
+	clientv3 "go.etcd.io/etcd/client/v3"
```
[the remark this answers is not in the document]
michaljurecko: :man_shrugging: IDE probably got it different, I'll fix it

internal/pkg/service/stream/mapping/recordctx/fasthttp.go
```diff
@@ -0,0 +1,145 @@
...
+	"github.com/valyala/fasthttp"
+
+	"github.com/keboola/keboola-as-code/internal/pkg/utils/errors"
+)
+
+type fastHTTPContext struct {
+	timestamp     time.Time
+	req           *fasthttp.RequestCtx
+	lock          sync.Mutex
+	ip            net.IP
```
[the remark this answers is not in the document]
michaljurecko: :+1:

internal/pkg/service/stream/plugin/sink.go
michaljurecko: https://github.com/keboola/keboola-as-code/blob/7f440142904dffdfb412f2cd1e0c316a922014a2/internal/pkg/service/stream/sink/pipeline/pipeline.go#L26-L29

I will copy the comment also here.

internal/pkg/service/stream/sink/router/pipeline.go
michaljurecko: It is related, but I can move it, if you want.

internal/pkg/service/stream/sink/router/pipeline.go
michaljurecko: It is correct, we don't want to block method callers, multiple pipelines are closed in parallel.

internal/pkg/service/stream/definition/key/branch.go
michaljurecko: Usually, it doesn't need to be solved, but at ~100,000 calls/s, it has some impact.

internal/pkg/service/stream/plugin/sink.go
michaljurecko: https://github.com/keboola/keboola-as-code/pull/1903/commits/373b6b71c26f9d085ac7f6cb3f5315a121d04d87

internal/pkg/service/stream/definition/repository/sink/sink_watch.go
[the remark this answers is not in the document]
michaljurecko: https://github.com/keboola/keboola-as-code/pull/1903/commits/373b6b71c26f9d085ac7f6cb3f5315a121d04d87

internal/pkg/service/stream/mapping/recordctx/fasthttp.go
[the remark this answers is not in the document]
michaljurecko: https://github.com/keboola/keboola-as-code/pull/1903/commits/373b6b71c26f9d085ac7f6cb3f5315a121d04d87

internal/pkg/service/stream/sink/router/pipeline.go
Matovidlo: https://github.com/keboola/keboola-as-code/pull/1906

internal/pkg/service/stream/plugin/sink.go
Matovidlo: https://github.com/keboola/keboola-as-code/pull/1906

Matovidlo (commented):
Create another PR with suggested changes

internal/pkg/service/stream/plugin/sink.go
michaljurecko: :heavy_check_mark:

internal/pkg/service/stream/sink/router/router.go
```diff
@@ -0,0 +1,338 @@
...
+		r.wg.Wait()
+
+		// Wait for closing all pipelines
+		r.closeAllPipelines(ctx, "shutdown")
+		r.wg.Wait()
+
+		r.logger.Infof(ctx, "sink router shutdown done")
+	})
+
+	// Start sinks mirroring, only necessary data is saved
```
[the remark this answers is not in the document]
michaljurecko: https://github.com/keboola/keboola-as-code/blob/57ea30452de29d209d2d23acd7726a6b705249d1/internal/pkg/service/common/etcdop/watchmirror.go#L13-L17

internal/pkg/service/stream/sink/router/router.go
```diff
@@ -0,0 +1,338 @@
...
+			StartMirroring(ctx, &r.wg)
+		if err := <-errCh; err != nil {
+			return nil, err
+		}
+	}
+
+	return r, nil
+}
+
+func (r *router) DispatchToSources(sources []key.SourceKey, c recordctx.Context) SourcesResult {
```
hosekpeter, line 134: why do we have DispatchToSources and DispatchToSource? When do we use Sources and when Source?

hosekpeter (approved):

internal/pkg/service/stream/sink/router/router.go
michaljurecko: We now , use the top level `DispatchToSources` method now, but in the future, and for tests, also other make sense.

---------

HTTP source:

- `SourceKey` - contains branch ID
  - It can have different settings per branch.
- But, HTTP source endpoint is one, for all branches: `/stream/<projectID/<sourceID>/<secret>`.
- And the record should be dispatched to all branches (so to all `Sources` entities), to all sinks in the sources.

So there are 3 layers:
- `Sources` (branches)
- `Souces` (per branch)
- `Sink` (per branch and source)

PR #1905, opened 2024-07-12
feat: Add hostname field to volume metadata

michaljurecko:

**Changes:**
- Added required `hostname` field to stream service config.
- Added required `hostname` field to volume metadata entity.
- So Source node can create network connection to Disk Writer node based on volume metadata from DB.

-----------


provisioning/stream/kubernetes/templates/config/config-map.yaml
```diff
@@ -145,50 +147,50 @@ data:
               static: 100MB
               # Allocate disk space as % from the previous slice size. Validation rules: min=100,max=500
               relative: 110
-            encoding:
-              encoder:
-                # Concurrency of the format writer for the specified file type. 0 = auto = num of CPU cores. Validation rules: min=0,max=256
+          encoding:
+            encoder:
```
michaljurecko, line 151: Here was typo in indentation, all fields have default values, so no error was reported.

Matovidlo (approved):
LGTM

hosekpeter (approved):

PR #1908, opened 2024-07-12
feat: Writer volume - report also node port

michaljurecko:

**Changes:**
- Disk writer volume is reported to etcd with `hostname:port`, not only with `hostname`.
- The port can be loaded from the configuration, but this will not work if the configuration of all nodes is not the same.

-----------


internal/pkg/service/common/prefixtree/prefixtree.go
```diff
@@ -21,6 +21,7 @@ type TreeReadOnly[T any] interface {
 	AllFromPrefix(key string) []T
 	FirstFromPrefix(key string) (value T, found bool)
 	LastFromPrefix(key string) (value T, found bool)
+	WalkAll(fn func(key string, value T) (stop bool))
 	WalkPrefix(key string, fn func(key string, value T) (stop bool))
```
michaljurecko, line 25: Added missing method to the interface.

internal/pkg/service/stream/storage/level/local/diskreader/volume_test.go
```diff
@@ -296,26 +296,26 @@ func newVolumeTestCase(tb testing.TB) *volumeTestCase {
...
 	info := volumeModel.Spec{
-		NodeID:   tc.VolumeNodeID,
-		Hostname: tc.VolumeNodeHostname,
-		Path:     tc.VolumePath,
-		Type:     tc.VolumeType,
-		Label:    tc.VolumeLabel,
+		NodeID:      tc.VolumeNodeID,
+		NodeAddress: tc.VolumeNodeAddress,
+		Path:        tc.VolumePath,
+		Type:        tc.VolumeType,
```
michaljurecko, line 317: `hostname` -> `address = hostname:port`

Matovidlo (approved):
LGTM

hosekpeter (approved):

PR #1909, opened 2024-07-14
feat: Add column.Renderer to csv.Encoder, modify the interface

michaljurecko:

**Changes:**
- `column.Renderer` is now part of the `csv.Encoder`.
- Modified the pipeline interface:
  - From: `WriteRecord(values []any) error`
  - To: `WriteRecord(record recordctx.Context) error`
- Why:
  - It simplifies other parts of the service.
  - CSV encoding is no more divided to two parts, the encoder gets record context and produces CSV rows.
  - Other formats will be implemented more easily if needed.

image

### DIFF
--- a/internal/pkg/service/stream/mapping/recordctx/fasthttp.go
+++ b/internal/pkg/service/stream/mapping/recordctx/fasthttp.go
@@ -2,7 +2,6 @@ package recordctx
 
 import (
 	"context"
-	"io"
 	"net"
 	"net/http"
 	"sort"
@@ -25,8 +24,6 @@ type fastHTTPContext struct {
 	headersString *string
 	bodyString    *string
 	bodyStringErr error
-	bodyBytes     []byte
-	bodyBytesErr  error
 	bodyMap       *orderedmap.OrderedMap
 	bodyMapErr    error
 }
@@ -87,7 +84,7 @@ func (c *fastHTTPContext) BodyString() (string, error) {
 	defer c.lock.Unlock()
 
 	if c.bodyString == nil && c.bodyStringErr == nil {
-		if bytes, err := c.bodyBytesWithoutLock(); err == nil {
+		if bytes, err := c.BodyBytes(); err == nil {
 			v := string(bytes)
 			c.bodyString = &v
 		} else {
@@ -103,9 +100,7 @@ func (c *fastHTTPContext) BodyString() (string, error) {
 }
 
 func (c *fastHTTPContext) BodyBytes() ([]byte, error) {
-	c.lock.Lock()
-	defer c.lock.Unlock()
-	return c.bodyBytesWithoutLock()
+	return c.req.Request.Body(), nil // returned buffer is valid until the request is released
 }
 
 func (c *fastHTTPContext) BodyMap() (*orderedmap.OrderedMap, error) {
@@ -113,7 +108,7 @@ func (c *fastHTTPContext) BodyMap() (*orderedmap.OrderedMap, error) {
 	defer c.lock.Unlock()
 
 	if c.bodyMap == nil && c.bodyMapErr == nil {
-		if bodyBytes, err := c.bodyBytesWithoutLock(); err != nil {
+		if bodyBytes, err := c.BodyBytes(); err != nil {
 			c.bodyMapErr = err
 		} else if bodyMap, err := parseBody(string(c.req.Request.Header.ContentType()), bodyBytes); err != nil {
 			c.bodyMapErr = errors.PrefixError(err, "cannot parse request body")
@@ -123,13 +118,6 @@ func (c *fastHTTPContext) BodyMap() (*orderedmap.OrderedMap, error) {
 	}
 
 	return c.bodyMap, c.bodyMapErr
-}
-
-func (c *fastHTTPContext) bodyBytesWithoutLock() ([]byte, error) {
-	if c.bodyBytes == nil && c.bodyBytesErr == nil {
-		c.bodyBytes, c.bodyBytesErr = io.ReadAll(c.req.RequestBodyStream())
-	}
-	return c.bodyBytes, c.bodyBytesErr
 }
 
 func (c *fastHTTPContext) headersToMap() *orderedmap.OrderedMap {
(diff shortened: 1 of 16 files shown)


internal/pkg/service/stream/mapping/recordctx/fasthttp.go
```diff
@@ -103,17 +100,15 @@ func (c *fastHTTPContext) BodyString() (string, error) {
 }
 
 func (c *fastHTTPContext) BodyBytes() ([]byte, error) {
-	c.lock.Lock()
-	defer c.lock.Unlock()
-	return c.bodyBytesWithoutLock()
+	return c.req.Request.Body(), nil // returned buffer is valid until the request is released
```
michaljurecko, line 103: Optimalization for the fasthttp, we can get directly the reused body buffer.

internal/pkg/service/stream/storage/level/local/diskwriter/network/transport/client.go
```diff
@@ -90,6 +90,8 @@ func NewClient(d clientDependencies, config network.Config, nodeID string) (*Cli
 		}
 		sessWg.Wait()
 
+		// Wait for remaining goroutines
+		c.logger.Info(ctx, "waiting for goroutines")
```
michaljurecko, line 94: Added log messages for beter debuging of the TCP/KCP tests.

internal/pkg/service/stream/storage/level/local/diskwriter/network/transport/server.go
```diff
@@ -210,14 +211,19 @@ func (s *Server) acceptConnection(ctx context.Context, listener net.Listener) er
 		return err
 	}
 
+	if s.isClosed() {
+		_ = conn.Close()
+		return nil
+	}
+
 	// Create multiplexer
 	sess, err := yamux.Server(conn, multiplexerConfig(s.logger, s.config))
```
michaljurecko, line 220: I debuged the KCP test, with `-timeout 2s` flag, so I have seen, where it is blocked.

The key is to realize that the fact that the client has closed the connection does not mean that the server has already processed this information

So, after creating a connection/stream, we will check whether the server does not shut down, so that this connection/stream does not block the shutdown.

------------

We'll see if it's enough like this, it's a problem to simulate the issue locally.

internal/pkg/service/stream/storage/level/local/diskwriter/network/transport/transport_test.go
```diff
@@ -97,6 +102,17 @@ func testTransportSmallData(t *testing.T, protocol network.TransportProtocol) {
...
+	// Wait 2x for server handler
+	for range 2 {
+		select {
+		case <-ctx.Done():
+			assert.Fail(t, "timeout")
+		case <-receivedDone:
+		}
+	}
+	sort.Strings(received)
+	assert.Equal(t, []string{"bar", "foo"}, received)
```
michaljurecko, line 114: Here was missing assert.

internal/pkg/service/stream/storage/level/local/encoding/encoder/encoder.go
```diff
@@ -1,9 +1,11 @@
 package encoder
 
+import "github.com/keboola/keboola-as-code/internal/pkg/service/stream/mapping/recordctx"
+
 // Encoder writers record values as bytes to the underlying writer.
 // It is used inside the Writer pipeline, at the beginning, before the compression.
 type Encoder interface {
-	WriteRecord(values []any) error
+	WriteRecord(record recordctx.Context) error
 	Flush() error
```
michaljurecko, line 9: The key modification.

internal/pkg/service/stream/storage/level/local/encoding/encoder/csv/csv_benchmark_test.go
```diff
@@ -200,15 +206,9 @@ func BenchmarkCSVWrite(b *testing.B) {
...
-		// 8 columns, only the count is important for CSV
 		column.UUID{},
 		column.Datetime{},
 		column.Body{},
-		column.Template{},
-		column.Template{},
-		column.Template{},
-		column.Template{},
-		column.Template{},
 	}
```
michaljurecko, line 212: The benchmark now includes the `column.Renderer` so it must be updated.

Speeds are same, ... but there is no more zero memory allocations, because, the renderer must allocate some memory (6 allocs per op).

Matovidlo (approved):
LGTM

hosekpeter (approved):

internal/pkg/service/stream/storage/level/local/encoding/encoder/csv/csv.go
```diff
@@ -2,30 +2,55 @@ package csv
...
 
 type Encoder struct {
-	columns column.Columns
-	pool    *fastcsv.WritersPool
+	columns     column.Columns
+	writersPool *fastcsv.WritersPool
+	valuesPool  *sync.Pool
 }
 
+var columnRenderer = column.NewRenderer() //nolint:gochecknoglobals // contains Jsonnet VMs sync.Pool
```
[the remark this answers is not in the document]
michaljurecko: :man_shrugging: practically there is currently no difference

If the sync.Pool wasn't there, I'll create a new renderer for each encoder.

I would put it in dependencies if I needed to use dummy implementation in some tests, but it doesn't look like it would be necessary.

internal/pkg/service/stream/storage/level/local/encoding/pipeline.go
```diff
@@ -207,13 +209,8 @@ func (w *pipeline) WriteRecord(timestamp time.Time, values []any) error {
 		return errors.New(`writer is closed`)
 	}
 
-	// Check values count
-	if len(values) != len(w.slice.Columns) {
-		return errors.Errorf(`expected %d columns in the row, given %d`, len(w.slice.Columns), len(values))
-	}
```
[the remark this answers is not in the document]
michaljurecko: good question, I forgot to explain it.

It can't happen anymore, it can't be simulated, because it's now solved within one unit, so the number of columns is always the same.

PR #1911, opened 2024-07-15
feat: Add connection.Manager for Source -> Disk Writer nodes connections

michaljurecko:

**Changes:**
- Created `connection.Manager` + test with multiple writer nodes.

-----

**Please, see [the test](https://github.com/keboola/keboola-as-code/blob/michaljurecko-PSGO-591-conn-manager/internal/pkg/service/stream/storage/level/local/diskwriter/router/connection/connection_test.go) at first.**

-----------

https://github.com/keboola/keboola-as-code/blob/e443d62612e21389dffe0cbecf49cbfa2fd5d9ab/internal/pkg/service/stream/storage/level/local/diskwriter/router/connection/connection.go#L1-L8


internal/pkg/service/common/etcdop/session.go
```diff
@@ -161,7 +161,10 @@ func (b SessionBuilder) Start(ctx context.Context, wg *sync.WaitGroup, logger lo
 				if !errors.Is(err, context.Canceled) {
 					delay := s.backoff.NextBackOff()
 					s.logger.Infof(ctx, "waiting %s before the retry", delay)
-					<-time.After(delay)
+					select {
+					case <-ctx.Done():
+					case <-time.After(delay):
+					}
 				}
```
michaljurecko, line 168: Offtopic detail.

internal/pkg/service/stream/storage/level/local/volume/registration/registration.go
```diff
@@ -45,10 +46,18 @@ func RegisterVolumes[V volume.Volume](cfg Config, d dependencies, volumes *volum
...
 			}
-			return txn.Do(ctx).Err()
+			if err := txn.Do(ctx).Err(); err != nil {
+				err := errors.PrefixError(err, `cannot register volumes to database`)
+				logger.Error(ctx, err.Error())
+				return err
+			}
+
+			logger.Infof(ctx, `registered "%d" volumes`, len(all))
+			return nil
```
michaljurecko, line 60: Added log messages.

internal/pkg/service/stream/sink/router/pipeline.go
```diff
@@ -26,24 +28,80 @@ type pipelineRef struct {
 	openRetryAfter time.Time
 }
 
+func (p *pipelineRef) writeRecord(c recordctx.Context) (pipeline.RecordStatus, error) {
+	if err := p.ensureOpened(c.Ctx(), c.Timestamp()); err != nil {
+		return pipeline.RecordError, err
+	}
+	return p.pipeline.WriteRecord(c)
+}
+
```
michaljurecko, line 37: Offtopic: Shuffled the code so that it is clear how it works, and changes can be made there (next PR).

michaljurecko: Here you can see that the pipeline is lazy, and it is opened at the first use.

internal/pkg/service/stream/dependencies/dependencies.go
```diff
@@ -124,5 +125,5 @@ type LocalStorageScope interface {
 type Mocked interface {
 	dependencies.Mocked
 	TestConfig() config.Config
-	TestSinkPipelineController() *TestSinkPipelineController
+	TestSinkPipelineOpener() *pipeline.TestOpener
 }
```
michaljurecko, line 129: Improved tests.

internal/pkg/service/stream/storage/level/local/diskwriter/network/transport/client.go
```diff
@@ -99,8 +99,28 @@ func NewClient(d clientDependencies, config network.Config, nodeID string) (*Cli
...
+	if err != nil {
+		return nil, err
+	}
+
+	if err := <-initDone; err != nil {
+		return nil, err
+	}
+
+	return conn, nil
 }
```
michaljurecko, line 124: `ConnectTo` renamed to `OpenConnectionOrErr`, so if the connection attempt failed, the error is returned.

Added `OpenConnection`, there, it doesn't matter if the connection can't be established on the first try. The retry loop is always started.

internal/pkg/service/stream/storage/level/local/diskwriter/network/transport/client_conn.go
```diff
@@ -97,24 +114,26 @@ func (c *ClientConnection) Close() (err error) {
 	return err
 }
 
-func (c *ClientConnection) dialLoop(ctx context.Context, targetAddr string, initDone chan error) {
+func (c *ClientConnection) dialLoop(ctx context.Context, initDone chan error) {
 	// Close connection session on shutdown
 	defer func() {
-		if err := c.Close(); err != nil {
-			err := errors.PrefixError(err, `cannot close connection`)
-			logger.Error(ctx, err.Error())
```
michaljurecko, line 105: typo

internal/pkg/service/stream/storage/level/local/diskwriter/router/connection/connection_test.go
```diff
@@ -0,0 +1,184 @@
...
+	"github.com/keboola/keboola-as-code/internal/pkg/utils/netutils"
+)
+
+func TestConnectionManager(t *testing.T) {
+	t.Parallel()
+
+	ctx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
+	defer cancel()
+
+	etcdCfg := etcdhelper.TmpNamespace(t)
```
michaljurecko, line 33: At first glance, a simple, but quite complex and important test.

It is nice that within one test you can start several nodes and test the interactions between them.

---------------------

To make it perfect, it would be nice to test network outages using `toxiproxy`, but there is no time for that now.

michaljurecko: `etcdCfg` - all nodes must use the same etcd namespace/credentials.

internal/pkg/service/common/etcdop/watchmirror.go
```diff
@@ -75,36 +80,43 @@ func SetupMirror[T any, V any](
 	}
 }
 
-func (s MirrorSetup[T, V]) StartMirroring(ctx context.Context, wg *sync.WaitGroup) (mirror *Mirror[T, V], initErr <-chan error) {
-	mirror = &Mirror[T, V]{
+func (s MirrorSetup[T, V]) Build() *Mirror[T, V] {
+	return &Mirror[T, V]{
```
michaljurecko, line 84: I have to create separate methods: `MirrorSetup.Build` (maybe it should be `MirrorBuilder`, rename it if you want), `Mirror.StartMirroring`, see next comments.

internal/pkg/service/stream/storage/level/local/diskwriter/router/connection/connection.go
```diff
@@ -114,8 +113,8 @@ func NewManager(ctx context.Context, d dependencies, cfg network.Config, nodeID
 			WithOnUpdate(func(_ etcdop.MirrorUpdatedKeys[*volumeData]) {
 				m.updateConnections(ctx)
 			}).
-			StartMirroring(ctx, &m.wg)
-		if err := <-errCh; err != nil {
+			Build()
```
michaljurecko, line 116: Here is the issue:
- `WithOnUpdate` - `m.updateConnections` is using `m.volumes`.
- The on update callback is triggered by a background etcd watcher
- So Go detects a dara race here.

image

internal/pkg/service/stream/storage/level/local/diskwriter/router/connection/connection.go
```diff
@@ -0,0 +1,241 @@
...
+	StorageRepository() *storageRepo.Repository
+}
+
+func NewManager(ctx context.Context, d dependencies, cfg network.Config, nodeID string) (*Manager, error) {
+	m := &Manager{
+		logger:      d.Logger().WithComponent("storage.router.connections"),
+		connections: make(map[string]*connection),
+	}
+
+	ctx, cancel := context.WithCancel(ctx)
```
Matovidlo, line 75: I don't like much the shadowing, could we rename it ? e.g `managerCtx`

internal/pkg/service/stream/storage/level/local/diskwriter/router/connection/connection_test.go
```diff
@@ -0,0 +1,184 @@
...
+	s.Process().Shutdown(ctx, errors.New("bye bye source"))
+	s.Process().WaitForShutdown()
+	sourceLogger.AssertJSONMessages(t, `{"level":"info","message":"exited"}`)
+	sourceLogger.Truncate()
+
+	// Shutdown w2 and w4
+	w2.Process().Shutdown(ctx, errors.New("bye bye writer 2"))
+	w2.Process().WaitForShutdown()
+	w4.Process().Shutdown(ctx, errors.New("bye bye writer 4"))
+	w4.Process().WaitForShutdown()
```
Matovidlo, line 110: What if we call `close()` without defer here. Does the shutdown works correctly with calling `close()` of context directly? I would do it for the test, as this Shutdowns are going directly and takes context that was created at the start of the test.

internal/pkg/service/stream/storage/level/local/diskwriter/router/connection/connection.go
```diff
@@ -0,0 +1,241 @@
...
+						Node: &nodeData{
+							ID:      vol.NodeID,
+							Address: vol.NodeAddress,
+						},
+					}
+				},
+			).
+			WithOnUpdate(func(_ etcdop.MirrorUpdatedKeys[*volumeData]) {
+				m.updateConnections(ctx)
+			}).
```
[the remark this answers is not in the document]
michaljurecko: - Events are about volumes.
- One node can have multiple volumes.
- Volumes count is small.
- Volumes are registered on disk writer node startup, there are no other changes.
- So iterate all volumes on change is a simple KISS approach,

internal/pkg/service/stream/storage/level/local/diskwriter/router/connection/connection_test.go
michaljurecko: cc @jachym-tousek-keboola 

No, it doesn't work that way, it's not designed to be terminated using the context.

Each component has usually its own root context, some wait groups, ...

The component context is cancelled using the `OnShutdown` callback.

`servicectx.Process` is a LIFO of shutdown callbacks.

We don't need to turn off everything at once, but turn it off gradually, in the opposite order of how it started.

Example:
- close etcd session
- THEN close etcd client connection

OR:
- close all stream pipelines
- THEN close all streams in yamux multiplexer (there should be no opened stream)
- THEN close real TCP/KCP connetions (there should be no traffic)

internal/pkg/service/stream/storage/level/local/diskwriter/router/connection/connection.go
michaljurecko: @jachym-tousek-keboola 
In my opinion, shadowing is better in this use-case.
Because in the code bellow, you have only one `ctx` variable, and therefore you cannot make a mistake, by using some different context.

internal/pkg/service/stream/storage/level/local/diskwriter/network/transport/client_conn.go
```diff
@@ -87,6 +98,12 @@ func (c *ClientConnection) Close() (err error) {
 	c.lock.Lock()
 	defer c.lock.Unlock()
 
+	if c.isClosed() {
+		return nil
+	}
+
+	close(c.closed)
```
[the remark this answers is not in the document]
michaljurecko: Stream can be closed from both sides, mainly in tests, when the client and server nodes terminate in the same time.

We always call the termination of the stream/connection, even if it could have been terminated in the meantime by the other side.

It makes sense?

internal/pkg/service/stream/storage/level/local/diskwriter/network/transport/client_conn.go
```diff
@@ -97,24 +114,26 @@ func (c *ClientConnection) Close() (err error) {
...
 
-func (c *ClientConnection) dialLoop(ctx context.Context, targetAddr string, initDone chan error) {
+func (c *ClientConnection) dialLoop(ctx context.Context, initDone chan error) {
 	// Close connection session on shutdown
 	defer func() {
-		if err := c.Close(); err != nil {
-			err := errors.PrefixError(err, `cannot close connection`)
-			logger.Error(ctx, err.Error())
+		if !c.isClosed() {
+			if err := c.Close(); err != nil {
```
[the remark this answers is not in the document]
michaljurecko: See https://github.com/keboola/keboola-as-code/pull/1911#discussion_r1677958494

Yes, `c.isClosed ` is redundant here, it is inside the `c.Close`, I will remove it.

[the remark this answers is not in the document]
Matovidlo: Close it when it is done.

internal/pkg/service/stream/storage/level/local/diskwriter/router/connection/connection.go
Matovidlo: other approach is to call the context in the parameter `baseContext` for example.

internal/pkg/service/stream/storage/level/local/diskwriter/router/connection/connection_test.go
Matovidlo: When I will have time in future, I would like to do the graceful shutdown using context and waitgroups. However I would close this comment as my expectation was that it is possible to gracefully shutdown the program using `cancel()`.

Matovidlo (approved):
LGTM

internal/pkg/service/stream/storage/level/local/diskwriter/router/connection/connection_test.go
michaljurecko: It is graceful shutdown based on ctx and waitgroups, see `OnShutdown` occurences in the repo code.

This is a level higher, the termination of independent components (programs) that have some dependency on each other.

So component == program, in your comment.

------------------

All you have to do is prepare a short example, and I'll be happy to discuss it with you.

michaljurecko: To add some external reference, the `uber` is doing it in the same way, see `OnStop` hook.
https://github.com/uber-go/fx/blob/8fa92e9e702f362210d2eff5dfea6ac4b9b7a37b/docs/lifecycle.md?plain=1#L42

michaljurecko: Or https://www.rudderstack.com/blog/implementing-graceful-shutdown-in-go/

**Setup/Shutdown**

There are cases when blocking with ctx code is not the best approach ...

hosekpeter (approved):

PR #1912, opened 2024-07-16
feat: Add userError field to task result

jachym-tousek-keboola:
Jira: https://keboola.atlassian.net/browse/PSGO-679

**Changes:**
- Service errors details are now propagated to task result.

-----------

### DIFF
--- a/internal/pkg/service/common/task/node.go
+++ b/internal/pkg/service/common/task/node.go
@@ -60,6 +60,8 @@ type Node struct {
 	taskEtcdPrefix etcdop.PrefixT[Task]
 	taskLocksMutex *sync.Mutex
 	taskLocks      map[string]bool
+
+	exceptionIDPrefix string
 }
 
 type dependencies interface {
@@ -71,7 +73,7 @@ type dependencies interface {
 	EtcdSerde() *serde.Serde
 }
 
-func NewNode(nodeID string, d dependencies, opts ...NodeOption) (*Node, error) {
+func NewNode(nodeID string, exceptionIDPrefix string, d dependencies, opts ...NodeOption) (*Node, error) {
 	// Validate
 	if nodeID == "" {
 		panic(errors.New("task.Node: node ID cannot be empty"))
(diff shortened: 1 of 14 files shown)


Matovidlo (approved):
LGTM

hosekpeter (approved):

internal/pkg/service/common/task/node_test.go
```diff
@@ -502,7 +516,7 @@ task/123/my-receiver/my-export/some.task/%s
 {"level":"info","message":"started task","component":"task","task":"123/my-receiver/my-export/some.task/%s","node":"node2"}
 {"level":"debug","message":"lock acquired \"runtime/lock/task/my-lock\"","component":"task","task":"123/my-receiver/my-export/some.task/%s","node":"node2"}
 {"level":"info","message":"some message from the task (2)","component":"task","task":"123/my-receiver/my-export/some.task/%s","node":"node2"}
-{"level":"warn","message":"task failed (%s): some error (2) - unexpected [%s]","component":"task","task":"123/my-receiver/my-export/some.task/%s","node":"node2"}
+{"level":"warn","message":"task failed (%s): no space right on device","component":"task","task":"123/my-receiver/my-export/some.task/%s","node":"node2"}
 {"level":"debug","message":"lock released \"runtime/lock/task/my-lock\"","component":"task","task":"123/my-receiver/my-export/some.task/%s","node":"node2"}
```
michaljurecko, line 520: This is some funny typo, isn't it? `no space left on device`

internal/pkg/service/common/task/node.go
```diff
@@ -319,6 +320,23 @@ func (n *Node) runTask(logger log.Logger, task Task, cfg Config) (result Result,
...
+		if errWithUserMessage, ok := result.Error.(svcerrors.WithUserMessage); ok {
+			task.UserError.Message = errWithUserMessage.ErrorUserMessage()
+		} else {
+			task.UserError.Message = "Unknown error"
+		}
+
+		if errWithExceptionID, ok := result.Error.(svcerrors.WithExceptionID); ok {
+			task.UserError.ExceptionID = errWithExceptionID.ErrorExceptionID()
+		}
 	}
```
michaljurecko, line 340: The exceptionID should be always present in the message + optionally in the `expcetionId` field.

Workflow:
- User gets (in the UI) `something wrong `message with exception ID.
- He contacts our support.
- We see in the logs the full error message, with the same exception ID.

internal/pkg/service/common/task/node_test.go
Matovidlo: I would resolve it, it is just a test, can contain anything

internal/pkg/service/common/task/node.go
michaljurecko: Please see the `httpserver` package as reference, if there is no exceptionID you should generate it, for the log purposes mentioned above. The key is to somehow implement the `workflow`.

internal/pkg/service/common/task/node_test.go
michaljurecko: :+1: Ok, it was not clear to me whether or not it is some type of error that can be used in production as well.

michaljurecko:
I probably won't find time for this anymore, try to come to an agreement in the team.

internal/pkg/service/stream/components.go
```diff
@@ -38,7 +38,7 @@ func StartComponents(ctx context.Context, serviceScp dependencies.ServiceScope,
 	// Common task scope
 	var taskScp commonDeps.TaskScope
 	if componentsMap[ComponentAPI] {
-		taskScp, err = commonDeps.NewTaskScope(ctx, cfg.NodeID, serviceScp)
+		taskScp, err = commonDeps.NewTaskScope(ctx, cfg.NodeID, api.ExceptionIDPrefix, serviceScp)
```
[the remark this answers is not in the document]
michaljurecko: @jachym-tousek-keboola Instad of `keboola-stream-api-`, I suggest to use `keboola-stream-`, the constant can be in `components.go`, if you don't have a better place.

------------

The prefix doesn't matter, we could use `keboola-stream-` everywhere and not distinguish between API and background work if you want.

internal/pkg/service/common/task/node_test.go
```diff
@@ -423,6 +424,11 @@ task/123/my-receiver/my-export/some.task/%s
   "node": "node1",
   "lock": "runtime/lock/task/my-lock",
   "error": "some error (1) - expected",
+  "userError": {
+    "name": "unknownError",
+    "message": "Unknown error",
+    "exceptionId": "test-node-%s"
```
michaljurecko, line 430: `test-node` -> `test-service`, it's a bit confusing

PR #1913, opened 2024-07-16
fix: Telemetry events compatibility

jachym-tousek-keboola:
Jira: https://keboola.atlassian.net/browse/PSGO-687

**Changes:**
- Telemetry events of streams are now backwards compatible with buffer events.

-----------

### DIFF
--- a/internal/pkg/service/stream/storage/event/sender.go
+++ b/internal/pkg/service/stream/storage/event/sender.go
@@ -17,7 +17,8 @@ import (
 	"github.com/keboola/keboola-as-code/internal/pkg/utils/errors"
 )
 
-const componentID = keboola.ComponentID("keboola.keboola-stream")
+// Schema: https://github.com/keboola/event-schema/blob/main/schema/ext.keboola.keboola-buffer..json
+const componentID = keboola.ComponentID("keboola.keboola-buffer")
 
 type Sender struct {
 	logger log.Logger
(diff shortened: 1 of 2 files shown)


internal/pkg/service/stream/storage/event/sender.go
```diff
@@ -17,7 +17,8 @@ import (
 	"github.com/keboola/keboola-as-code/internal/pkg/utils/errors"
 )
 
-const componentID = keboola.ComponentID("keboola.keboola-stream")
+// Schema: https://github.com/keboola/event-schema/blob/main/schema/ext.keboola.keboola-buffer..json
+const componentID = keboola.ComponentID("keboola.keboola-buffer")
```
Matovidlo, line 21: We have new component stream, but we call it `buffer`. I don't like it. As we would like to deprecate it from project, this way we include telemetry compatibility. What's the use case of this PR? We would like to inspect new telemetry with `buffer` query or where we face problem of backward compatibility?

Matovidlo: https://keboolaglobal.slack.com/archives/C054DF6DZ1B/p1721213248279279?thread_ts=1721047861.443159&cid=C054DF6DZ1B

Matovidlo (approved):
LGTM

michaljurecko (approved):

PR #1916, opened 2024-07-17
fix: File/Slice entity must contain all needed cfg fields - cfg snapshot

michaljurecko:

**Changes:**
- The customization of settings per source/sink did not work (for some keys).
- The file/slice must contain full snapshot of configuration (with applied patches from the source/sink).
  - All required fields for the source/writer nodes must be present.
  - Because, if the stack/source/sink configuration is changed, in-flight writes must be completed according to the original settings.
  - Some changes in the configuration were not reflected, and there were missing cfg fields in Slice/File (they were loaded from some other config).

-----------


internal/pkg/service/stream/storage/level/local/diskreader/config.go
```diff
@@ -8,16 +8,13 @@ type Config struct {
 	// WaitForVolumeIDTimeout defines how long to wait for the existence of a file with the ID,
 	// see Open function and Volume.waitForVolumeID method.
 	WaitForVolumeIDTimeout time.Duration
-	// FileOpener provides file opening.
+	// OverwriteFileOpener overwrites file opening.
 	// A custom implementation can be useful for tests.
-	FileOpener FileOpener
+	OverwriteFileOpener FileOpener
```
michaljurecko, line 13: For values ​​that cannot be easily serialized, we use `nil`, these are some additional settings for tests. I moved the creation of the default implementation to other places in the code.

internal/pkg/service/stream/config/config_test.go
```diff
@@ -331,80 +330,37 @@ func TestTableSinkConfigPatch_ToKVs(t *testing.T) {
...
-  {
-    "key": "storage.level.local.volume.allocation.relative",
-    "type": "int",
-    "description": "Allocate disk space as % from the previous slice size.",
-    "value": 110,
-    "defaultValue": 110,
-    "overwritten": false,
-    "protected": true,
-    "validation": "min=100,max=500"
-  },
```
michaljurecko, line 371: I removed the disk allocation from the customizable settings. If it is needed in the future, it can be added.

This made refactoring easier.

internal/pkg/service/stream/storage/level/local/diskreader/config.go
michaljurecko: So I can use `encoding.Config` in the File/Slice entity.

internal/pkg/service/stream/storage/test/file.go
```diff
@@ -39,20 +41,22 @@ func NewFileOpenedAt(openedAtStr string) model.File {
 	fileKey := NewFileKeyOpenedAt(openedAtStr)
 	return model.File{
 		FileKey: fileKey,
-		Type:    model.FileTypeCSV,
```
Matovidlo, line 42: We completely get rid of Type, so it is not possible to define that we store anything else than CSV, slice as CSV etc?

internal/pkg/service/stream/storage/level/local/encoding/pipeline_test.go
```diff
@@ -33,14 +33,13 @@ func TestEncodingPipeline_Basic(t *testing.T) {
 	ctx := context.Background()
 	logger := log.NewDebugLogger()
 	clk := clock.New()
-	slice := test.NewSlice()
 
-	cfg := encoding.NewConfig()
-	cfg.Encoder.Factory = encoder.FactoryFn(dummyEncoderFactory)
+	slice := test.NewSlice()
+	slice.LocalStorage.Encoding.Encoder.Factory = encoder.FactoryFn(dummyEncoderFactory)
```
Matovidlo, line 37: How can storage contain Encoding? Storage stores bytes. Cannot they be mixed? (Encoded, not encoded) within `LocalStorage`.

internal/pkg/service/stream/storage/test/file.go
Matovidlo: Yes but I would expect that `file.Type` == `csv`. It makes quite a sense

internal/pkg/service/stream/config/config_test.go
michaljurecko: I think that there are none circumstances, that's why I threw it away.

internal/pkg/service/stream/storage/test/file.go
michaljurecko: > Yes but I would expect that `file.Type` == `csv`. It makes quite a sense

Yes, but the encoding pkg is responsible for that, and in order to connect it somehow sensibly, that's how it is now:
- `File.LocalStorage.Encoding.Type`
- `Slice.LocalStorage.Encoding.Type`

internal/pkg/service/stream/storage/level/local/diskreader/volume.go
```diff
@@ -167,11 +167,17 @@ func (v *Volume) OpenReader(slice *model.Slice) (r Reader, err error) {
 		v.removeReader(slice.SliceKey)
 	}()
 
+	// Get file opener
+	var opener FileOpener = DefaultFileOpener{}
+	if v.config.OverwriteFileOpener != nil {
+		opener = v.config.OverwriteFileOpener
```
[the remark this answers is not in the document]
michaljurecko: Ok, I will rename it.

internal/pkg/service/stream/storage/level/local/diskwriter/network/connection/connection_test.go
```diff
@@ -125,7 +125,7 @@ func TestConnectionManager(t *testing.T) {
 {"level":"info","message":"waiting %s for 0 streams","component":"storage.node.writer.network.server"}
 {"level":"info","message":"waiting for streams done","component":"storage.node.writer.network.server"}
 {"level":"info","message":"closing 0 streams","component":"storage.node.writer.network.server"}
-{"level":"info","message":"closing 0 sessions","component":"storage.node.writer.network.server"}
+{"level":"info","message":"closing %d sessions","component":"storage.node.writer.network.server"}
 {"level":"info","message":"waiting for goroutines","component":"storage.node.writer.network.server"}
 {"level":"info","message":"closed disk writer server","component":"storage.node.writer.network.server"}
```
michaljurecko, line 130: Here it is important, it is graceful shutdown of streams (`closing 0 streams`), the connection can be terminated first from server or client.

internal/pkg/service/stream/storage/level/local/diskreader/volume.go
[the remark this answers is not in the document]
michaljurecko: https://github.com/keboola/keboola-as-code/pull/1916/commits/1f54253dd5628dea3471e9830ca8f0ba298119eb

internal/pkg/service/stream/storage/level/local/diskreader/reader.go
```diff
@@ -50,19 +50,21 @@ type reader struct {
 func New(
 	ctx context.Context,
 	logger log.Logger,
-	slice *model.Slice,
+	sliceKey model.SliceKey,
+	localCompression compression.Config,
+	targetCompression compression.Config,
```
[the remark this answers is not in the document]
michaljurecko: The function gets only what it needs.

Watch mirror does not have to store unimportant parts.

Matovidlo (approved):
LGTM

hosekpeter (approved):

PR #1919, opened 2024-07-18
fix: Increase timeout of proxy concurrency test

Matovidlo:
**Changes:**
- Change behaviour of single context per test case -> context of single go routine

-----------

### DIFF
--- a/internal/pkg/service/appsproxy/proxy/proxy_test.go
+++ b/internal/pkg/service/appsproxy/proxy/proxy_test.go
@@ -1677,7 +1677,7 @@ func TestAppProxyRouter(t *testing.T) {
 		{
 			name: "concurrency-test",
 			run: func(t *testing.T, client *http.Client, m []*mockoidc.MockOIDC, appServer *testutil.AppServer, service *testutil.DataAppsAPI, dnsServer *dnsmock.Server) {
-				ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+				ctx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
 				defer cancel()
 
 				wg := sync.WaitGroup{}


internal/pkg/service/appsproxy/proxy/proxy_test.go
```diff
@@ -1677,15 +1677,14 @@ func TestAppProxyRouter(t *testing.T) {
...
 				wg := sync.WaitGroup{}
 				counter := atomic.NewInt64(0)
 				for i := 0; i < 100; i++ {
 					wg.Add(1)
 					go func() {
 						defer wg.Done()
+						ctx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
+						defer cancel()
 
 						m[0].QueueUser(&mockoidc.MockUser{
```
michaljurecko, line 1689: :-1: Why do you want to create 100 contexts when one is enough? 
Those goroutines must start with a difference of a few ms.

Matovidlo: I think that's how context should work. Each request has 30 seconds to finish instead of having test 30 seconds to finish. If I would do the context switching a lot I would have for example for last request just a few milliseconds (e.g 100 ms) and the test would fail.

michaljurecko: I think that all requests should end in 30 seconds, and they are started "at once". You can send an example here if you want - that would mean that CI is not able to run 100 goroutines and switch between them.

Matovidlo: Changed as you proposed, I don't see a point in this thread. Just 2 different opinions.

michaljurecko: Yes, 2 different options, both works, 1. with one ctx, 2. with 100x ctx :slightly_smiling_face: 

You choose the option with 100x ctx, I still don't understand why.

michaljurecko: In my opinion, this is the wrong reasoning

>  I would have for example for last request just a few milliseconds (e.g 100 ms) and the test would fail.

But agree with @jachym-tousek-keboola  what you prefer, but there should be some logical reason behind it.

Matovidlo: each request should take 30 seconds, not more not less.

michaljurecko: > not less

?

michaljurecko: all requests run in parallel and the difference in their start time is negligible

PR #1920, opened 2024-07-18
feat: Add column type path

jachym-tousek-keboola:
Jira: https://keboola.atlassian.net/browse/PSGO-729

**Changes:**
- Add column type path

-----------

Benchmark - around 10 times faster than jsonnet:

```
/code > go test ./internal/pkg/service/stream/mapping/table/column/renderer_benchmark_test.go -bench=. -benchmem
goos: linux
goarch: amd64
cpu: 11th Gen Intel(R) Core(TM) i7-1165G7 @ 2.80GHz
BenchmarkColumn_Path-8                   1000000              1124 ns/op             432 B/op         14 allocs/op
BenchmarkColumn_Template_Jsonnet-8        100167             13670 ns/op            9645 B/op        137 allocs/op
BenchmarkColumn_UUID-8                   1721452               710.6 ns/op           104 B/op          4 allocs/op
PASS
ok      command-line-arguments  4.566s

```


internal/pkg/service/stream/mapping/table/column/renderer_test.go
```diff
@@ -75,6 +76,166 @@ func TestRenderer_Headers(t *testing.T) {
...
+
+	body := `{"key1":[{"key2":"val2","key3":"val3"}]}`
+	header := http.Header{"Content-Type": []string{"application/json"}}
+
+	val, err := renderer.CSVValue(c, recordctx.FromHTTP(time.Now(), &http.Request{Header: header, Body: io.NopCloser(strings.NewReader(body))}))
+	assert.NoError(t, err)
+	assert.Equal(t, `{"key1":[{"key2":"val2","key3":"val3"}]}`, val)
+}
+
+func TestRenderer_Path_Json_UndefinedKey_Error(t *testing.T) {
```
Matovidlo, line 174: I miss case with out of index. e.g `key1[2].key3` case that end's up with an error.

internal/pkg/service/stream/mapping/table/column/renderer_test.go
```diff
@@ -75,6 +76,166 @@ func TestRenderer_Headers(t *testing.T) {
...
+	c := column.Path{
+		Path: "key1.key2",
+	}
+
+	body := `{"key1":{"key2":"val2"},"key3":"val3"}`
+	header := http.Header{"Content-Type": []string{"application/json"}}
+
+	val, err := renderer.CSVValue(c, recordctx.FromHTTP(time.Now(), &http.Request{Header: header, Body: io.NopCloser(strings.NewReader(body))}))
+	assert.NoError(t, err)
+	assert.Equal(t, "\"val2\"", val)
```
Matovidlo, line 92: "\\"val2\\"" -> \`"val2"\` so it's consistent

internal/pkg/service/stream/mapping/table/column/renderer_test.go
```diff
@@ -75,6 +76,166 @@ func TestRenderer_Headers(t *testing.T) {
...
+
+	body := `{"key1":[{"key2":"val2"},{"key3":"val3"}]}`
+	header := http.Header{"Content-Type": []string{"application/json"}}
+
+	val, err := renderer.CSVValue(c, recordctx.FromHTTP(time.Now(), &http.Request{Header: header, Body: io.NopCloser(strings.NewReader(body))}))
+	assert.NoError(t, err)
+	assert.Equal(t, `"val3"`, val)
+}
+
+func TestRenderer_Path_Json_ArrayIndex_RawString(t *testing.T) {
```
Matovidlo, line 143: This always returns representation of json object as string? E.g `"1"` or `1`?

michaljurecko:
I'm not doing full review, but it looks nice. :clap: @jachym-tousek-keboola 

---------

It has only one disadvantage, `OrderedMap` is used, so the entire JSON must be parsed and kept in memory.
- With large JSON inputs, a lot of time is spent getting one or a pair of values.
- A benchmark with some 100kB generated JSON would by nice.

As a improvement (next PR/next task) I recommend:

https://github.com/valyala/fastjson
- Use `ParserPool` provided by the library.
- Parse the body only once, add some lazy methods to the `recordctx.Context`.
- It should be faster with less memory allocations.

I purposely did not say it in advance, because it is a more complex implementation, which should be 15x faster.

internal/pkg/service/stream/api/openapi/openapi.yaml
```diff
@@ -1259,7 +1259,7 @@ definitions:
                         tableId: in.c-bucket.table
                         mapping:
                             columns:
-                                - type: uuid-v7
+                                - type: uuid
                                   name: id-col
                                   primaryKey: true
```
michaljurecko, line 1264: Check please migration script.

Matovidlo: https://github.com/keboola/keboola-as-code/pull/1756/commits/eb2591e66699ad9f33e711a4a0a0fa93f0a5c8b6

Matovidlo (approved):
LGTM

PR #1921, opened 2024-07-19
feat: gRPC prerequisites/dependencies

michaljurecko:

**Changes:**
- Prerequisites for gRPC PR.

-----------

### DIFF
--- a/internal/pkg/service/stream/components.go
+++ b/internal/pkg/service/stream/components.go
@@ -5,6 +5,8 @@ import (
 	"sort"
 	"strings"
 
+	commonDeps "github.com/keboola/keboola-as-code/internal/pkg/service/common/dependencies"
+	"github.com/keboola/keboola-as-code/internal/pkg/service/common/distlock"
 	"github.com/keboola/keboola-as-code/internal/pkg/service/stream/api"
 	"github.com/keboola/keboola-as-code/internal/pkg/service/stream/config"
 	"github.com/keboola/keboola-as-code/internal/pkg/service/stream/dependencies"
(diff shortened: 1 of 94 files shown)


internal/pkg/service/stream/storage/level/local/model/file.go
```diff
@@ -13,6 +12,4 @@ type File struct {
 	Assignment assignment.Assignment `json:"assignment"`
 	// Allocation configures pre-allocation of the filesystem free space.
 	Allocation diskalloc.Config `json:"allocation"`
-	// Encoding defines how is the  result format encoded, for example tabular data to CSV.
-	Encoding encoding.Config `json:"encoding"`
```
michaljurecko, line 17: Sorry, one more smaller refactoring, I need to pass only required configuration to the disk writer node, 

so `{file/slice}.LocalStorage.Encoding` has been moved to `{file/slice}.Encoding`, it is also related to some previous comment, in a previous PR.

Encoding is handled by the source node, in the "local" environment/level, ... but disk writer is providing "local storage",

internal/pkg/service/stream/keboolasink/bridge/fixtures/create_sink_snapshot.txt
```diff
@@ -148,7 +148,8 @@ storage/slice/level/local/123/456/my-source/my-sink/2000-01-01T01:00:00.000Z/my-
   },
   "local": {
     "dir": "123/456/my-source/my-sink/2000-01-01T01-00-00-000Z/2000-01-01T01-00-00-000Z",
-    "filename": "slice.csv.gz",
+    "filenamePrefix": "slice",
+    "filenameExtension": "csv.gz",
     "allocatedDiskSpace": "100MB"
```
michaljurecko: I need to split the `filename` field to the two, will be explained later (writing from each source node requires a separate disk file).

internal/pkg/service/stream/storage/level/local/diskwriter/volume.go
```diff
@@ -176,16 +177,16 @@ func (v *Volume) Metadata() volume.Metadata {
 	}
 }
 
-func (v *Volume) OpenWriter(slice *model.Slice) (w Writer, err error) {
+func (v *Volume) OpenWriter(sliceKey model.SliceKey, slice localModel.Slice) (w Writer, err error) {
 	// Check context
```
michaljurecko, line 181: Optimalization for etcd mirror.

internal/pkg/service/stream/components.go
```diff
@@ -27,26 +29,62 @@ type Components []Component
...
 	for _, c := range components {
 		componentsMap[c] = true
 	}
 
+	// Common distribution scope
+	var distScp commonDeps.DistributionScope
+	if componentsMap[ComponentStorageWriter] || componentsMap[ComponentHTTPSource] || componentsMap[ComponentStorageCoordinator] {
+		distScp = commonDeps.NewDistributionScope(cfg.NodeID, cfg.Distribution, d)
+	}
+
```
michaljurecko, line 43: I updated `dependencies` scopes, some of them were missing.

internal/pkg/service/stream/storage/level/local/model/file.go
Matovidlo: It's ok but now it makes more sense :smile:

internal/pkg/service/stream/components.go
```diff
@@ -27,26 +29,62 @@ type Components []Component
...
 
-func StartComponents(ctx context.Context, d dependencies.ServiceScope, cfg config.Config, components ...Component) error {
+func StartComponents(ctx context.Context, d dependencies.ServiceScope, cfg config.Config, components ...Component) (err error) {
 	componentsMap := make(map[Component]bool)
 	for _, c := range components {
 		componentsMap[c] = true
 	}
 
+	// Common distribution scope
+	var distScp commonDeps.DistributionScope
```
Matovidlo, line 39: `distScp` -> `distScope`, Combination of `dist` and `scp` is very weird as,  Scp is command for secure copy files, and I was unsure why we use that.

Matovidlo (approved):
LGTM

internal/pkg/service/stream/components.go
michaljurecko: Rename it however you want :slightly_smiling_face:  , but please in the whole repository.

PR #1925, opened 2024-07-23
fix: Add missing options to config-map

Matovidlo:


Matovidlo:
@jachym-tousek-keboola Should we also remove old configuration options?

PR #1927, opened 2024-07-23
fix: Filename in diskwriter, include source node ID

michaljurecko:
Jira: XXXXX

**Changes:**
- ...

-----------


internal/pkg/service/common/dependencies/mocked.go
```diff
@@ -56,11 +56,12 @@ type mocked struct {
...
-	enableTasks            bool
-	enableDistribution     bool
 	enableDistributedLocks bool
 	enableOrchestrator     bool
 
+	tasksNodeID        string
+	distributionNodeID string
+
 	ctx         context.Context
 	clock       clock.Clock
```
michaljurecko, line 66: In the first commit is fix of mocked dependencies, ... the `nodeID` is always provided by a method argument, ... so the ID is always visible in tests/dependencies, there is no default value.

internal/pkg/service/stream/storage/level/local/diskwriter/network/rpc/pb/networkFile.proto
```diff
@@ -14,7 +14,8 @@ service NetworkFile {
 }
 
 message OpenRequest {
-  SliceKey slice_key = 1;
+  string source_node_id = 1;
+  SliceKey slice_key = 2;
 }
```
michaljurecko, line 20: Source node reports its own node ID to the disk writer node, on network file open.

internal/pkg/service/stream/storage/level/local/diskwriter/writer_test.go
```diff
@@ -207,14 +207,16 @@ func TestWriter_AllocateSpace_Disabled(t *testing.T) {
...
+	SourceNodeID string
+	Volume       *diskwriter.Volume
+	Slice        *model.Slice
 }
 
 func newWriterTestCase(tb testing.TB) *writerTestCase {
 	tb.Helper()
 	tc := &writerTestCase{}
 	tc.volumeTestCase = newVolumeTestCase(tb)
+	tc.SourceNodeID = tc.SourceNodeID
```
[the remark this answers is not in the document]
michaljurecko: yes, linter found it, typo, fixed

PR #1954, opened 2024-08-14
feat: Stream file import

jachym-tousek-keboola:
Jira: https://keboola.atlassian.net/browse/PSGO-748

**Changes:**
- Implemented file import to keboola.

-----------


internal/pkg/service/stream/keboolasink/bridge/schema/file.go
```diff
@@ -0,0 +1,23 @@
...
+
+type (
+	// File is an etcd prefix that stores all Keboola-specific data we need for upload and import.
+	File struct {
+		etcdop.PrefixT[keboolasink.File]
+	}
+)
+
+func forFile(s *serde.Serde) File {
+	return File{PrefixT: etcdop.NewTypedPrefix[keboolasink.File]("storage/keboola/file", s)}
```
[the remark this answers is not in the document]
michaljurecko: yes, but if we had any production data, migration would be required

internal/pkg/service/stream/keboolasink/bridge/file.go
```diff
@@ -131,18 +131,38 @@ func (b *Bridge) registerFileImporter() {
 
 			// Get file upload credentials
 			var keboolaFile *keboolasink.File
-			err = b.schema.File().ForFile(file.FileKey).GetOrNil(b.client).WithResultTo(&keboolaFile).Do(ctx).Err()
+			prefix := b.schema.File().ForFile(file.FileKey)
```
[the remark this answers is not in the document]
michaljurecko: It represens a etcd `Key` which contains serialized entity of the type (`T`) `keboolasink.File`.

So it is not a `prefix`, but a `key`, `k`, ....

Maybe it wasn't completely bad as one line, I don't know if it makes sense to split it. But if so, I would probably call it `k` or `key`.

internal/pkg/service/stream/keboolasink/bridge/file.go
```diff
@@ -131,18 +131,38 @@ func (b *Bridge) registerFileImporter() {
...
+			err = prefix.GetOrNil(b.client).WithResultTo(&keboolaFile).Do(ctx).Err()
```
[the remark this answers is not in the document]
michaljurecko: yes, it is same but in 2 lines instead of 1

internal/pkg/service/stream/keboolasink/bridge/file.go
```diff
@@ -131,18 +131,38 @@ func (b *Bridge) registerFileImporter() {
...
-				return err
+			// Check if job already exists
+			var job *keboola.StorageJob
+			if keboolaFile.StorageJobID != nil {
+				job, err = api.GetStorageJobRequest(keboola.StorageJobKey{ID: *keboolaFile.StorageJobID}).Send(ctx)
+				if err != nil {
+					return err
+				}
+
+				if job.Status == "success" {
```
michaljurecko, line 151: It would be nice to add constants to go-client.

internal/pkg/service/stream/keboolasink/bridge/file.go
```diff
@@ -131,18 +131,38 @@ func (b *Bridge) registerFileImporter() {
...
+			// Create job to import data if no job exists yet or if it failed
+			if job == nil || job.Status == "error" {
+				job, err = api.LoadDataFromFileRequest(keboolaFile.TableKey, keboolaFile.FileKey).Send(ctx)
+				if err != nil {
+					return err
+				}
+
+				// Save job ID to etcd
+				keboolaFile.StorageJobID = &job.ID
+				prefix.Put(b.client, *keboolaFile).Do(ctx)
```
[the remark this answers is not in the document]
michaljurecko: There is no point in wrapping Put in AtomicOp, since there is no read phase there.

---------------

You have to check error, so `if err := prefix.Put(b.client, *keboolaFile).Do(ctx).Err(); err !=nil` ...

---------------

I'm thinking about putting an etcd lock ther, ... for various edge-cases when restarting pods and the like, ... we can discuss it on the next call.

internal/pkg/service/stream/storage/node/coordinator/fileimport/operator.go
```diff
@@ -216,6 +217,32 @@ func (o *operator) checkFile(ctx context.Context, file *fileData) {
...
+	}
+
+	// If there is an error, increment retry delay
+	if err != nil {
+		o.logger.Error(ctx, err.Error())
+		err := o.storage.File().IncrementRetryAttempt(file.FileKey, o.clock.Now(), err.Error()).Do(ctx).Err()
+		if err != nil {
+			o.logger.Errorf(ctx, "cannot increment file import retry: %s", err)
+		}
+		return
```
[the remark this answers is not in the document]
michaljurecko: Please, send links to `two similar error handling places`

internal/pkg/service/stream/keboolasink/bridge/file.go
hosekpeter: https://github.com/keboola/go-client/pull/139

internal/pkg/service/stream/config/config_test.go
```diff
@@ -232,11 +232,13 @@ storage:
...
             operator:
                 # Import triggers check interval. Validation rules: required,minDuration=100ms,maxDuration=30s
-                checkInterval: 1s
+                fileRotationCheckInterval: 1s
                 # Timeout of the file rotation operation. Validation rules: required,minDuration=30s,maxDuration=15m
                 fileRotationTimeout: 5m0s
                 # Timeout of the file close operation. Validation rules: required,minDuration=10s,maxDuration=10m
                 fileCloseTimeout: 1m0s
+                # Interval of checking files in the importing state. Validation rules: required,minDuration=100ms,maxDuration=30s
+                fileImportCheckInterval: 1s
```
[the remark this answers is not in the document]
michaljurecko: in principle it probably doesn't matter, but `500ms` should be enough, so use `500ms` please

internal/pkg/service/stream/keboolasink/bridge/file.go
```diff
@@ -63,7 +65,11 @@ func (b *Bridge) createStagingFile(ctx context.Context, now time.Time, sink defi
 		ctx,
 		attribute.String("token.ID", token.Token.ID),
 		attribute.String("file.name", name),
-		attribute.String("file.key", file.FileKey.String()),
+		attribute.String("project.id", file.ProjectID.String()),
+		attribute.String("branch.id", file.BranchID.String()),
+		attribute.String("source.id", file.SourceID.String()),
+		attribute.String("sink.id", file.SinkID.String()),
+		attribute.String("file.id", file.FileID.String()),
 	)
```
michaljurecko, line 73: Finally, the `file.Telemetry()` method should be used here:

```
attrs := file.Telemetry()
attrs = appent(attrs, attribute.String("token.ID", token.Token.ID), ...)
ctx = ...
```

Martin has some code prepared, so this can be finalized later.

internal/pkg/service/stream/keboolasink/bridge/file.go
```diff
@@ -89,9 +95,16 @@ func (b *Bridge) createStagingFile(ctx context.Context, now time.Time, sink defi
 
 	// Save credentials to database
 	ctx = ctxattr.ContextWith(ctx, attribute.String("file.resourceID", stagingFile.FileID.String()))
-	entity := keboolasink.FileUploadCredentials{SinkKey: file.SinkKey, FileUploadCredentials: *stagingFile}
+	entity := keboolasink.File{
+		SinkKey: file.SinkKey,
```
michaljurecko, line 99: Entity is probably not the best choice for a variable name, especially when the code is longer.

Maybe you can use `keboolaFile` ?

internal/pkg/service/stream/keboolasink/bridge/file.go
```diff
@@ -103,9 +116,70 @@ func (b *Bridge) deleteCredentialsOnFileDelete() {
...
+			var keboolaFile *keboolasink.File
+			err = b.schema.File().ForFile(file.FileKey).GetOrNil(b.client).WithResultTo(&keboolaFile).Do(ctx).Err()
+			if err != nil {
+				return err
+			}
+
+			if keboolaFile == nil {
+				return errors.New("missing keboola details for file")
+			}
+
```
michaljurecko, line 147: `GetOrNil` -> `GetOrErr` (we want to rename it to `GetMustExist`) and you can omit the `if keboolaFile == nil ` condition.

internal/pkg/service/stream/keboolasink/bridge/file.go
```diff
@@ -103,9 +116,70 @@ func (b *Bridge) deleteCredentialsOnFileDelete() {
...
+			// Create job to import data if no job exists yet or if it failed
+			if job == nil || job.Status == keboola.StorageJobStatusError {
+				job, err = api.LoadDataFromFileRequest(keboolaFile.TableKey, keboolaFile.FileKey).Send(ctx)
+				if err != nil {
+					return err
+				}
+
+				// Save job ID to etcd
+				keboolaFile.StorageJobID = &job.ID
+				b.schema.File().ForFile(file.FileKey).Put(b.client, *keboolaFile).Do(ctx)
```
michaljurecko, line 173: Here is missing error check `.Err()`

internal/pkg/service/stream/storage/node/coordinator/fileimport/operator.go
```diff
@@ -0,0 +1,272 @@
...
+	lock := o.locks.NewMutex(fmt.Sprintf("operator.sink.file.%s", file.FileKey.SinkKey))
+	if err := lock.Lock(ctx); err != nil {
+		o.logger.Errorf(ctx, "cannot lock %q: %s", lock.Key(), err)
+		return
+	}
+	defer func() {
+		if err := lock.Unlock(ctx); err != nil {
+			o.logger.Warnf(ctx, "cannot unlock lock %q: %s", lock.Key(), err)
+		}
+	}()
```
michaljurecko, line 241: Lock should be before the `o.plugins.ImportFile`

internal/pkg/service/stream/storage/node/coordinator/fileimport/operator.go
```diff
@@ -0,0 +1,272 @@
...
+		o.importFile(ctx, file)
+	default:
+		// nop
+	}
+}
+
+func (o *operator) importFile(ctx context.Context, file *fileData) {
+	o.logger.Info(ctx, "importing file")
+
+	var err error
```
michaljurecko, line 222: Here should be a `ctx` with th `FileImportTimeout` - the timeout applies to the whole operation, not only to the state switch.

internal/pkg/service/stream/storage/node/coordinator/fileimport/operator.go
```diff
@@ -0,0 +1,272 @@
...
+	}
+	defer func() {
+		if err := lock.Unlock(ctx); err != nil {
+			o.logger.Warnf(ctx, "cannot unlock lock %q: %s", lock.Key(), err)
+		}
+	}()
+
+	// Update the entity, the ctx may be cancelled
+	dbCtx, dbCancel := context.WithTimeout(context.WithoutCancel(ctx), o.config.FileImportTimeout.Duration())
+	defer dbCancel()
```
michaljurecko, line 245: DB timeout can be some smaller value than `FileImportTimeout`.

internal/pkg/service/stream/storage/node/coordinator/fileimport/operator_test.go
```diff
@@ -0,0 +1,321 @@
...
+	assert.Len(t, slices, 2)
+	assert.Equal(t, model.SliceUploaded, slices[0].State)
+	assert.Equal(t, model.SliceWriting, slices[1].State)
+
+	ts.clk.Add(ts.interval)
+
+	assert.EventuallyWithT(t, func(c *assert.CollectT) {
+		ts.logger.AssertJSONMessages(c, `{"level":"info","message":"successfully imported file","component":"storage.node.operator.file.import"}`)
+	}, 5*time.Second, 10*time.Millisecond)
+
```
michaljurecko, line 145: Isn't it in the wrong order?

First, I wait until the mirror syncs to the new state, then I trigger `checkFile` via `clk.Add`.

michaljurecko: In addition, some comments would be good, since it is not entirely clear why it is there - if you don't know - in detail - how it works.

internal/pkg/service/stream/storage/node/coordinator/filerotation/operator.go
```diff
@@ -150,19 +152,27 @@ func Start(d dependencies, config targetConfig.OperatorConfig) error {
...
 
 				// Update opened slices counts, per file
 				for _, event := range events {
+					fileKey := event.Value.FileKey
 					switch event.Type {
 					case etcdop.CreateEvent:
-						o.openedSlicesCount[event.Value.FileKey]++
+						o.openedSlicesCount[fileKey]++
+						if event.Value.State == model.SliceUploading {
+							o.filesNotEmpty[fileKey] = o.filesNotEmpty[fileKey] || !event.Value.LocalStorage.IsEmpty
```
michaljurecko, line 165: `filesNotEmpty` :thinking: 
- suggestions:
- `fileNotEmpty`
- `slicesNotEmpty`

michaljurecko: Maybe it could be 2 switch statementst, each with a comment what it is doing.

internal/pkg/service/stream/storage/node/coordinator/filerotation/operator.go
```diff
@@ -336,35 +366,40 @@ func (o *operator) closeFile(ctx context.Context, file *fileData) {
...
+
+	if err == nil {
+		o.logger.Info(ctx, "successfully closed file")
+	}
 }
 
-func (o *operator) waitForSlicesUpload(ctx context.Context, fileKey model.FileKey) error {
+func (o *operator) waitForSlicesUpload(ctx context.Context, fileKey model.FileKey) (allEmpty bool, err error) {
 	for {
 		// Order is important, to make it bulletproof, we have to get the notifier channel before the check
```
michaljurecko, line 387: :thinking: 

Maybe `allEmpty` shouldn't be part of this method.

It can have own lock and own method, or be inlined to the `closeFile`.

It is not directly related to waiting.

michaljurecko (commented):
I have a couple of questions/notes, maybe I misunderstood your intention somewhere, we can call about it.

internal/pkg/service/stream/keboolasink/bridge/schema/file.go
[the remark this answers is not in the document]
Matovidlo: Change all the `upload/crendentials` in etcd to `""` so the keys match. Probably we would do it aside but now it is not required. @michaljurecko just double check this

Matovidlo (approved):
LGTM.

PR #1958, opened 2024-08-21
feat: PProf and Datadog profiling

michaljurecko:
**Changes:**
- PProf and Datadog profiler can be enabled via config.
- Added docs with examples.

-----------

### DIFF
--- a/cmd/apps-proxy/main.go
+++ b/cmd/apps-proxy/main.go
@@ -3,11 +3,13 @@ package main
 
 import (
 	"context"
+	"net/http"
 	"os"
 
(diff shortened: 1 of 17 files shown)


Matovidlo (approved):
Should I create Jira task to update provisioning also for `apps-proxy` and `templates-api` for enabling profiling?

michaljurecko:
@Matovidlo `apps-proxy` and `templates-api` provisioning is using default values + ENVs - there is no change needed.
- The stream service has a lot of options, so it has `config-map` with the configuration file.

PR #1964, opened 2024-08-23
refactor: Improve mocked deps, fix minor issues

michaljurecko:
**Changes:**
- Default value of the `cfg.Hostname` is now `localhost` for all tests, so it doesn't have to be explicitly specified.
- Removed deprecated `orchestrator` package:
  - In the new streams service version, it is replaced with the etcd mirror and related utilis.
  - The main problem was that it had too much responsibility - and thus it was not immediately clear how it worked.
- Removed `TaskScope`, `DistributionScope`, `DistributedLockScope` from `common/mocked.go`.
  - This solves the problem with the node ID not being passed - there was static value `test-node`.
  - These scopes had no specific settings in mocked.go - they were useless there.
  - Now, the dependencies in the test are more closer to the production dependencies.
- Removed `mock.TestContext()`, context is passed as `ctx` argument.

-----------


internal/pkg/service/stream/components.go
```diff
@@ -35,6 +35,15 @@ func StartComponents(ctx context.Context, serviceScp dependencies.ServiceScope,
...
 
+	// Common task scope
+	var taskScp commonDeps.TaskScope
+	if componentsMap[ComponentAPI] {
+		taskScp, err = commonDeps.NewTaskScope(ctx, cfg.NodeID, serviceScp)
+		if err != nil {
+			return err
+		}
+	}
+
```
michaljurecko, line 46: TaskScope is required only by the `APIScope`, so it is no more part of the `ServiceScope`.

internal/pkg/mapper/template/jsonnetfiles/jsonnetfiles_test.go
```diff
@@ -11,7 +12,7 @@ import (
 
 func createStateWithMapper(t *testing.T, jsonnetCtx *jsonnet.Context) *state.State {
 	t.Helper()
-	d := dependencies.NewMocked(t)
+	d := dependencies.NewMocked(t, context.Background())
 	mockedState := d.MockedState()
```
michaljurecko, line 16: `ctx` is passed to the dependencies construtors, so there will be always only one root context in tests.

internal/pkg/service/stream/dependencies/coordinator.go
```diff
@@ -39,38 +30,36 @@ func (v *coordinatorScope) StatisticsL2Cache() *cache.L2 {
...
+	return NewMockedCoordinatorScopeWithConfig(tb, ctx, nil, opts...)
 }
 
-func NewMockedCoordinatorScopeWithConfig(tb testing.TB, modifyConfig func(*config.Config), opts ...dependencies.MockedOption) (CoordinatorScope, Mocked) {
+func NewMockedCoordinatorScopeWithConfig(tb testing.TB, ctx context.Context, modifyConfig func(*config.Config), opts ...dependencies.MockedOption) (CoordinatorScope, Mocked) {
 	tb.Helper()
-	svcScp, mock := NewMockedServiceScopeWithConfig(
-		tb,
-		modifyConfig,
-		append([]dependencies.MockedOption{dependencies.WithEnabledDistribution("test-node"), dependencies.WithEnabledDistributedLocks()}, opts...)...,
```
michaljurecko, line 59: There was static value `test-node` for the distribution scope.

At the same time, in this part of the code, the config was not yet available so that it was possible to use `cfg.NodeID`.

-----------

It was solved by removing the scopes from the `Mocked` interface, ... the `nodeID` is thus taken from the config, just like in the production setup.

Matovidlo (approved):
LGTM

PR #1965, opened 2024-08-23
feat: Stream file import (part 2)

jachym-tousek-keboola:
Jira: [keboola.atlassian.net/browse/PSGO-748](https://keboola.atlassian.net/browse/PSGO-748)

**Changes:**
- Finished missing details around file import

-----------

Builds on top of https://github.com/keboola/keboola-as-code/pull/1954 and https://github.com/keboola/keboola-as-code/pull/1950

### DIFF
--- a/internal/pkg/service/stream/storage/node/coordinator/fileimport/operator.go
+++ b/internal/pkg/service/stream/storage/node/coordinator/fileimport/operator.go
@@ -25,6 +25,8 @@ import (
 	"github.com/keboola/keboola-as-code/internal/pkg/utils/errors"
 )
 
+const dbOperationTimeout = 30 * time.Second
+
 type operator struct {
 	config       targetConfig.OperatorConfig
 	clock        clock.Clock
@@ -237,7 +239,7 @@ func (o *operator) importFile(ctx context.Context, file *fileData) {
 		o.logger.Error(ctx, err.Error())
 
 		// Update the entity, the ctx may be cancelled
-		dbCtx, dbCancel := context.WithTimeout(context.WithoutCancel(ctx), 30*time.Second)
+		dbCtx, dbCancel := context.WithTimeout(context.WithoutCancel(ctx), dbOperationTimeout)
 		defer dbCancel()
 
 		err := o.storage.File().IncrementRetryAttempt(file.FileKey, o.clock.Now(), err.Error()).RequireLock(lock).Do(dbCtx).Err()
@@ -270,7 +272,7 @@ func (o *operator) doImportFile(ctx context.Context, lock *etcdop.Mutex, file *f
 	}
 
 	// New context for database operation, we may be running out of time
-	dbCtx, dbCancel := context.WithTimeout(context.WithoutCancel(ctx), 30*time.Second)
+	dbCtx, dbCancel := context.WithTimeout(context.WithoutCancel(ctx), dbOperationTimeout)
 	defer dbCancel()
 
 	// Switch file to the imported state
--- a/internal/pkg/service/stream/storage/node/readernode/sliceupload/operator.go
+++ b/internal/pkg/service/stream/storage/node/readernode/sliceupload/operator.go
@@ -24,6 +24,8 @@ import (
 	"github.com/keboola/keboola-as-code/internal/pkg/utils/errors"
 )
 
+const dbOperationTimeout = 30 * time.Second
+
 type operator struct {
 	config     stagingConfig.OperatorConfig
 	clock      clock.Clock
@@ -204,7 +206,7 @@ func (o *operator) uploadSlice(ctx context.Context, volume *diskreader.Volume, s
 		o.logger.Error(ctx, err.Error())
 
 		// Update the entity, the ctx may be cancelled
-		dbCtx, dbCancel := context.WithTimeout(context.WithoutCancel(ctx), 30*time.Second)
+		dbCtx, dbCancel := context.WithTimeout(context.WithoutCancel(ctx), dbOperationTimeout)
 		defer dbCancel()
 
 		err = o.storage.Slice().IncrementRetryAttempt(slice.SliceKey, o.clock.Now(), err.Error()).Do(dbCtx).Err()
@@ -237,7 +239,7 @@ func (o *operator) doUploadSlice(ctx context.Context, volume *diskreader.Volume,
 	}
 
 	// New context for database operation, we may be running out of time
-	dbCtx, dbCancel := context.WithTimeout(context.WithoutCancel(ctx), 30*time.Second)
+	dbCtx, dbCancel := context.WithTimeout(context.WithoutCancel(ctx), dbOperationTimeout)
 	defer dbCancel()
 
 	// Switch slice to the uploaded state
--- a/internal/pkg/service/stream/storage/node/coordinator/filerotation/operator.go
+++ b/internal/pkg/service/stream/storage/node/coordinator/filerotation/operator.go
@@ -24,6 +24,8 @@ import (
 	"github.com/keboola/keboola-as-code/internal/pkg/utils/errors"
 )
 
+const dbOperationTimeout = 30 * time.Second
+
 type operator struct {
 	config       targetConfig.OperatorConfig
 	clock        clock.Clock
@@ -352,7 +354,7 @@ func (o *operator) closeFile(ctx context.Context, file *fileData) {
 	defer unlock()
 
 	// Update the entity, the ctx may be cancelled
-	dbCtx, dbCancel := context.WithTimeout(context.WithoutCancel(ctx), 30*time.Second)
+	dbCtx, dbCancel := context.WithTimeout(context.WithoutCancel(ctx), dbOperationTimeout)
 	defer dbCancel()
 
 	// If there is no error, switch file to the importing state
--- a/internal/pkg/service/stream/storage/node/coordinator/slicerotation/operator.go
+++ b/internal/pkg/service/stream/storage/node/coordinator/slicerotation/operator.go
@@ -25,6 +25,8 @@ import (
 	"github.com/keboola/keboola-as-code/internal/pkg/utils/errors"
 )
 
+const dbOperationTimeout = 30 * time.Second
+
 type operator struct {
 	config       stagingConfig.OperatorConfig
 	clock        clock.Clock
@@ -291,7 +293,7 @@ func (o *operator) closeSlice(ctx context.Context, slice *sliceData) {
 	}
 
 	// Update the entity, the ctx may be cancelled
-	dbCtx, dbCancel := context.WithTimeout(context.WithoutCancel(ctx), 30*time.Second)
+	dbCtx, dbCancel := context.WithTimeout(context.WithoutCancel(ctx), dbOperationTimeout)
 	defer dbCancel()
 
 	err = o.storage.Slice().SwitchToUploading(slice.SliceKey, o.clock.Now(), stats.Local.RecordsCount == 0).Do(dbCtx).Err()


internal/pkg/service/stream/keboolasink/bridge/file.go
```diff
@@ -115,62 +116,67 @@ func (b *Bridge) deleteCredentialsOnFileDelete() {
...
-				}
+	// Authorized API
+	api := b.publicAPI.WithToken(token.TokenString())
 
-				if job.Status == keboola.StorageJobStatusSuccess {
-					return nil
-				}
-			}
+	defer func() {
+		ctx, cancel := context.WithTimeout(ctx, uploadEventSendTimeout)
```
Matovidlo, line 138: I'll build a branch above this to set `uploadEventSendTimeout` configurable.

internal/pkg/service/stream/keboolasink/bridge/slice.go
```diff
@@ -15,50 +15,50 @@ import (
...
 		return err
 	}
 
 	defer func() {
-		err = reader.Close(ctx)
+		err := reader.Close(ctx)
 		if err != nil {
-			err = &plugin.ReaderClosureError{SliceKey: slice.SliceKey}
 			b.logger.Warnf(ctx, "unable to close reader: %v", err)
 			return
```
Matovidlo, line 38: This is not supposed to end up with error. Return `nil` and write that we set it to `nil` due to successfull upload and we expect to switch the state of slice to `Uploaded` even though the reader is closed.

internal/pkg/service/stream/keboolasink/bridge/slice.go
```diff
@@ -15,50 +15,50 @@ import (
...
 		}
 
 		ctx, cancel := context.WithTimeout(ctx, uploadEventSendTimeout)
-		err = b.SendSliceUploadEvent(ctx, b.publicAPI.WithToken(token.String()), time.Since(start), &err, slice, stats)
+		err = b.SendSliceUploadEvent(ctx, b.publicAPI.WithToken(token.String()), time.Since(start), &err, slice.SliceKey, stats)
 		cancel()
 		if err != nil {
-			err = &plugin.SendSliceUploadEventError{SliceKey: slice.SliceKey}
 			b.logger.Warnf(ctx, "unable to send slice upload event: %v", err)
 			return
```
Matovidlo, line 46: This is not supposed to end up with error. Return `nil` and write that we set it to `nil` due to successfull upload and we expect to switch the state of slice to `Uploaded` state even though the event was not dispatched.

internal/pkg/service/stream/keboolasink/bridge/file.go
```diff
@@ -115,62 +116,67 @@ func (b *Bridge) deleteCredentialsOnFileDelete() {
...
+		err = b.SendFileImportEvent(ctx, api, time.Since(start), &err, file.FileKey, stats)
+		cancel()
+		if err != nil {
+			b.logger.Warnf(ctx, "unable to send slice upload event: %v", err)
```
Matovidlo, line 142:
```suggestion
			b.logger.Warnf(ctx, "unable to send file upload event: %v", err)
```

internal/pkg/service/stream/keboolasink/bridge/file.go
```diff
@@ -115,62 +116,67 @@ func (b *Bridge) deleteCredentialsOnFileDelete() {
...
+			return
```
Matovidlo, line 143: This is not supposed to end up with error. Return `nil` and write that we set it to `nil` due to successfull import and we expect to switch the state of file to `Imported` state even though the event was not dispatched.

internal/pkg/service/stream/storage/level/local/diskreader/volume.go
```diff
@@ -127,26 +129,26 @@ func (v *Volume) Metadata() volume.Metadata {
...
 
 	// Setup logger
 	logger := v.logger.With(
-		attribute.String("projectId", slice.ProjectID.String()),
-		attribute.String("branchId", slice.BranchID.String()),
-		attribute.String("sourceId", slice.SourceID.String()),
-		attribute.String("sinkId", slice.SinkID.String()),
-		attribute.String("fileId", slice.FileID.String()),
-		attribute.String("sliceId", slice.SliceID.String()),
+		attribute.String("projectId", sliceKey.ProjectID.String()),
```
Matovidlo, line 140: Shouldn't this be part of telemetry, so the logger inherits it? Or is it incorrect and we have to do it this way?

internal/pkg/service/stream/storage/level/local/diskreader/volume.go
```diff
@@ -174,7 +176,7 @@ func (v *Volume) OpenReader(slice *model.Slice) (r Reader, err error) {
 	}
 
 	// Open file
-	filePath := slice.LocalStorage.FileName(v.Path(), "my-node")
+	filePath := slice.FileName(v.Path(), "my-node")
```
Matovidlo, line 179: Probably better to pass the `FileName` function instead of whole `LocalStorage` object ?

internal/pkg/service/stream/storage/level/local/diskreader/volume.go
```diff
@@ -127,26 +129,26 @@ func (v *Volume) Metadata() volume.Metadata {
 	}
 }
 
-func (v *Volume) OpenReader(slice *model.Slice) (r Reader, err error) {
+func (v *Volume) OpenReader(sliceKey model.SliceKey, slice localModel.Slice, encodingCompression compression.Config, stagingCompression compression.Config) (r Reader, err error) {
```
Matovidlo, line 132:
```suggestion
func (v *Volume) OpenReader(sliceKey model.SliceKey, slice localModel.Slice, encodingCompression, stagingCompression compression.Config) (r Reader, err error) {
```

internal/pkg/service/stream/keboolasink/bridge/file.go
Matovidlo: Just add comment here  https://github.com/keboola/keboola-as-code/pull/1968/files#diff-61f7272294f4d27b17650f7b104d5c118a303c08bb484f0c9f656bc600b0a868R10 :slightly_smiling_face:

internal/pkg/service/stream/keboolasink/bridge/slice.go
Matovidlo: Ok then add commentary.

internal/pkg/service/stream/storage/level/local/diskreader/volume.go
Matovidlo: @michaljurecko what do you think? Is it ok to have telemetry on `diskreader` level as there is no whole `model.File` present?

internal/pkg/service/stream/keboolasink/bridge/file.go
```diff
@@ -115,62 +115,55 @@ func (b *Bridge) deleteCredentialsOnFileDelete() {
...
+	}
 
-			// Get file upload credentials
-			var keboolaFile keboolasink.File
-			err = b.schema.File().ForFile(file.FileKey).GetOrErr(b.client).WithResultTo(&keboolaFile).Do(ctx).Err()
-			if err != nil {
-				return err
-			}
+	// Get file details
+	keboolaFile, err := b.schema.File().ForFile(file.FileKey).GetOrErr(b.client).Do(ctx).ResultOrErr()
```
hosekpeter, line 129: remove empty line

internal/pkg/service/stream/storage/node/readernode/sliceupload/operator.go
```diff
@@ -179,26 +191,28 @@ func (o *operator) checkSlice(ctx context.Context, data *sliceData) {
...
+	var err error
 	// Empty slice does not need to be uploaded to staging. Just mark as uploaded
-	if stats.Local.RecordsCount != 0 {
-		o.logger.Infof(ctx, `uploading slice %q`, data.Slice.SliceKey)
-		// Use plugin system to upload slice to stagin storage. Set is an in-progress upload
+	if !data.Slice.LocalStorage.IsEmpty {
+		// Get slice statistics
+		var stats statistics.Aggregated
+		stats, err = o.statistics.SliceStats(ctx, data.SliceKey)
+		if err != nil {
```
hosekpeter, line 206: Do we need var stats statistic.Aggregated fo this case?

Can't we use stats,err:=o.statistics.SliceStats(ctx, data.SliceKey) and remove var err error?

internal/pkg/service/stream/storage/node/coordinator/fileimport/operator.go
```diff
@@ -237,10 +237,17 @@ func (o *operator) importFile(ctx context.Context, file *fileData) {
 
 	var err error
 	if !file.IsEmpty {
-		// Import the file to specific provider
-		err = o.plugins.ImportFile(ctx, &file.File)
+		// Get file statistics
+		var stats statistics.Aggregated
```
michaljurecko, line 241: Offtopic: in the code bellow, there is `o.locks.NewMutex(fmt.Sprintf("operator.sink.file.%s", file.FileKey.SinkKey))`, you said you wanted to do it in a centralized way, so the name of the lock will be defined only in one place.

internal/pkg/service/stream/storage/node/coordinator/fileimport/operator.go
```diff
@@ -237,10 +237,17 @@ func (o *operator) importFile(ctx context.Context, file *fileData) {
...
+		stats, err = o.statistics.FileStats(ctx, file.FileKey)
 		if err != nil {
-			err = errors.PrefixError(err, "error when waiting for file import")
+			err = errors.PrefixError(err, "cannot get file statistics")
+		} else {
+			// Import the file to specific provider
+			err = o.plugins.ImportFile(ctx, &file.File, stats.Staging)
+			if err != nil {
+				err = errors.PrefixError(err, "error when waiting for file import")
+			}
```
michaljurecko, line 250: This error handling is quite error-prone, it would be better to wrap it in some method and use `return err`.

michaljurecko: So 

```
	var err error
	if !file.IsEmpty {
	  ....
	}
```

May be:
```
	err := o.doImportFile(....)
```

michaljurecko: Maybe also `dbCtx` and `SwitchToImported` may be part of the method.

So the `if` condition before the `IncrementRetryAttempt` will cover all errors.

`dbCtx` there would be 2x, but that doesn't seem like a problem to me.

michaljurecko: The `successfully imported file` may be closer to the `SwitchToImported`, in `if/else`.

internal/pkg/service/stream/storage/node/coordinator/filerotation/operator_test.go
```diff
@@ -27,11 +30,22 @@ import (
...
-func TestFileRotation(t *testing.T) {
-	t.Parallel()
+const fileExpirationDiff = time.Minute
 
-	ctx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
-	defer cancel()
+type testState struct {
+	interval      time.Duration
+	importTrigger targetConfig.ImportTrigger
+	clk           *clock.Mock
```
michaljurecko, line 38: Personally, I would reverse the order, first I would put the tests in the code, and then these helpers.

michaljurecko (approved):
LGTM

I don't see any big issue there, choose what makes sense to you from the comments/suggestions.

internal/pkg/service/stream/keboolasink/bridge/file.go
Matovidlo: already adjusted it. Probablly won't be reconfigured.

internal/pkg/service/stream/storage/node/readernode/sliceupload/operator.go
Matovidlo: In original code it is done like that. I would like linter to take care of such definitions. Can you check whether it is possible to enforce this sort of sorthanded definitions checker? :slightly_smiling_face:

internal/pkg/service/stream/storage/node/coordinator/filerotation/operator_test.go
Matovidlo: Can be reversed.

internal/pkg/service/stream/storage/node/readernode/sliceupload/operator.go
hosekpeter: @jachym-tousek-keboola Ohh sorry, I didn't notice that there aren't `returns`.

hosekpeter (approved):
LGTM

internal/pkg/service/stream/storage/node/coordinator/fileimport/operator.go
```diff
@@ -224,41 +224,22 @@ func (o *operator) importFile(ctx context.Context, file *fileData) {
...
 
+	// Import file
+	err := o.doImportFile(ctx, lock, file)
 	// If there is an error, increment retry delay
 	if err != nil {
 		o.logger.Error(ctx, err.Error())
+
+		// Update the entity, the ctx may be cancelled
+		dbCtx, dbCancel := context.WithTimeout(context.WithoutCancel(ctx), 30*time.Second)
+		defer dbCancel()
```
Matovidlo, line 241: I don't like that we have now 2 db contexts, so if we adjust one dbContext and the other not, we will come into inconsistency.
Either add constant for `dbContextTimeout` or refactor it so we have just 1 dbCtx.

Matovidlo (approved):

internal/pkg/service/stream/storage/node/coordinator/fileimport/operator.go
```diff
@@ -269,8 +250,36 @@ func (o *operator) importFile(ctx context.Context, file *fileData) {
...
+		}
+	}
+
+	// Update the entity, the ctx may be cancelled
+	dbCtx, dbCancel := context.WithTimeout(context.WithoutCancel(ctx), 30*time.Second)
+	defer dbCancel()
 
-	if err == nil {
-		o.logger.Info(ctx, "successfully imported file")
+	// Switch file to the importing state
```
michaljurecko, line 276: `// Switch file to the imported state`

internal/pkg/service/stream/storage/node/coordinator/fileimport/operator.go
```diff
@@ -224,41 +224,22 @@ func (o *operator) importFile(ctx context.Context, file *fileData) {
...
-	defer dbCancel()
-
-	// If there is no error, switch file to the importing state
-	if err == nil {
-		err = o.storage.File().SwitchToImported(file.FileKey, o.clock.Now()).RequireLock(lock).Do(dbCtx).Err()
-		if err != nil {
-			err = errors.PrefixError(err, "cannot switch file to the imported state")
-		}
-	}
+	defer unlocker()
```
michaljurecko, line 231: suggestion: `defer unlock()`

internal/pkg/service/stream/storage/node/coordinator/sinklock/lock.go
```diff
@@ -0,0 +1,25 @@
...
+	if err := lock.Lock(ctx); err != nil {
+		logger.Errorf(ctx, "cannot lock %q: %s", lock.Key(), err)
+		return lock, nil
+	}
+	return lock, func() {
+		if err := lock.Unlock(ctx); err != nil {
+			logger.Warnf(ctx, "cannot unlock lock %q: %s", lock.Key(), err)
+		}
+	}
+}
```
michaljurecko, line 25: :+1:

internal/pkg/service/stream/storage/node/coordinator/fileimport/operator_test.go
```diff
@@ -349,3 +269,83 @@ func TestFileImportEmpty(t *testing.T) {
...
+	source := test.NewHTTPSource(sourceKey)
+	ts.sink = dummy.NewSinkWithLocalStorage(key.SinkKey{SourceKey: source.SourceKey, SinkID: "my-sink"})
+	require.NoError(t, ts.dependencies.DefinitionRepository().Branch().Create(&branch, ts.clk.Now(), test.ByUser()).Do(ctx).Err())
+	require.NoError(t, ts.dependencies.DefinitionRepository().Source().Create(&source, ts.clk.Now(), test.ByUser(), "create").Do(ctx).Err())
+	require.NoError(t, ts.dependencies.DefinitionRepository().Sink().Create(&ts.sink, ts.clk.Now(), test.ByUser(), "create").Do(ctx).Err())
+}
+
+func (ts *testState) prepareFile(t *testing.T, ctx context.Context) model.File {
+	t.Helper()
+	files, err := ts.dependencies.StorageRepository().File().ListIn(ts.sink.SinkKey).Do(ctx).All()
```
michaljurecko, line 337: `prepareFile`/`prepareSlice` is little confusing, file/slice already exists.

michaljurecko (approved):
LGTM

internal/pkg/service/stream/storage/node/coordinator/fileimport/operator.go
michaljurecko: The timeout can be configurable, ... but 2 contexts were created here by creating the `doImportFile` function, which improves the code and error handling.

PR #1967, opened 2024-08-23
feat: Add wildcard support for AssertKeys function

Matovidlo:
**Changes:**
- Update test for assertion output.

-----------

### DIFF
--- a/internal/pkg/utils/etcdhelper/assert_test.go
+++ b/internal/pkg/utils/etcdhelper/assert_test.go
@@ -306,13 +306,11 @@ func TestAssertKeys_Difference(t *testing.T) {
 	// Expected error
 	wildcards.Assert(t, strings.TrimSpace(`
 %A
-	            	Diff:
-	            	--- Expected
-	            	+++ Actual
-	            	@@ -2,3 +2,3 @@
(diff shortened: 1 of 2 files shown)


internal/pkg/utils/etcdhelper/assert_test.go
```diff
@@ -306,13 +306,11 @@ func TestAssertKeys_Difference(t *testing.T) {
...
-	            	  (string) (len=4) "key1",
-	            	- (string) (len=4) "key3"
-	            	+ (string) (len=4) "key2"
-	            	 }
+	Error:      	These keys are in expected but not actual ectd state:
+	            	[001] key3
+
+%A
+	Error:      	These keys are in actual but not expected ectd state:
+	            	[001] key2
```
michaljurecko, line 314: Please add a new test with some wildcard, for example `%s`.

michaljurecko (commented):
Please add at least one test for the new feature - wildcards.

internal/pkg/utils/etcdhelper/assert_test.go
Matovidlo: added

PR #1968, opened 2024-08-23
feat: Add configuration for keboolabridge upload send event

Matovidlo:
**Changes:**
- Add configuration for keboolabridge upload send event.

-----------


internal/pkg/service/stream/config/config_test.go
```diff
@@ -271,6 +271,9 @@ storage:
                     slicesCount: 100
                     # Min remaining expiration to trigger file import. Validation rules: required,minDuration=5m,maxDuration=45m
                     expiration: 30m0s
+keboolaBridge:
+    # Timeout of uploading send event of slice or file
+    uploadEventSendTimeout: 30s
 `), strings.TrimSpace(string(bytes)))
```
michaljurecko, line 278: It should go to the `sink` part, e.g. `sink.keboola`, or `sink.keboola.table`.

-----------------

One timeout can be for all events:

`sink.keboola.table.eventSendTimeout`, or `sink.keboola.eventSendTimeout`

Matovidlo: Updated, added `importTimeout` as Jachym wanted.  We have discussion in another thread. If applicable, decide there or here whether it is enough to have single `timeout` for upload event.

michaljurecko (approved):
LGTM

PR #1970, opened 2024-08-26
feat: Add support for round robin pipeline balancer

Matovidlo:
**Changes:**
- Configuration of `pipelineBalancer`.
- Add support for RR pipeline balancer.

-----------

### DIFF
--- a/internal/pkg/service/stream/storage/level/local/diskwriter/network/config.go
+++ b/internal/pkg/service/stream/storage/level/local/diskwriter/network/config.go
@@ -7,11 +7,15 @@ import (
 )
 
 const (
-	TransportProtocolKCP = TransportProtocol("kcp")
-	TransportProtocolTCP = TransportProtocol("tcp")
+	TransportProtocolKCP                = TransportProtocol("kcp")
+	TransportProtocolTCP                = TransportProtocol("tcp")
+	RandomBalancerType     BalancerType = "rand"
+	RoundRobinBalancerType BalancerType = "roundRobin"
 )
 
 type TransportProtocol string
+
+type BalancerType string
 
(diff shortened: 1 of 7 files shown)


internal/pkg/service/stream/storage/level/local/diskwriter/network/config.go
```diff
@@ -42,5 +43,6 @@ func NewConfig() Config {
 		ShutdownTimeout:        30 * time.Second,
 		KCPInputBuffer:         8 * datasize.MB,
 		KCPResponseBuffer:      512 * datasize.KB,
+		PipelineBalancer:       "rand",
```
michaljurecko, line 46: Can you use the constant please? If there is an dependencies issues, move constant to the network package - where the configuration structure is.

internal/pkg/service/stream/storage/level/local/diskwriter/network/router/balancer.go
```diff
@@ -11,3 +19,17 @@ import (
...
+	case randomBalancerType:
+		return NewRandomBalancer(), nil
+
+	case roundRobinBalancerType:
+		return NewRoundRobinBalancer(), nil
+
+	default:
+	}
+
+	return nil, errors.New("inavlid balancer selected")
```
michaljurecko, line 34: It can be directy in the `default` branch.

michaljurecko (approved):

internal/pkg/service/stream/storage/level/local/diskwriter/network/config.go
Matovidlo: There is dependency issue, i'll move it

PR #1978, opened 2024-08-28
Update SnowflakeWriterComponentId function for GCP stack

hosekpeter:
Jira: [PSGO-767](https://keboola.atlassian.net/browse/PSGO-767)

**Changes:**
- added snowflake writer ID keboola.wr-db-snowflake-gcs-s3 
- added unit test

-----------

[PSGO-767]: https://keboola.atlassian.net/browse/PSGO-767?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/internal/pkg/template/jsonnet/function/function.go
+++ b/internal/pkg/template/jsonnet/function/function.go
@@ -9,6 +9,7 @@ package function
 import (
 	"github.com/google/go-jsonnet/ast"
 	"github.com/keboola/go-client/pkg/keboola"
+
 	"github.com/keboola/keboola-as-code/internal/pkg/encoding/jsonnet"
 	"github.com/keboola/keboola-as-code/internal/pkg/model"
 	"github.com/keboola/keboola-as-code/internal/pkg/template/input"
@@ -147,7 +148,7 @@ func SnowflakeWriterComponentID(components *model.ComponentsMap) *jsonnet.Native
 		Name:   `SnowflakeWriterComponentId`,
 		Params: ast.Identifiers{},
 		Func: func(params []any) (any, error) {
-			//for _, b := range backend {
+			// for _, b := range backend {
 			//	if b == project.BackendSnowflake {
 			//		if _, found := components.Get(SnowflakeWriterIDGCPS3); found {
 			//			return SnowflakeWriterIDGCPS3.String(), nil
--- a/internal/pkg/template/jsonnet/function/function.go
+++ b/internal/pkg/template/jsonnet/function/function.go
@@ -9,7 +9,6 @@ package function
 import (
 	"github.com/google/go-jsonnet/ast"
 	"github.com/keboola/go-client/pkg/keboola"
-
 	"github.com/keboola/keboola-as-code/internal/pkg/encoding/jsonnet"
 	"github.com/keboola/keboola-as-code/internal/pkg/model"
 	"github.com/keboola/keboola-as-code/internal/pkg/template/input"
@@ -20,6 +19,7 @@ const (
 	SnowflakeWriterIDAws   = keboola.ComponentID("keboola.wr-db-snowflake")
 	SnowflakeWriterIDAzure = keboola.ComponentID("keboola.wr-snowflake-blob-storage")
 	SnowflakeWriterIDGCP   = keboola.ComponentID("keboola.wr-db-snowflake-gcs")
+	SnowflakeWriterIDGCPS3 = keboola.ComponentID("keboola.wr-db-snowflake-gcs-s3")
 )
 
 // ConfigID Jsonnet function maps configuration ID used in the template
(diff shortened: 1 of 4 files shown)


Matovidlo:
@hosekpeter linter

internal/pkg/template/jsonnet/function/function.go
```diff
@@ -142,11 +144,18 @@ func ComponentIsAvailable(components *model.ComponentsMap) *jsonnet.NativeFuncti
 }
 
 // SnowflakeWriterComponentID Jsonnet function returns component ID of the Snowflake Writer it the stack.
-func SnowflakeWriterComponentID(components *model.ComponentsMap) *jsonnet.NativeFunction {
+func SnowflakeWriterComponentID(components *model.ComponentsMap, backend ...string) *jsonnet.NativeFunction {
```
Matovidlo, line 147:
```suggestion
func SnowflakeWriterComponentID(components *model.ComponentsMap, backends ...string) *jsonnet.NativeFunction {
```

internal/pkg/template/jsonnet/function/function.go
```diff
@@ -142,11 +144,18 @@ func ComponentIsAvailable(components *model.ComponentsMap) *jsonnet.NativeFuncti
...
 	return &jsonnet.NativeFunction{
 		Name:   `SnowflakeWriterComponentId`,
 		Params: ast.Identifiers{},
 		Func: func(params []any) (any, error) {
+			for _, b := range backend {
```
Matovidlo, line 152:
```suggestion
			for _, b := range backends {
```

internal/pkg/template/jsonnet/function/function.go
hosekpeter: [8e6264e](https://github.com/keboola/keboola-as-code/pull/1978/commits/8e6264ea49517da656ae72bc9c86069549b9ab1c)

internal/pkg/template/jsonnet/function/function.go
hosekpeter: [8e6264e](https://github.com/keboola/keboola-as-code/pull/1978/commits/8e6264ea49517da656ae72bc9c86069549b9ab1c)

internal/pkg/template/jsonnet/function/function.go
hosekpeter: @michaljurecko I wonder if it is correct to have an "options" parameter for this function or if this parameter should always be required. as for contextual use/load.

1. SnowflakeWriterComponentID(components *model.ComponentsMap, backend ...string)
2. SnowflakeWriterComponentID(components *model.ComponentsMap, backend []string)

We can add it either from templates.requirement or from dependencies.

I don't know exactly how these contexts work.

michaljurecko: The `backend` argument is now required to make good decision about `componentID`, but I don't understand why you used `backend ...string` instad of `backend []string`.

hosekpeter: Because this function is used: https://github.com/keboola/keboola-as-code/blob/ab8f731fddb84c4eff9037a0d582d0a1f99becc6/internal/pkg/template/context/load/context.go#L42

michaljurecko: Ok, so it is logical problem

`load` phase of the template runs on the Templates API startup and when the template update is detected.
- It is not project scoped operation :exclamation: So no backends are available here.
- Or in other words, the `SnowflakeWriterComponentID` was stack scoped, and you want to make it project scoped.
- The templates are pre-loaded so that they can be listed through the API and the template `use` is faster.
- The Snowflake Writer ID is used in `.keboola/repository.json`, in the list of components that a template contains.
  - See https://developers.keboola.com/cli/templates/structure/#snowflake-writer
  
  -----------
  
  
The only solution I can think of is to remove the function from the Load context.

Or refactor the Templates API complexly.

hosekpeter: But there is a field `requirements` and then the repository.json can include the `backend` or is it something else?

Matovidlo: What if we just include `[]string{}` backends as parameter? Is it better to just remove it?

hosekpeter: @Matovidlo yes i can change it and the remove this function from `load context`.

hosekpeter: @michaljurecko ok, thank for your explanation and your time. 

I will change it and remove it from `load context`

michaljurecko: Ok, but you should find out where it is used, and agree with the authors of the template.

michaljurecko: It looks like the placeholder `<keboola.wr-db-snowflake>`, is not used at all 
https://github.com/keboola/keboola-as-code-templates/blob/dde46bc71984e11949e482001a5b685670fa4e77/.keboola/repository.json

So maybe it is safe to remove it at all :man_shrugging:

Matovidlo (approved):

PR #1983, opened 2024-08-30
fix: e2e test stream bridge suggestions

Matovidlo:


test/stream/bridge/keboola/keboola_test.go
```diff
@@ -65,80 +79,88 @@ func TestKeboolaBridgeWorkflow(t *testing.T) {
...
+			[]model.Slice{
+				{State: model.SliceImported},
+				{State: model.SliceImported},
+				{State: model.SliceImported},
+				{State: model.SliceImported},
+				{State: model.SliceUploaded},
+				{State: model.SliceWriting},
+				{State: model.SliceUploaded},
+				{State: model.SliceWriting},
+			},
```
michaljurecko, line 163: Please, don't use the `model.File/Slice` for one field.

Create `expectedFile/Slice` struct, or use directly existing `[]FileState/SliceState`.

Matovidlo: 66ad417e103c503f0e35c1dc0afefbc5824c6261
I decided go with `File/Slice State` due to rework of clock to be mocked so we are able to check it as expected `model.File` to actual `model.File`
Here is sample of `assert.Equal` of fields that are problematic to check
```
BranchKey: (key.BranchKey) {
      OpenedAt: (utctime.UTCTime) {
  -    wall: (uint64) 0,
  -    ext: (int64) 0,
  +    wall: (uint64) 204000000,
  +    ext: (int64) 63860853766,
       loc: (*time.Location)(<nil>)
  @@ -33,3 +33,12 @@
    LocalStorage: (model.File) {
  -  Dir: (string) "",
  +  Dir: (string) (len=52) "154/15158/my-source/my-sink/2024-09-02T06-02-46-204Z",
     Assignment: (assignment.Assignment) {
      Config: (assignment.Config) {
  -    Count: (int) 0,
  -    PreferredTypes: ([]string) <nil>
  +    Count: (int) 2,
  +    PreferredTypes: ([]string) (len=1) {
  +     (string) (len=3) "hdd"
  +    }
      },
  -   Volumes: ([]model.ID) <nil>
  +   Volumes: ([]model.ID) (len=2) {
  +    (model.ID) (len=10) "BK1kuTCPcm",
  +    (model.ID) (len=10) "2nzYzYEW59"
  +   }
     },
     Allocation: (diskalloc.Config) {
  -   Enabled: (bool) false,
  -   Static: (datasize.ByteSize) 0,
  -   Relative: (int) 0,
  +   Enabled: (bool) true,
  +   Static: (datasize.ByteSize) 104857600,
  +   Relative: (int) 110,
      OverrideAllocator: (diskalloc.Allocator) <nil>
  @@ -78,6 +97,11 @@
      wall: (uint64) 0,
  -   ext: (int64) 0,
  +   ext: (int64) 63860857367,
      loc: (*time.Location)(<nil>)
  @@ -91,5 +115,5 @@
}
```
Probably for some fields could be used `wildcard` package but I don't know how to incorporate it within `assert.Equal` which enforces check of each file for exact value due to reflection

michaljurecko: For me, it's fine to check only the result state, maybe we will improve it in the future, ... I just don't like the abuse of the whole entity to check only the state.

michaljurecko (approved):

michaljurecko:
ping @jachym-tousek-keboola

PR #1984, opened 2024-09-02
feat: Skip slice upload and file import when Sink is disabled

jachym-tousek-keboola:
Jira: https://keboola.atlassian.net/browse/PSGO-776 + https://keboola.atlassian.net/browse/PSGO-777

**Changes:**
- Slice upload operator skips slices in a disabled or deleted sink
- File import operator skips files in a disabled or deleted sink

-----------

### DIFF
--- a/internal/pkg/service/stream/storage/node/coordinator/fileimport/operator.go
+++ b/internal/pkg/service/stream/storage/node/coordinator/fileimport/operator.go
@@ -48,9 +48,9 @@ type operator struct {
 }
 
 type fileData struct {
-	model.FileKey
-	model.FileState
-	model.Retryable
+	FileKey model.FileKey
+	State   model.FileState
+	Retry   model.Retryable
 	IsEmpty bool
 	File    plugin.File
 	Attrs   []attribute.KeyValue
@@ -64,7 +64,7 @@ type fileData struct {
 }
 
 type sinkData struct {
-	key.SinkKey
+	SinkKey key.SinkKey
 	Enabled bool
 }
 
@@ -123,10 +123,10 @@ func Start(d dependencies, config targetConfig.OperatorConfig) error {
 			},
 			func(_ string, file model.File, rawValue *op.KeyValue, oldValue **fileData) *fileData {
 				out := &fileData{
-					FileKey:   file.FileKey,
-					FileState: file.State,
-					Retryable: file.Retryable,
-					IsEmpty:   file.StagingStorage.IsEmpty,
+					FileKey: file.FileKey,
+					State:   file.State,
+					Retry:   file.Retryable,
+					IsEmpty: file.StagingStorage.IsEmpty,
 					File: plugin.File{
 						FileKey:  file.FileKey,
 						Provider: file.TargetStorage.Provider,
@@ -245,7 +245,7 @@ func (o *operator) checkFile(ctx context.Context, file *fileData) {
 		return
 	}
 
-	if !file.Retryable.Allowed(o.clock.Now()) {
+	if !file.Retry.Allowed(o.clock.Now()) {
 		return
 	}
 
@@ -255,7 +255,7 @@ func (o *operator) checkFile(ctx context.Context, file *fileData) {
 		return
 	}
 
-	switch file.FileState {
+	switch file.State {
 	case model.FileImporting:
 		o.importFile(ctx, file)
 	default:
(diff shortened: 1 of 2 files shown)


internal/pkg/service/stream/storage/node/coordinator/fileimport/operator.go
```diff
@@ -34,18 +37,20 @@ type operator struct {
...
+	sinks *etcdop.MirrorMap[definition.Sink, key.SinkKey, *sinkData]
 }
 
 type fileData struct {
-	FileKey model.FileKey
-	State   model.FileState
-	Retry   model.Retryable
+	model.FileKey
+	model.FileState
+	model.Retryable
```
michaljurecko, line 53: @jachym-tousek-keboola  Why did you use embedding here?

internal/pkg/service/stream/storage/node/coordinator/fileimport/operator.go
```diff
@@ -214,11 +245,17 @@ func (o *operator) checkFile(ctx context.Context, file *fileData) {
...
 	}
 
-	if !file.Retry.Allowed(o.clock.Now()) {
+	if !file.Retryable.Allowed(o.clock.Now()) {
+		return
+	}
+
+	// Skip file import if sink is deleted or disabled
+	sink, ok := o.sinks.Get(file.FileKey.SinkKey)
+	if !ok || !sink.Enabled {
```
Matovidlo, line 254: Just to be sure, wouldn't it be better to `incrementRetryTimeout` if the sink is no enabled or not existing. Doing such check when sink is disabled forever utilizes in my opinion pointlessly the CPU/MEM

internal/pkg/service/stream/storage/node/readernode/sliceupload/operator.go
```diff
@@ -167,34 +198,40 @@ func (o *operator) checkSlices(ctx context.Context, wg *sync.WaitGroup) {
...
+	}
+
+	if !slice.Retryable.Allowed(o.clock.Now()) {
 		return
 	}
 
-	if !data.Retryable.Allowed(o.clock.Now()) {
+	// Skip file import if sink is deleted or disabled
+	sink, ok := o.sinks.Get(slice.SinkKey)
+	if !ok || !sink.Enabled {
```
Matovidlo, line 222: I would move it after `volume` check. It does not need `repository` and etcd to be checked when `slice` has no volume to be read on.

internal/pkg/service/stream/storage/node/coordinator/fileimport/operator.go
michaljurecko: But `FileState` for example has not method, ... and embeding is not intended to skip the field name only.

It is now a mixture of methods from 3 types, without any obvious intention/purpose.

@Matovidlo what do you think?

internal/pkg/service/stream/storage/node/coordinator/fileimport/operator.go
michaljurecko: We can call about it if you want.

Matovidlo: I understand that the intent of `IncrementRetryAttempt` was originally to be called when `etcd` change of state is not performed (e.g entity is not changed or we could not increase the attempt which is also `etcd` operation). This is repository function that communicates with `etcd`. I would expect that this would increase the timeout **but leave entity as not processed** to load off the loops.

We can leave it as it is and make another set of benchmarks. If it is a problem we will do appropriate adjustements.

internal/pkg/service/stream/storage/node/readernode/sliceupload/operator.go
michaljurecko: It is loaded from the mirror, so it is fast.

internal/pkg/service/stream/storage/node/coordinator/fileimport/operator.go
Matovidlo: I am not fan of embedding without reason. The reason behind embedding is to "inherit" methods of the embedded structures. However when we do not need them, therefore they should be present under specific key. I would rather consolidate it both to use the structures by name instead of inheriting it, just because we are lazy to write `data.FileKey` or `file.FileKey` .
```
type base struct {
    num int
}
func (b base) describe() string {
    return fmt.Sprintf("base with num=%v", b.num)
}

type container struct {
    base
    str string
}
```
Typicall usage is to use
```
 c := container{
        base: base{
            num: 1,
        },
        str: "some name",
    }
fmt.Println("describe:", c.describe())
```
If we do not use any method of embedded structures (or structure does not contain any method at all) it is pointless.

michaljurecko (approved):

Matovidlo (approved):

PR #1986, opened 2024-09-03
fix: Minor issues from benchmarks

michaljurecko:
**Changes:**
- Improved docs.
- Removed import of the `kcp` package from production code, so in `pprof` results are no `kcp-go` goroutines.
  - The `kcp` time scheduler is started by the package import, it is stored in a global variable.
  - It had no effect, it was just weird.
- Lock name is returned even if it is not locked.

-----------


Matovidlo (approved):
LGTM

internal/pkg/service/stream/storage/level/local/encoding/pipeline.go
```diff
@@ -281,9 +281,13 @@ func (p *pipeline) WriteRecord(record recordctx.Context) error {
...
+	// Set sync timeout
+	ctx, cancel := context.WithTimeout(record.Ctx(), 20*time.Second)
+	defer cancel()
+
 	// Wait for sync and return sync error, if any
-	if err := notifier.Wait(record.Ctx()); err != nil {
-		return err
+	if err := notifier.Wait(ctx); err != nil {
+		return errors.PrefixError(err, "error when waiting for sync")
 	}
```
michaljurecko, line 291: @Matovidlo I added here a 20s timeout, to see, if the `408` errors are caused by the long sync, or not.

Matovidlo (approved):
LGTM

PR #2001, opened 2024-09-09
fix: slow/fast disk stateful set removal, benchmark adjustments

Matovidlo:
**Changes:**
- Remove 2 disks from stateful set.
- Adjust k6 script to be able to add `port` into URL
- Add node and pod affinity for benchmark deployment.

-----------

### DIFF
--- a/provisioning/stream/kubernetes/templates/benchmark/job.yaml
+++ b/provisioning/stream/kubernetes/templates/benchmark/job.yaml
@@ -8,6 +8,7 @@ metadata:
     app: stream-benchmark
 spec:
   backoffLimit: 0
+  parallelism: 4
   template:
     metadata:
       labels:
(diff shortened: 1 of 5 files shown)


provisioning/stream/kubernetes/templates/benchmark/job.yaml
```diff
@@ -19,6 +20,23 @@ spec:
...
+      affinity:
+        nodeAffinity:
+          requiredDuringSchedulingIgnoredDuringExecution:
+            nodeSelectorTerms:
+            - matchExpressions:
+              - key: kubernetes.io/hostname
+                operator: NotIn
+                values:
+                - gke-dev-keboola-gcp-us-central-main-a-c6968d17-jm8r
+                - gke-dev-keboola-gcp-us-central-main-a-c6968d17-6zmw
```
michaljurecko, line 32: Please, add a comment.

michaljurecko: Also to the `podAntiAffinity`.

michaljurecko (approved):

michaljurecko:
What will be done with this PR?
https://github.com/keboola/keboola-as-code/pull/1991

Matovidlo:
@michaljurecko I think we should merge it since we do not have proposal how to do persistent benchmarking of services. We would like to build the `k6` in CI and push it to registry. We had discussion where should this script be. It can be less universal and built within each github repository or we would like to have it in some `tools` repository, where the k6 script has connection to other services as well.

provisioning/stream/kubernetes/templates/benchmark/job.yaml
Matovidlo: Fixed and comments are added.

PR #2002, opened 2024-09-09
feat: Add adjustable payload size in k6.

Matovidlo:
**Changes:**
- Default is 1, which repesents minimal payload
- Add streamBodyRequest configuration option on HTTP source

-----------


michaljurecko (approved):
Please, update also `config-map.yml`.

PR #2004, opened 2024-09-10
fix: Wrong split of encoded chunk.

Matovidlo:
**Changes:**
- Fixed bug, missing `for` loop.
  - It the payload was bigger than `2xmaxChunkSize`, it was always written only to the 2 chunks.
  - Size of the second chunk then was over the `maxChunkSiz`e, and the next write failed on negative slice index.
- Improved variables naming

-----------


michaljurecko:
@Matovidlo it is your PR, I cannot request your review.

michaljurecko (approved):

PR #2007, opened 2024-09-11
feat: Remove recordctx.BodyString(), skip []byte -> string conv

michaljurecko:
**Changes:**
- Reduced memory allocations:
 - There is no more conversion of the request body `[]byte -> string`.
 - Removed `context.WithTimeout` from the hot path.

-----------

### DIFF
--- a/internal/pkg/service/stream/api/mapper/source_response.go
+++ b/internal/pkg/service/stream/api/mapper/source_response.go
@@ -10,6 +10,7 @@ import (
 	api "github.com/keboola/keboola-as-code/internal/pkg/service/stream/api/gen/stream"
 	"github.com/keboola/keboola-as-code/internal/pkg/service/stream/definition"
 	"github.com/keboola/keboola-as-code/internal/pkg/service/stream/definition/key"
+	"github.com/keboola/keboola-as-code/internal/pkg/service/stream/mapping/csvfmt"
 	"github.com/keboola/keboola-as-code/internal/pkg/service/stream/mapping/recordctx"
 	"github.com/keboola/keboola-as-code/internal/pkg/service/stream/mapping/table/column"
 	"github.com/keboola/keboola-as-code/internal/pkg/utils/errors"
@@ -79,9 +80,14 @@ func (m *Mapper) NewTestResultResponse(sourceKey key.SourceKey, sinks []definiti
 				return nil, err
 			}
 
+			csvValueBytes, err := csvfmt.Format(csvValue)
+			if err != nil {
+				return nil, err
+			}
+
 			row.Columns = append(row.Columns, &api.TestResultColumn{
 				Name:  c.ColumnName(),
-				Value: csvValue,
+				Value: string(csvValueBytes),
 			})
 		}
 
--- a/internal/pkg/service/stream/mapping/csvfmt/csvfmt.go
+++ b/internal/pkg/service/stream/mapping/csvfmt/csvfmt.go
@@ -1,0 +1,52 @@
+package csvfmt
+
+import (
+	"strconv"
+	"unsafe"
+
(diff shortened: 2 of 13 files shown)


internal/pkg/service/stream/storage/level/local/encoding/pipeline.go
```diff
@@ -281,12 +281,8 @@ func (p *pipeline) WriteRecord(record recordctx.Context) error {
...
 	p.acceptedWrites.Add(timestamp, 1)
 
-	// Set sync timeout
-	ctx, cancel := context.WithTimeout(record.Ctx(), 20*time.Second)
-	defer cancel()
-
 	// Wait for sync and return sync error, if any
-	if err := notifier.Wait(ctx); err != nil {
+	if err := notifier.Wait(record.Ctx()); err != nil {
 		return errors.PrefixError(err, "error when waiting for sync")
```
michaljurecko, line 286: This `ctx` is here to debug timeout issues, but there are no more timeout issues, and the request has already `30s` timeout.

It reduces count of memory allocations by `20%`:

internal/pkg/service/stream/api/mapper/source_response.go
```diff
@@ -79,9 +80,14 @@ func (m *Mapper) NewTestResultResponse(sourceKey key.SourceKey, sinks []definiti
...
+			csvValueBytes, err := csvfmt.Format(csvValue)
+			if err != nil {
+				return nil, err
+			}
+
 			row.Columns = append(row.Columns, &api.TestResultColumn{
 				Name:  c.ColumnName(),
-				Value: csvValue,
+				Value: string(csvValueBytes),
 			})
```
michaljurecko, line 91: Test sink API response.

internal/pkg/service/stream/mapping/csvfmt/csvfmt.go
```diff
@@ -0,0 +1,52 @@
...
+	case nil:
+		return nil, nil
+	default:
+		return nil, errors.Errorf("unable to cast %#v of type %T to string", v, v)
+	}
+}
+
+func strToBytes(s string) []byte {
+	return unsafe.Slice(unsafe.StringData(s), len(s))
+}
```
michaljurecko, line 52: It should be `safe` :) If we don't modify the slice - Go strings are immutable.

internal/pkg/service/stream/storage/level/local/encoding/encoder/csv/fastcsv/writer.go
```diff
@@ -87,43 +85,3 @@ func (w *writer) WriteRow(cols *[]any) error {
 
 	return nil
 }
-
-func (w *writer) toCSVValue(v any) (string, error) {
-	// Inspired by cast.ToStringE(), but without slow reflection
-	switch s := v.(type) {
-	case string:
```
michaljurecko, line 94: We need it now also in the API mapper, so I moved it as public to the `csvfmt` pkg.

internal/pkg/service/stream/storage/level/local/encoding/encoder/csv/fastcsv/writer.go
```diff
@@ -26,7 +24,7 @@ func (w *writer) WriteRow(cols *[]any) error {
 	// Write each column
 	for n, col := range *cols {
 		// Cast the value to string
-		toWrite, err := w.toCSVValue(col)
+		toWrite, err := csvfmt.Format(col)
 		if err != nil {
 			return ValueError{ColumnIndex: n, err: err}
 		}
```
michaljurecko, line 30: Body column, returns `[]byte`, and there is no more conversion to string in the `csvfmt.Format`

Matovidlo (approved):

PR #2012, opened 2024-09-12
feat/fix: Add lock to sink update API endpoint

michaljurecko:
**Changes:**
- Add lock to sink update API endpoint.
  - It can also create a new bucket/table (same as `create sink` endpoint).

-----------


Matovidlo (approved):
LGTM

PR #2014, opened 2024-09-12
feat: Add source and sink metrics

jachym-tousek-keboola:
Jira: https://keboola.atlassian.net/browse/PSGO-805

**Changes:**
- Refactored sink pipeline to return bytes written.
- Added stream metrics.

-----------

### DIFF
--- a/internal/pkg/service/common/task/node_test.go
+++ b/internal/pkg/service/common/task/node_test.go
@@ -583,7 +583,7 @@ task/123/my-receiver/my-export/some.task/%s
 					attribute.Bool("is_success", false),
 					attribute.Bool("is_application_error", true),
 					attribute.String("error", "no space right on device"),
-					attribute.String("error_type", "other"),
+					attribute.String("error_type", "insufficientStorage"),
 				},
 				Events: []tracesdk.Event{
 					{
@@ -650,7 +650,7 @@ task/123/my-receiver/my-export/some.task/%s
 								attribute.String("task_type", "some.task"),
 								attribute.Bool("is_success", false),
 								attribute.Bool("is_application_error", true),
-								attribute.String("error_type", "other"),
+								attribute.String("error_type", "insufficientStorage"),
 							),
 						},
 					},
--- a/internal/pkg/service/stream/sink/router/router.go
+++ b/internal/pkg/service/stream/sink/router/router.go
@@ -77,8 +77,8 @@ func New(d dependencies, sourceType string) (*Router, error) {
 		closed:      make(chan struct{}),
 		pipelines:   make(map[key.SinkKey]*pipelineRef),
 		meters:      &meters{
-			sourceDuration: d.Telemetry().Meter().Histogram("keboola.go.stream.source.in.duration", "Duration of source requests.", "ms"),
-			sinkDuration:   d.Telemetry().Meter().Histogram("keboola.go.stream.sink.in.duration", "Duration of source requests dispatched to sink.", "ms"),
+			sourceDuration: d.Telemetry().Meter().FloatHistogram("keboola.go.stream.source.in.duration", "Duration of source requests.", "ms"),
+			sinkDuration:   d.Telemetry().Meter().FloatHistogram("keboola.go.stream.sink.in.duration", "Duration of source requests dispatched to sink.", "ms"),
 		},
 	}
 
--- a/internal/pkg/service/stream/sink/router/router.go
+++ b/internal/pkg/service/stream/sink/router/router.go
@@ -56,6 +56,7 @@ type meters struct {
 	sourceDuration metric.Float64Histogram
 	sourceBytes    metric.Int64Histogram
 	sinkDuration   metric.Float64Histogram
+	sinkBytes      metric.Int64Histogram
 }
 
 type dependencies interface {
@@ -81,6 +82,7 @@ func New(d dependencies, sourceType string) (*Router, error) {
 			sourceDuration: d.Telemetry().Meter().FloatHistogram("keboola.go.stream.source.in.duration", "Duration of source requests.", "ms"),
 			sourceBytes:    d.Telemetry().Meter().IntHistogram("keboola.go.stream.source.in.bytes", "Source request length.", "B"),
 			sinkDuration:   d.Telemetry().Meter().FloatHistogram("keboola.go.stream.sink.in.duration", "Duration of source requests dispatched to sink.", "ms"),
+			sinkBytes:      d.Telemetry().Meter().IntHistogram("keboola.go.stream.sink.in.bytes", "Bytes written to sink.", "B"),
 		},
 	}
 
@@ -283,7 +285,7 @@ func (r *Router) SourcesCount() int {
 func (r *Router) dispatchToSink(sink *sinkData, c recordctx.Context) *SinkResult {
 	startTime := r.clock.Now()
 
-	status, _, err := r.writeRecord(sink, c)
+	status, n, err := r.writeRecord(sink, c)
 	result := &SinkResult{
 		SinkID:     sink.sinkKey.SinkID,
 		StatusCode: resultStatusCode(status, err),
@@ -308,6 +310,7 @@ func (r *Router) dispatchToSink(sink *sinkData, c recordctx.Context) *SinkResult
 	)
 	durationMs := float64(r.clock.Now().Sub(startTime)) / float64(time.Millisecond)
 	r.meters.sinkDuration.Record(finalizationCtx, durationMs, metric.WithAttributes(attrs...))
+	r.meters.sinkBytes.Record(finalizationCtx, int64(n), metric.WithAttributes(attrs...))
 
 	return result
 }
--- a/internal/pkg/service/stream/sink/router/router.go
+++ b/internal/pkg/service/stream/sink/router/router.go
@@ -283,7 +283,7 @@ func (r *Router) SourcesCount() int {
 func (r *Router) dispatchToSink(sink *sinkData, c recordctx.Context) *SinkResult {
 	startTime := r.clock.Now()
 
-	status, err := r.writeRecord(sink, c)
+	status, _, err := r.writeRecord(sink, c)
 	result := &SinkResult{
 		SinkID:     sink.sinkKey.SinkID,
 		StatusCode: resultStatusCode(status, err),
@@ -312,9 +312,9 @@ func (r *Router) dispatchToSink(sink *sinkData, c recordctx.Context) *SinkResult
 	return result
 }
 
-func (r *Router) writeRecord(sink *sinkData, c recordctx.Context) (pipeline.RecordStatus, error) {
+func (r *Router) writeRecord(sink *sinkData, c recordctx.Context) (pipeline.RecordStatus, int, error) {
 	if r.isClosed() {
-		return pipeline.RecordError, ShutdownError{}
+		return pipeline.RecordError, 0, ShutdownError{}
 	}
 	return r.pipelineRef(sink).writeRecord(c)
 }
--- a/internal/pkg/service/stream/sink/router/router.go
+++ b/internal/pkg/service/stream/sink/router/router.go
@@ -54,7 +54,8 @@ type Router struct {
 
 type meters struct {
 	sourceDuration metric.Float64Histogram
-	sinkDuration metric.Float64Histogram
+	sourceBytes    metric.Int64Histogram
+	sinkDuration   metric.Float64Histogram
 }
 
 type dependencies interface {
@@ -76,8 +77,9 @@ func New(d dependencies, sourceType string) (*Router, error) {
 		collection:  newCollection(),
 		closed:      make(chan struct{}),
 		pipelines:   make(map[key.SinkKey]*pipelineRef),
-		meters:      &meters{
+		meters: &meters{
 			sourceDuration: d.Telemetry().Meter().FloatHistogram("keboola.go.stream.source.in.duration", "Duration of source requests.", "ms"),
+			sourceBytes:    d.Telemetry().Meter().IntHistogram("keboola.go.stream.source.in.bytes", "Source request length.", "B"),
 			sinkDuration:   d.Telemetry().Meter().FloatHistogram("keboola.go.stream.sink.in.duration", "Duration of source requests dispatched to sink.", "ms"),
 		},
 	}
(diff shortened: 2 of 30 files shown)


internal/pkg/service/stream/sink/router/router.go
```diff
@@ -76,8 +77,9 @@ func New(d dependencies, sourceType string) (*Router, error) {
 		collection:  newCollection(),
 		closed:      make(chan struct{}),
 		pipelines:   make(map[key.SinkKey]*pipelineRef),
-		meters:      &meters{
+		meters: &meters{
 			sourceDuration: d.Telemetry().Meter().FloatHistogram("keboola.go.stream.source.in.duration", "Duration of source requests.", "ms"),
+			sourceBytes:    d.Telemetry().Meter().IntHistogram("keboola.go.stream.source.in.bytes", "Source request length.", "B"),
```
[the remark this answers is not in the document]
Matovidlo: I suggest to create the metric, run the service, run for example k6 test and inspect the metrics on running instance.

internal/pkg/service/common/task/node_test.go
```diff
@@ -650,7 +650,7 @@ task/123/my-receiver/my-export/some.task/%s
 								attribute.String("task_type", "some.task"),
 								attribute.Bool("is_success", false),
 								attribute.Bool("is_application_error", true),
-								attribute.String("error_type", "other"),
+								attribute.String("error_type", "insufficientStorage"),
```
Matovidlo, line 653: Why `unexpectedError` = `insufficientStorage`?

internal/pkg/service/stream/sink/router/router.go
```diff
@@ -40,23 +47,43 @@ type Router struct {
 
 	// wg waits for all writes/goroutines
 	wg sync.WaitGroup
+
+	// OTEL metrics
+	meters *meters
```
Matovidlo, line 52: Is it meters? Isn't this rather `metrics`?

internal/pkg/service/stream/sink/router/router.go
```diff
@@ -267,6 +269,9 @@ func (r *Router) DispatchToSource(sourceKey key.SourceKey, c recordctx.Context)
 	)
 	durationMs := float64(r.clock.Now().Sub(startTime)) / float64(time.Millisecond)
 	r.meters.sourceDuration.Record(finalizationCtx, durationMs, metric.WithAttributes(attrs...))
+	if bytes, err := c.BodyBytes(); err == nil {
+		r.meters.sourceBytes.Record(finalizationCtx, int64(len(bytes)), metric.WithAttributes(attrs...))
+	}
```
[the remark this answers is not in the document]
Matovidlo: Basically the `error` should be handled in higher layer. If we find some problem when reading `Request.Body()` there is simply issue in the request and we send back the user error which should be reported in the layer above. 
Then simply report the `len(bytes)` and suppress the error

internal/pkg/service/stream/storage/level/local/diskwriter/network/router/balancer/balancer.go
```diff
@@ -12,12 +12,12 @@ import (
 // If no pipeline is ready, the NoPipelineReadyError is returned.
 // The implementation must provide high throughput.
 type Balancer interface {
-	WriteRecord(c recordctx.Context, pipelines []SlicePipeline) (pipeline.RecordStatus, error)
+	WriteRecord(c recordctx.Context, pipelines []SlicePipeline) (pipeline.RecordStatus, int, error)
```
Matovidlo, line 15: I am not huge fan of more than `2` return values. Can we put the number into `RecordStatus`?

internal/pkg/service/stream/storage/level/local/encoding/encoder/csv/fastcsv/writer.go
```diff
@@ -17,28 +17,35 @@ func newWriter(out io.Writer) *writer {
...
+func (w *writer) WriteRow(cols *[]any) (int, error) {
+	// Bytes written
+	n := 0
+
 	// Reset re-used buffer
 	w.row.Reset()
 
 	// Write each column
-	for n, col := range *cols {
+	for i, col := range *cols {
```
Matovidlo, line 28:
```suggestion
	for index, col := range *cols {
```
It's 2 numbers there and it would look better.

internal/pkg/service/stream/storage/level/local/encoding/encoder/csv/fastcsv/writer.go
```diff
@@ -50,38 +57,46 @@ func (w *writer) WriteRow(cols *[]any) error {
...
 			}
 
 			toWrite = toWrite[stop:]
 
 			// Encode the special character
 			if len(toWrite) > 0 && toWrite[0] == '"' {
-				if _, err := w.row.WriteString(`""`); err != nil {
-					return err
+				if b, err := w.row.WriteString(`""`); err != nil {
+					return n, err
```
Matovidlo, line 71: Do we care about `n` when we have `error` while writing? E.g I wrote 1MB of characters but end up with error. The meter does not matter because there will be issue that it cannot fit the `row` anymore.

internal/pkg/telemetry/meter.go
```diff
@@ -5,26 +5,32 @@ import (
 )
 
 type Meter interface {
-	Counter(name, desc, unit string, opts ...metric.Int64CounterOption) metric.Int64Counter
-	UpDownCounter(name, desc, unit string, opts ...metric.Int64UpDownCounterOption) metric.Int64UpDownCounter
-	Histogram(name, desc, unit string, opts ...metric.Float64HistogramOption) metric.Float64Histogram
+	IntCounter(name, desc, unit string, opts ...metric.Int64CounterOption) metric.Int64Counter
+	IntUpDownCounter(name, desc, unit string, opts ...metric.Int64UpDownCounterOption) metric.Int64UpDownCounter
+	IntHistogram(name, desc, unit string, opts ...metric.Int64HistogramOption) metric.Int64Histogram
+	FloatHistogram(name, desc, unit string, opts ...metric.Float64HistogramOption) metric.Float64Histogram
```
Matovidlo, line 11: Why did we change the naming? I would inderstand that if we have `IntCounter` and `FloatCounter` but we have just 1.
For Histogram it makes sense.

internal/pkg/service/common/task/node_test.go
michaljurecko: Because we added support for  `errors.WithName` interface.

internal/pkg/service/stream/sink/router/router.go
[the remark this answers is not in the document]
michaljurecko: Probably forgot to mention that the histogram makes something like a separate metric for each interval.

-----------

For bytes should be `Counter` sufficient, we will know min/max/average (per project, sink, ...)

-----------

Default histogram buckets are:
```
.005, .01, .025, .05, .1, .25, .5, 1, 2.5, 5, 10 (seconds)
```

------

You should set some custom buckets for our durations - so we can measure more precisely which bucket the duration falls into:

```
WithExplicitBucketBoundaries(.005, .01, .025, .05, .1, .15, .2,.25, .3,.4,.5,.75,1, 2.5, 5, 7.5, 10,20,30),
```

internal/pkg/service/stream/sink/router/router.go
```diff
@@ -40,23 +47,43 @@ type Router struct {
...
 		plugins:     d.Plugins(),
 		definitions: d.DefinitionRepository(),
 		collection:  newCollection(),
 		closed:      make(chan struct{}),
 		pipelines:   make(map[key.SinkKey]*pipelineRef),
+		meters: &meters{
+			sourceDuration: d.Telemetry().Meter().FloatHistogram("keboola.go.stream.source.in.duration", "Duration of source requests.", "ms"),
+			sourceBytes:    d.Telemetry().Meter().IntHistogram("keboola.go.stream.source.in.bytes", "Source request length.", "B"),
+			sinkDuration:   d.Telemetry().Meter().FloatHistogram("keboola.go.stream.sink.in.duration", "Duration of source requests dispatched to sink.", "ms"),
+			sinkBytes:      d.Telemetry().Meter().IntHistogram("keboola.go.stream.sink.in.bytes", "Bytes written to sink.", "B"),
```
michaljurecko, line 85: :thinking: 

`keboola.go.stream.source.in.duration` -> `keboola.go.stream.source.{x/rec/record}.duration`

`keboola.go.stream.sink.in.duration` -> `keboola.go.stream.sink.{x/rec/record}.duration`
`keboola.go.stream.sink.in.bytes` -> `keboola.go.stream.sink.{out/encoded/uncompressed/raw}.bytes`

internal/pkg/service/stream/sink/router/router.go
```diff
@@ -229,6 +254,20 @@ func (r *Router) DispatchToSource(sourceKey key.SourceKey, c recordctx.Context)
 		result.Finalize()
 	}
 
+	// Create context for task finalization, the original context could have timed out.
+	// If release of the lock takes longer than the ttl, lease is expired anyway.
+	finalizationCtx, finalizationCancel := context.WithTimeout(context.WithoutCancel(c.Ctx()), 5*time.Second)
+	defer finalizationCancel()
```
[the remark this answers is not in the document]
michaljurecko: Sure :)

-----------

BTW, I suggest to replace it with:
```
finalizationCtx := context.WithoutCancel(c.Ctx())
```

@Matovidlo  and I recently found out that context timeout in hot path requires 20% more memory allocation.

internal/pkg/service/stream/sink/router/router.go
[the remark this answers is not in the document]
michaljurecko: I suggest to add `BodyLen()`  method without error ( returns `0` on error).

The `BodyBytes()` method is caching body AND err, so it is OK.

internal/pkg/service/stream/sink/router/router.go
```diff
@@ -249,12 +297,27 @@ func (r *Router) dispatchToSink(sink *sinkData, c recordctx.Context) *SinkResult
 		r.logger.Errorf(context.Background(), `write record error: %s`, err.Error())
 	}
 
+	// Create context for task finalization, the original context could have timed out.
+	// If release of the lock takes longer than the ttl, lease is expired anyway.
+	finalizationCtx, finalizationCancel := context.WithTimeout(context.WithoutCancel(c.Ctx()), 5*time.Second)
+	defer finalizationCancel()
+
```
michaljurecko, line 304: Same, please don't use context timeout.

internal/pkg/service/stream/storage/level/local/diskwriter/network/router/balancer/balancer.go
michaljurecko: RecordStatus here tells whether the record was `accepted` or `processed`.

I take it as a standard Write method + status, I don't mind.

internal/pkg/service/stream/storage/level/local/encoding/encoder/csv/fastcsv/writer.go
michaljurecko: On error, the buffer is discarded, either the entire row is written or nothing, so it should return `0`.

internal/pkg/telemetry/meter.go
michaljurecko: In my opinion, it is a good step for the future.

michaljurecko (approved):
LGTM

Try to think about beter metric names.

internal/pkg/service/stream/sink/router/router.go
```diff
@@ -40,23 +47,39 @@ type Router struct {
...
+		clock:       d.Clock(),
 		logger:      d.Logger().WithComponent("sink.router"),
 		plugins:     d.Plugins(),
 		definitions: d.DefinitionRepository(),
 		collection:  newCollection(),
 		closed:      make(chan struct{}),
 		pipelines:   make(map[key.SinkKey]*pipelineRef),
+		meters:      &meters{
+			sourceDuration: d.Telemetry().Meter().Histogram("keboola.go.stream.source.in.duration", "Duration of source requests.", "ms"),
+			sinkDuration:   d.Telemetry().Meter().Histogram("keboola.go.stream.sink.in.duration", "Duration of source requests dispatched to sink.", "ms"),
```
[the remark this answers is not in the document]
michaljurecko: We probably just haven't finished it, there should definitely be 2.

internal/pkg/service/stream/storage/level/local/encoding/encoder/csv/fastcsv/writer.go
Matovidlo: So basically it would be better to return `0, err`.

internal/pkg/telemetry/meter.go
Matovidlo: Then it would be good to add other methods as well, it's just 1 line of code as preparation.

internal/pkg/service/stream/sink/router/router.go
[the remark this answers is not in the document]
michaljurecko: (timeout is not expected here, metrics are collected to the memory)

internal/pkg/service/stream/sink/router/router.go
michaljurecko: OpenTelemetry is using `meter`/`meters` wording, see GO pkgs.

internal/pkg/service/stream/sink/router/router.go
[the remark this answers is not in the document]
michaljurecko: - You should remove ``context.WithTimeout` , because performance.
- It is safe to remove, timeout is not expected.

internal/pkg/service/stream/storage/level/local/diskwriter/network/router/balancer/balancer.go
michaljurecko: For me, these two solutions make sense:
- `(pipeline.RecordStatus, int, error)`
- `(pipeline.WriteResult, error)`, `pipeline.WriteResult`  is struct

michaljurecko: This is really a detail, I suggest you to call about it after Martin's vacation, and don't block the merge, it can always be adjusted.

Matovidlo: I would go with `WriteResult` seems better

internal/pkg/service/stream/sink/router/router.go
michaljurecko: Ok, but this is official OTEL Go package:
image

Of course, feel free to rename it.

Matovidlo:
@michaljurecko I do not have special check for metrics but it would be fine to have session how to check new meters/histograms before releasing to production. I used to use `prometheus` exporter docker image or `grafana` docker image and import source container to show the metrics in graph.
What do you suggest when inspecting metrics in dev environment

Matovidlo (approved):

michaljurecko:
@Matovidlo it's a standard OpenTelemetry setup, in the dev/prod environment you can set up any environment you want, I don't know what the added value of that call should be from my side (I just want to use the time for more specific things).

michaljurecko (approved):

PR #2026, opened 2024-09-18
refactor: etcdop: remove deprecated many.go

michaljurecko:
**Changes:**
- Removed deprecated `many.go` file,it is  no longer necessary/it is confusing.
  - (task from etcd documentation call)

-----------


hosekpeter (approved):
LGTM

PR #2029, opened 2024-09-19
refactor: Simplify statistics code

michaljurecko:
Related to: https://github.com/keboola/keboola-as-code/pull/1976

**Changes:**
- Reduced code complexity/length.

-----------

(maybe it's best to call about it?)

## Intro

- We have statistics stored per `<slice>/<sourceNodeId>`, values for parent objects are calculated on read, as a sum of all values from prefix.
  - So the `write` side is lock free.
- Product team requires "statistics reset" feature.
  - We don't want to delete low-level slice statistics, because they are used also internally.
  - Therefore we created `<sink>/_reset` key, it substract a snapshot from statistics values.
  - We use uint64 types, therefore the reset must be counted at the end, or we must change the type (datasize.ByteSize).
- There is also rollup slices/files of statistics to the sink (or higher levels) - when files are deleted.
  - **In my opinion, this has no effect on the reset, and the total object sum, including the reset, should enter the rollup.**
  -  Because, THERE is no `_reset` at the files/slices level now (and it doesn't make sense even in the future).
  -  If we try to make a separate rollup for `_reset`  value, the code is more complicated, but it has no additional informational value.
  - We currently have a "reset statistics" endpoint only for the sink, but there should be no difference for the source/branch/project.


internal/pkg/service/stream/storage/statistics/repository/stats_schema.go
```diff
@@ -17,7 +17,9 @@ const (
 	// rollupSumKey contains the sum of all statistics from the object children that were deleted.
 	rollupSumKey = "_sum"
 	// rollupResetKey contains the sum of all statistics from the object children that are ignored.
-	rollupResetKey = "_reset"
+	// ~ char is important, in alphabetical order, the reset will always be at the end - after the statistics to which it belongs.
+	// We are using uint64 types, so we have to prevent underflow - the reset must be subtracted at the end.
+	rollupResetKey = "~reset"
 )
```
michaljurecko, line 23: This change will simplify the code.

michaljurecko: See ASCII table: https://www.asciitable.com/

michaljurecko: https://www.editpad.org/tool/alphabetizer

image

internal/pkg/service/stream/storage/statistics/repository/stats_sum.go
```diff
@@ -1,33 +0,0 @@
...
-func SumStats(ctx context.Context, now time.Time, prefix iterator.DefinitionT[statistics.Value]) (out statistics.Value, err error) {
-	var outReset statistics.Value
-	if err := sumStatsOp(now, prefix, &out, &outReset).Do(ctx).Err(); err != nil {
-		return out, err
-	}
-	return out.Add(outReset), nil
-}
-
-// sumStatsOp sums all stats from the iterator.
-func sumStatsOp(now time.Time, prefix iterator.DefinitionT[statistics.Value], outSum *statistics.Value, outReset *statistics.Value) *iterator.ForEachT[statistics.Value] {
```
michaljurecko, line 23: SumStats was unused, and sumStatsOp was used on 2 places, but it can be simplified too.

internal/pkg/service/stream/storage/statistics/statistics.go
```diff
@@ -67,30 +67,40 @@ type Aggregated struct {
...
-// Note that the value can't be negative, in case of an underflow the fields will be 0.
-func (v Value) Add(v2 Value) Value {
-	r := v
-	if (r.ResetAt == nil) != (v2.ResetAt == nil) {
-		if r.ResetAt != nil {
-			return v2.Add(r)
-		}
-		r = r.sub(v2)
+func (v Value) With(v2 Value) Value {
+	if v.ResetAt != nil {
```
michaljurecko, line 71: `Add` -> `With` bacause it is immutable, so it is less confusing.

---------

- Added `add/sub` private methods.
- `With` method cannot be called on a reset value (to simplify code), bude can be called WITH reset value.
  - Because the `sum` container is never a reset value.

internal/pkg/service/stream/storage/statistics/repository/stats_reset.go
```diff
@@ -34,12 +35,14 @@ func (r *Repository) ResetSinkStats(sinkKey key.SinkKey) *op.AtomicOp[op.NoResul
...
-		resetSum := statistics.Value{}
-		objectSum = statistics.Value{}
-		return sumStatsOp(r.clock.Now(), objectPfx.GetAll(r.client), &objectSum, &resetSum)
+		return objectPfx.GetAll(r.client).ForEach(func(item statistics.Value, _ *iterator.Header) error {
+			if item.ResetAt == nil {
+				objectSum = objectSum.With(item)
+			}
+			return nil
+		})
 	})
```
michaljurecko, line 46: Removed `sumStatsOp`.

internal/pkg/service/stream/storage/statistics/statistics_test.go
```diff
@@ -125,45 +125,6 @@ func TestValue_With(t *testing.T) {
 				StagingSize:      1001,
 			},
 		},
-		{
-			name: "empty reset values",
-			value: statistics.Value{
-				ResetAt: ptr.Ptr(utctime.MustParse("2000-02-01T00:00:00.000Z")),
-			},
```
michaljurecko, line 132: `With`` method cannot be called on a reset value -> panic.

PR #2032, opened 2024-09-20
docs: etcd/etcdop framework

michaljurecko:
Jira: https://keboola.atlassian.net/browse/PSGO-791

**Changes:**
- Renamed `ActualAtomicOp` to `AtomicOpCtx` - request from the docs call.
- Added etcd/etcdop framework docs.

-----------

### DIFF
--- a/docs/etcd.md
+++ b/docs/etcd.md
@@ -1,0 +1,373 @@
+# etcd
+
+This document describes how and why we use the [etcd](https://etcd.io/docs/) database and the [etcdop](../internal/pkg/service/common/etcdop) - our high-level framework built on top of etcd.
+
+## Motivation
+
+### Templates API
+
+- The Templates API currently has no database for data; it is stateless.
+- Data regarding template instances is stored in the metadata of branches and configurations within the Storage API.
+- A database for locks was needed to avoid collisions.
+- Since the deployment of etcd was already planned for the Stream API, and it met the requirements, etcd was used for the Templates API as well.
+
+### Stream API
+
+- During development, it was discovered that we could not achieve the required performance with a relational database (MySQL).
+- It is not so much about the number of operations per second as about the overall **architecture of the database**.
+- A database was needed that would be a good core for the Stream API **distributed system**.
+- We use the **[Consistent Core](https://martinfowler.com/articles/patterns-of-distributed-systems/consistent-core.html)** pattern:
+  > A smaller cluster providing stronger consistency to allow the large data cluster to coordinate server activities without implementing quorum-based algorithms.
+- Thus, complex questions regarding consistency, quorum, and high availability were delegated to etcd.
+- etcd is used to store stream definitions and metadata about auxiliary files, slices, statistics, etc.
+- The data itself is not stored in etcd, but on mounted volumes.
+- etcd is also used by [Kubernetes](https://kubernetes.io/) and is therefore well-tested in production.
+
+## Features
+
+The following features of the etcd database are important to us:
+- [Data model](https://etcd.io/docs/v3.5/learning/data_model/)
+  - etcd is a [KV store](https://en.wikipedia.org/wiki/Key%E2%80%93value_database); key and value are bytes.
+  - Every change in the database increments a global counter - revision.
+  - The database [guarantees](https://etcd.io/docs/v3.5/learning/api_guarantees/) well-known properties: Atomicity, Consistency, Isolation, and Durability.
+  - It is a database focused on consistency.
+  - The maximum recommended size of all data is [8GB](https://etcd.io/docs/v3.5/dev-guide/limit/).
+- Multiple keys from a `prefix/`, or `start-end` range can be loaded using [Range Queries](https://etcd.io/docs/v3.5/learning/api/#range).
+- The database provides `if, then, else` [transactions](https://etcd.io/docs/v3.5/learning/api/#transaction).
+  - This is a different concept than transactions in relational databases.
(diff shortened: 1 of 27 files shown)


docs/etcd.md
```diff
@@ -0,0 +1,405 @@
...
+  - For example, for performance or database size optimization.
+- Possible extensions could include `gzip` compression or binary [Protocol Buffers](https://protobuf.dev) encoding.
+- It is possible to work with several encodings simultaneously:
+  - This allows for maintaining backward compatibility.
+  - Or implementing gradual migration.
+  - For example, a header/prefix can be used to determine encoding:
+    - `{` or `[` for JSON
+    - `PB:` for Protocol Buffers
+    - `GZIP:` for compression.
+  - The disadvantage is more complicated debugging, as values will not be directly readable.
```
Matovidlo, line 105: I saw once that it is possible to bind (pipe) output of database values through some decoder ( for example for protocol buffers decoder ) so the output is readable. Is this feasible for you to check how to pipe single command to for example decompress compressed values? 
Just it would be nice if you ever encountered such use case, if not just resolve the commentary :)

docs/etcd.md
```diff
@@ -0,0 +1,405 @@
...
+- Prefixes:
+  - `etcdop.Prefix` type is for raw values, represented by bytes.
+  - `etcdop.PrefixT[T]` type is for values encoded/decoded by the [serde](#serde).
+- Keys:
+  - `etcdop.Key` type is for raw values, represented by bytes.
+  - `etcdop.KeyT[T]` type is for values encoded/decoded by the [serde](#serde).
+
+Example of defining a typed prefix and key:
+```go
+s := serde.NewJSON(validator.Validate)
```
Matovidlo, line 121: Is there any specific thing about `validator.Validate`? Should we always use this setup of validator or can this be omitted?

docs/etcd.md
```diff
@@ -0,0 +1,405 @@
...
+  - Or implementing gradual migration.
+  - For example, a header/prefix can be used to determine encoding:
+    - `{` or `[` for JSON
+    - `PB:` for Protocol Buffers
+    - `GZIP:` for compression.
+  - The disadvantage is more complicated debugging, as values will not be directly readable.
+
+### Schema
+
+When using the `etcdop` framework, we must start with a schema.
```
Matovidlo, line 109: Do we use schema only to create prefix and keys under prefix? Basically we use it for `PUT` operations only? I was thinking what is the typical workflow when I have schema in place and I would like to write keys, and in similar way get the keys.
I think we typically use `repository` methods which get's the content itself, but what if we do not use repository just schema?

docs/etcd.md
```diff
@@ -0,0 +1,405 @@
...
+
+Example of defining a typed prefix and key:
+```go
+s := serde.NewJSON(validator.Validate)
+pfx1 := etcdop.NewTypedPrefix[model.Slice]("storage/slice", s)
+pfx2 := pfx1.Add("level").Add("local")
+key1 := pfx2.Key(slice1.Key())
+key2 := pfx2.Key(slice2.Key())
+require.NoError(t, key1.Put(slice1).Do(ctx).Err())
+require.NoError(t, key2.Put(slice2).Do(ctx).Err())
```
Matovidlo, line 127: I miss usage of  `GetAll` or any `get` or `listing` operation. Could you provide it somehow? :slightly_smiling_face: . When I tried to implement the storage cleaner instead of using the repository I used schema but it was unconvintient.
The documenation does not say so but we prefer to go through `repository` when accessing/manipulating with models/schemas.

docs/etcd.md
```diff
@@ -0,0 +1,405 @@
...
+  - Additionally, they don't use generic types (because they weren't available in Go at that time).
+  - Generic types can make the job a lot easier.
+
+**The `etcdop` framework provides:**
+- Saving entities in `JSON` format, as etcd itself only stores bytes - see [Serde](#serde).
+- `PrefixT[T]` and `KeyT[T]` generic types - abstractions for etcd prefix and key.
+  - `T` is the type of the entity, see [Schema](#schema).
+- A high-level wrapper for basic etcd operations, see [Basic Operations](#basic-operations).
+- Combination of basic operations into a transaction, see [TxnOp](#txnop).
+- Combination of basic operations and transactions into an atomic operation with multiple read phases, see [AtomicOp](#atomicop).
```
Matovidlo, line 69: Which of these 3 operations do I prefer to use.
1. Basic operations are similar as I would be using `etcdctl` or?
2. When do I use `TxnOp`? Only when I have intent to `merge` 2 transactions together?
3. When do I use `AtomicOp`? Do I use it only when PUT and GET operations should be done together? Could you provide some `complex` AtomicOp and TxnOp and go little bit in depth of the example? For example I want to create File only when Sink and Source exists etc. 
4. What was the flow that you come with such solution?

hosekpeter (approved):

docs/etcd.md
michaljurecko: I have no ready example.

docs/etcd.md
michaljurecko: I will add a link to the validator package.

docs/etcd.md
michaljurecko: :heavy_check_mark:

docs/etcd.md
michaljurecko: Ok, I will add `GetAll`.

Matovidlo (approved):

PR #2035, opened 2024-09-23
fix: Cluster lock timeout

jachym-tousek-keboola:
Jira: https://keboola.atlassian.net/browse/PSGO-831

### DIFF
--- a/internal/pkg/service/stream/storage/node/coordinator/clusterlock/file.go
+++ b/internal/pkg/service/stream/storage/node/coordinator/clusterlock/file.go
@@ -3,6 +3,7 @@ package clusterlock
 import (
 	"context"
 	"fmt"
+	"time"
 
 	"github.com/keboola/keboola-as-code/internal/pkg/log"
 	"github.com/keboola/keboola-as-code/internal/pkg/service/common/distlock"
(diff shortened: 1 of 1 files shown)


michaljurecko (approved):

internal/pkg/service/stream/storage/node/coordinator/clusterlock/file.go
```diff
@@ -3,21 +3,26 @@ package clusterlock
...
 
 func LockFile(ctx context.Context, locks *distlock.Provider, logger log.Logger, fileKey model.FileKey) (lock *etcdop.Mutex, unlock func(), err error) {
 	lock = locks.NewMutex(fmt.Sprintf("operator.file.%s", fileKey))
 	if err := lock.Lock(ctx); err != nil {
-		logger.Errorf(ctx, "cannot lock %q: %s", lock.Key(), err)
-		return nil, nil, err
+		return nil, nil, errors.PrefixErrorf(err, "cannot lock %q:", lock.Key())
 	}
 
+	logger.Debug(ctx, "acquired lock")
```
michaljurecko, line 21: - `lock.key` should by an attribute (also in the cannot lock)
- for now, it should be `info` (it will be a maximum of hundreds of messages per second).
- there should be also `released lock` message.

PR #2040, opened 2024-09-24
fix: Disk writer already exists issue

Matovidlo:
Happens when OOM is received on pod and new pod continues to write into slice pipeline that was not closed gracefully


Matovidlo:
@michaljurecko ok for me, approve it

michaljurecko (approved):

PR #2043, opened 2024-09-24
fix: Temporarily remove disks.

Matovidlo:
CD fails on that due to changes of disks.
This PR/commit has to be reverted

-----------


eddycek (approved):
oka, pak je potřeba udělat hned revert

PR #2048, opened 2024-09-25
fix: API for path and template columns

jachym-tousek-keboola:
The new fields were not correctly mapped to database and back to API because I didn't realize I have to adjust mapper to know about them. This is now fixed and covered by a test.


Matovidlo (approved):
LGTM

hosekpeter (approved):

PR #2052, opened 2024-09-26
feat: Add aligned writes for diskwriter.

Matovidlo:
We need to ensure by aligned write to properly truncate file if there is some newly written bytes that we do not want due to graceful/non graceful shutdown.


michaljurecko (approved):
LGTM

PR #2053, opened 2024-09-26
feat: Enable APi bg tasks cleanup in Streams

michaljurecko:
Jira: https://keboola.atlassian.net/browse/PSGO-845

**Changes:**
- In the past, `tasks` was widely used in Streams service.
- Now, they are used only for API background tasks, remaining work has been covered by operators.
- Task cleanup is no more a `task` (with lock), it was replaced by distribution group:
  - **In the old version**, the `tasks cleaner` was running on multiple API nodes in Tempaltes API.
  - But only one cleanup was running at time - because there was task lock.
  - **In the new version**, only one API nodes is responsible for the tasks cleanup (distribution group is used).

-----------

(review commit by commit)


Matovidlo (approved):
LGTM

PR #2055, opened 2024-09-26
feat: Set limit for too big row encoded

Matovidlo:
We need to make sure that the request of clients does not make buffer/memory overflow per single column encoded

### DIFF
--- a/internal/pkg/service/stream/config/config_test.go
+++ b/internal/pkg/service/stream/config/config_test.go
@@ -180,6 +180,8 @@ storage:
                     type: csv
                     # Concurrency of the format writer for the specified file type. 0 = auto = num of CPU cores. Validation rules: min=0,max=256
                     concurrency: 0
+                    # Set's the limit of single row to be encoded. Limit should be bigger than accepted request on source otherwise received message will never be encoded. Validation rules: minBytes=1kB,maxBytes=2MB
+                    limit: 1536KB
                 # Max size of the buffer before compression, if compression is enabled. 0 = disabled. Validation rules: maxBytes=16MB
                 inputBuffer: 2MB
                 # Max size of a chunk sent over the network to a disk writer node. Validation rules: required,minBytes=64kB,maxBytes=1MB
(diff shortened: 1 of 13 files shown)


internal/pkg/service/stream/storage/level/local/encoding/encoder/csv/csv_test.go
```diff
@@ -84,6 +86,33 @@ func TestCSVWriter(t *testing.T) {
...
+	_, err = csvEncoder.WriteRecord(record)
+	require.NoError(t, err)
+
+	record = recordctx.FromHTTP(
+		utctime.MustParse("2000-01-01T03:00:00.000Z").Time(),
+		&http.Request{Body: io.NopCloser(strings.NewReader("foobartoomuch"))},
+	)
+	_, err = csvEncoder.WriteRecord(record)
+	assert.Equal(t, "too big csv row to be written", err.Error())
+}
```
michaljurecko, line 114: Suggestion: `too big CSV row, limit %q`

internal/pkg/service/stream/config/config_test.go
```diff
@@ -173,6 +173,8 @@ storage:
                     type: csv
                     # Concurrency of the format writer for the specified file type. 0 = auto = num of CPU cores. Validation rules: min=0,max=256
                     concurrency: 0
+                    # Set's the limit of single row to be encoded. Limit should be bigger than accepted request on source otherwise received message will never be encoded. Validation rules: minBytes=1kB,maxBytes=2MB
+                    rowSizeLimit: 1536KB
                 # Max size of the buffer before compression, if compression is enabled. 0 = disabled. Validation rules: maxBytes=16MB
                 inputBuffer: 2MB
                 # Max size of a chunk sent over the network to a disk writer node. Validation rules: required,minBytes=64kB,maxBytes=1MB
```
michaljurecko, line 180: You should update also provisioning config-map.

michaljurecko (approved):

PR #2060, opened 2024-09-27
feat: Enable request body streaming

michaljurecko:


michaljurecko:
Update to the new version probably works.

michaljurecko:
Added missing lock.

PR #1799, opened 2024-06-03
fix: Add new linter copyloopvar to prevent copying local variables in loop

Matovidlo:
Jira: [PSGO-638](https://keboola.atlassian.net/browse/PSGO-638)

**Changes:**
- Simplifies PR reviews.

-----------

[PSGO-638]: https://keboola.atlassian.net/browse/PSGO-638?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #2065, opened 2024-09-30
build: Update linter

jachym-tousek-keboola:
Jira: https://keboola.atlassian.net/browse/PSGO-861 (part 1)

**Changes:**
- Linter now reports lossy integer conversions, I used https://github.com/ccoVeille/go-safecast to fix those.

-----------


build/ci/golangci.yml
```diff
@@ -115,7 +115,7 @@ linters:
     - errname
     - errorlint
     - exhaustive
-    - exportloopref
+    - copyloopvar
```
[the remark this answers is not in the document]
Matovidlo: https://github.com/keboola/keboola-as-code/pull/1799

internal/pkg/service/stream/source/type/httpsource/config.go
```diff
@@ -13,9 +13,9 @@ type Config struct {
...
 	MaxConnections     int               `configKey:"maxConnections" configUsage:"The maximum number of concurrent connections the server may serve." validate:"required"`
-	ReadBufferSize     datasize.ByteSize `configKey:"readBufferSize" configUsage:"Read buffer size, all HTTP headers must fit in" validate:"required"`
-	WriteBufferSize    datasize.ByteSize `configKey:"writeBufferSize" configUsage:"Write buffer size." validate:"required"`
-	MaxRequestBodySize datasize.ByteSize `configKey:"maxRequestBodySize" configUsage:"Max size of the HTTP request body." validate:"required"`
+	ReadBufferSize     datasize.ByteSize `configKey:"readBufferSize" configUsage:"Read buffer size, all HTTP headers must fit in" validate:"required,minBytes=1kB,maxBytes=1MB"`
+	WriteBufferSize    datasize.ByteSize `configKey:"writeBufferSize" configUsage:"Write buffer size." validate:"required,minBytes=1kB,maxBytes=1MB"`
+	MaxRequestBodySize datasize.ByteSize `configKey:"maxRequestBodySize" configUsage:"Max size of the HTTP request body." validate:"required,minBytes=100B,maxBytes=10MB"`
```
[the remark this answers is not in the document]
Matovidlo: It seems reasonable.  Maybe the request body is too big. 10MB request is not common. Also fragmentation will be used by transport protocol. Consider changing `maxRequestBodySize` up to `4MB`

internal/pkg/service/stream/storage/level/local/encoding/size/size.go
```diff
@@ -102,7 +102,9 @@ func NewMeterWithBackup(ctx context.Context, clk clock.Clock, logger log.Logger,
 func (m *Meter) Write(p []byte) (int, error) {
 	n, err := m.w.Write(p)
 	m.lock.Lock()
-	m.size += datasize.ByteSize(n)
+	if n > 0 {
+		m.size += datasize.ByteSize(n)
+	}
```
Matovidlo, line 107: Why do we need 1 if instruction ? 45 + 0 = 45, or?

internal/pkg/telemetry/fortest.go
```diff
@@ -32,8 +33,8 @@ const (
 
 type ForTest interface {
 	Telemetry
-	TraceID(n int) trace.TraceID
-	SpanID(n int) trace.SpanID
+	TraceID(n uint16) trace.TraceID
+	SpanID(n uint16) trace.SpanID
```
Matovidlo, line 37: I don't like 16 bit width variables. What happens if overflows? Returns error and Span, Trace will not be generated?

scripts/tools.sh
```diff
@@ -43,7 +43,7 @@ then
   # Disabled due to issues with windows curl command not working
   # Related to https://github.com/keboola/keboola-as-code/pull/1818/commits/df23a4e2855557a01460593f7d6ba59b5dc8825c
   # curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b "$GOBIN" v1.58.0
-  go install github.com/golangci/golangci-lint/cmd/golangci-lint@v1.57.2
+  go install github.com/golangci/golangci-lint/cmd/golangci-lint@latest
```
Matovidlo, line 46: Can we put there `v1.61.0`? When they do update our CI will return error. We should do the update once a month, otherwise we will do once a month PR for update linter due to CI fails.

internal/pkg/telemetry/fortest.go
```diff
@@ -307,13 +308,21 @@ func (g *testIDGenerator) Reset() {
 
 func (g *testIDGenerator) NewIDs(ctx context.Context) (trace.TraceID, trace.SpanID) {
 	v := g.traceID.Add(1)
-	traceID := toTraceID(testTraceIDBase + uint16(v))
+	i, err := safecast.ToUint16(v)
+	if err != nil {
+		panic(err)
```
Matovidlo, line 313: This seems quite aggresive. Is it just only for tests or also in production?

internal/pkg/service/stream/storage/level/local/encoding/size/size.go
Matovidlo:
```
Write writes len(p) bytes from p to the underlying data stream. It returns the number of bytes written from p (0 <= n <= len(p)) and any error encountered that caused the write to stop early. Write must return a non-nil error if it returns n < len(p). Write must not modify the slice data, even temporarily.
```
Therefore there cannot be negative number of written bytes, only error.

internal/pkg/telemetry/fortest.go
Matovidlo: I need more time for this

Matovidlo: https://github.com/keboola/keboola-as-code/pull/2065/commits/c7e95b3c795433ace4dbf0ff970ad3090fbad58e

hosekpeter (approved):

internal/pkg/service/stream/storage/level/local/encoding/size/size.go
Matovidlo: https://github.com/keboola/keboola-as-code/pull/2065/commits/a95b10318f90f9d4d3fd22ad5ec7d71d9d458cdb

PR #2069, opened 2024-09-30
build: Update to Go 1.23

jachym-tousek-keboola:
Jira: [keboola.atlassian.net/browse/PSGO-861](https://keboola.atlassian.net/browse/PSGO-861) (part 3)

**Changes:**
- Update Go to 1.23.1 + related library updates

-----------

Closes #1995


scripts/tools.sh
```diff
@@ -61,7 +61,7 @@ fi
 # air - code changes watcher
 if ! command -v air &> /dev/null
 then
-  go install github.com/air-verse/air@v1.52.3
+  go install github.com/air-verse/air@latest
```
Matovidlo, line 64: Please stabilize it to particular version, so we do not encounter any issues when not updating frequently

Matovidlo (approved):
LGTM, just change to particular version. `latest` is not good deafult due to possible CI issues when downloading `latest` versions of software. Same applies for linter for example

hosekpeter (approved):

PR #2076, opened 2024-10-03
fix: ReleaseBuffers issues

jachym-tousek-keboola:
**Changes:**
- Fix for #2058.

-----------

### DIFF
--- a/internal/pkg/service/stream/mapping/recordctx/fasthttp.go
+++ b/internal/pkg/service/stream/mapping/recordctx/fasthttp.go
@@ -83,7 +83,7 @@ func (c *fastHTTPContext) HeadersMap() *orderedmap.OrderedMap {
 }
 
 func (c *fastHTTPContext) ReleaseBuffers() {
-	c.req.ResetBody()
+	c.req.Request.ResetBody()
 	c.headersMap = nil
 	c.bodyMap = nil
 	c.jsonValue = nil
--- a/internal/pkg/service/stream/mapping/recordctx/fasthttp.go
+++ b/internal/pkg/service/stream/mapping/recordctx/fasthttp.go
@@ -19,7 +19,7 @@ import (
 type fastHTTPContext struct {
 	ctx           context.Context
 	timestamp     time.Time
-	req           *fasthttp.RequestCtx
+	reqCtx        *fasthttp.RequestCtx
 	lock          sync.Mutex
 	clientIP      net.IP
 	headersMap    *orderedmap.OrderedMap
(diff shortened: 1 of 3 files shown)


internal/pkg/service/stream/mapping/recordctx/fasthttp.go
```diff
@@ -83,7 +83,7 @@ func (c *fastHTTPContext) HeadersMap() *orderedmap.OrderedMap {
 }
 
 func (c *fastHTTPContext) ReleaseBuffers() {
-	c.req.ResetBody()
+	c.req.Request.ResetBody()
```
jachym-tousek-keboola, line 86: Turns out that `c.req.ResetBody()` actually calls `ctx.Response.ResetBody()`. We want to reset the request body here.

internal/pkg/service/stream/mapping/recordctx/http.go
```diff
@@ -84,6 +84,7 @@ func (c *httpContext) HeadersMap() *orderedmap.OrderedMap {
 }
 
 func (c *httpContext) ReleaseBuffers() {
+	_ = c.req.Body.Close()
```
jachym-tousek-keboola, line 87: This is more of a guess but shouldn't we close the request body `io.ReadCloser` here? We're not planning to read from it again. Fast http internally does this (and ignores the err).

internal/pkg/service/stream/mapping/recordctx/fasthttp.go
Matovidlo: https://github.com/keboola/keboola-as-code/pull/2058/files#r1778147284 I am not sure. Internally this calls the `c.req.Request.ResetBody()` but only when `keepBodyBuffer` is on. Are we sure that this eradicate the issue? I think with this approach we will perform another benchmark and conclude that it does not need to help

internal/pkg/service/stream/mapping/recordctx/http.go
Matovidlo: I don't think so if you want to reuse the buffers. It puts the body into the pool. If you do this you will reuse closed body and need to recreate the req.Body.
We can try it but I think this is not good approach. Could you test it locally with k6? You should have memory leak in the stream with previous version and some better performance .
https://github.com/keboola/keboola-as-code/blob/main/docs/stream/benchmarks.md

internal/pkg/service/stream/mapping/recordctx/fasthttp.go
jachym-tousek-keboola: > Internally this calls the `c.req.Request.ResetBody()`

It doesn't. You're in Response object here.

> Are we sure that this eradicate the issue? 

No. I don't think the issue is related to #2058 in the first place.

internal/pkg/service/stream/mapping/recordctx/http.go
jachym-tousek-keboola: I think the comment is unrelated to this line? `net/http` doesn't seem to be using any pool in the first place, only fasthttp does.

Matovidlo (approved):
LGTM

PR #2090, opened 2024-10-11
feat: Sink version detail endpoint

hosekpeter:
Jira: [PSGO-779](https://keboola.atlassian.net/browse/PSGO-779)

**Changes:**
- added a new endpoint
- added E2E tests

-----------


internal/pkg/service/stream/api/service/sink.go
```diff
@@ -395,6 +395,19 @@ func (s *service) ListSinkVersions(ctx context.Context, scope dependencies.SinkR
...
 
+func (s *service) SinkVersionDetail(ctx context.Context, scope dependencies.SinkRequestScope, payload *api.SinkVersionDetailPayload) (res *api.Version, err error) {
+	if err := s.sinkMustExist(ctx, scope.SinkKey()); err != nil {
+		return nil, err
+	}
+
+	sink, err := s.definition.Sink().Version(scope.SinkKey(), payload.VersionNumber).Do(ctx).ResultOrErr()
+	if err != nil {
+		return nil, err
+	}
```
Matovidlo, line 406:
```suggestion
	sinkKey := scope.SinkKey()
	if err := s.sinkMustExist(ctx, sinkKey); err != nil {
		return nil, err
	}

	sink, err := s.definition.Sink().Version(sinkKey, payload.VersionNumber).Do(ctx).ResultOrErr()
	if err != nil {
		return nil, err
	}
```
This simplifies the access to key. It can be called frequently. Maybe adjust also code in previous PRs with source and sink?

hosekpeter: we are using it in all cases, I see no reason why I should change it. :)

jachym-tousek-keboola: agreed, I don't think this is necessary

jachym-tousek-keboola (changes requested):
Let's finish the previous PR first. This one has conflicts currently and isn't reviewable anyway.

jachym-tousek-keboola (approved):

Matovidlo (approved):
LGTM

PR #2099, opened 2024-10-15
build: Group dependabot changes to one PR

jachym-tousek-keboola:
Another week and dependabot opened five more PRs... How about we just have it open one instead?

This was a feature dependabot users were asking for for years. Luckily for us it was finally implemented last year.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -9,3 +9,6 @@ updates:
     directory: "/" # Location of package manifests
     schedule:
       interval: "weekly"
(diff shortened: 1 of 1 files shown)


Matovidlo (approved):
LGTM perfect!

Matovidlo:
I don't know why it does works as default settings

jachym-tousek-keboola:
Because that's how dependabot was first implemented and changing the defaults would be bc break. :D

PR #2101, opened 2024-10-16
fix: Stabilize test of WriteRecord

Matovidlo:
Jira: [PSGO-755](https://keboola.atlassian.net/browse/PSGO-755)

**Changes:**
- The notifier is taken on encoder level, as we want to get notifier before writing actual record.
- In previous code we took notifier after WriteRecord. If we move notifier before WriteRecord within encoding pipeline, we change behaviour and add unwanted entries to be synced together.

-----------

### DIFF
--- a/internal/pkg/service/stream/storage/level/local/encoding/encoder/csv/csv.go
+++ b/internal/pkg/service/stream/storage/level/local/encoding/encoder/csv/csv.go
@@ -11,6 +11,8 @@ import (
 	"github.com/keboola/keboola-as-code/internal/pkg/service/stream/mapping/table"
 	"github.com/keboola/keboola-as-code/internal/pkg/service/stream/mapping/table/column"
 	"github.com/keboola/keboola-as-code/internal/pkg/service/stream/storage/level/local/encoding/encoder/csv/fastcsv"
+	"github.com/keboola/keboola-as-code/internal/pkg/service/stream/storage/level/local/encoding/encoder/result"
+	"github.com/keboola/keboola-as-code/internal/pkg/service/stream/storage/level/local/encoding/writesync/notify"
 	"github.com/keboola/keboola-as-code/internal/pkg/utils/errors"
 )
 
@@ -18,6 +20,7 @@ type Encoder struct {
 	columns     column.Columns
 	writersPool *fastcsv.WritersPool
 	valuesPool  *sync.Pool
+	notifier    func() *notify.Notifier
 }
 
 var columnRenderer = column.NewRenderer() //nolint:gochecknoglobals // contains Jsonnet VMs sync.Pool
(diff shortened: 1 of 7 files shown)


internal/pkg/service/stream/storage/level/local/encoding/encoder/csv/csv.go
```diff
@@ -11,13 +11,16 @@ import (
...
+	"github.com/keboola/keboola-as-code/internal/pkg/service/stream/storage/level/local/encoding/encoder/result"
+	"github.com/keboola/keboola-as-code/internal/pkg/service/stream/storage/level/local/encoding/writesync/notify"
 	"github.com/keboola/keboola-as-code/internal/pkg/utils/errors"
 )
 
 type Encoder struct {
 	columns     column.Columns
 	writersPool *fastcsv.WritersPool
 	valuesPool  *sync.Pool
+	notifier    func() *notify.Notifier
```
jachym-tousek-keboola, line 23: Why is this (and many other parameters in the code) using type `func() *notify.Notifier` instead of just `*notify.Notifier`?

Matovidlo: If you pass the notifier as It was in my previous version on local, it would not be able to receive new notifier when `TriggerSync` was called. The function causes read lock to obtain notifier if it has changed based on previous `TriggerSync`.
I will use test sample `TestEncodingPipeline_Sync_Wait_ToDisk` 
```
write foo1
write foo1
wait for writting till it is completed (use notifier#1 from constructor)
triggersync (create new notifier#2)
```
but in next case
```
write foo2
wait for writing till it is completed (*use notifier#1 from constructor*)
trigger sync (create new notifier#3)
```
Therefore the code wouldn't work.
In new code we have this
```
write foo1
write foo1
wait for writting till it is completed (use notifier#1 from constructor)
triggersync (create new notifier#2)
```
next case
```
write foo2
wait for writing till it is completed (*use new notifier from triggersync*)
trigger sync (create new notifier#3)
```

internal/pkg/service/stream/storage/level/local/encoding/pipeline_test.go
```diff
@@ -175,6 +177,8 @@ func TestEncodingPipeline_Sync_Wait_ToDisk(t *testing.T) {
 		tc.Logger.Infof(ctx, "TEST: write unblocked")
 	}()
 	tc.ExpectWritesCount(t, 2)
+	// Pri triger sync sa vytvori novy notifier, stary sa uzavrie a odblokuje 2 writy.
+	// Novy notifier sa musi predat do encoderu aby sa zablokoval dalsi zapis
```
jachym-tousek-keboola, line 181: English please.

jachym-tousek-keboola (changes requested):

internal/pkg/service/stream/storage/level/local/encoding/pipeline_test.go
Matovidlo: ah sorry, did not checked files before. My fault

jachym-tousek-keboola (approved):

hosekpeter (approved):

PR #2102, opened 2024-10-16
feat: List deleted sources

hosekpeter:
Jira: [PSGO-915](https://keboola.atlassian.net/browse/PSGO-915)

**Changes:**
- added a new endpoint
- added tests

-----------

[PSGO-915]: https://keboola.atlassian.net/browse/PSGO-915?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


api/stream/design.go
```diff
@@ -224,6 +224,20 @@ var _ = Service("stream", func() {
 		})
 	})
 
+	Method("ListDeletedSources", func() {
```
Matovidlo, line 227: Did not we want just same GET endpoint as `listSources` but filter on `Param(isDeleted)` level? So if `isDeleted` is provided, it filters out sources based on deleted flag, otherwise is lists regularly.

jachym-tousek-keboola: No, for deleted separate endpoint is okay in my opinion. I'd use filters for enabled / disabled.

PR #2106, opened 2024-10-18
fix: Do not obtain notifier before writeRow

Matovidlo:
Jira: [PSGO-755](https://keboola.atlassian.net/browse/PSGO-755)

**Changes:**
- It causes incorrect behaviour of adding writeRow to previous notifier instead of the new one, which causes guarantees of writing to be not aligned.

-----------

[PSGO-755]: https://keboola.atlassian.net/browse/PSGO-755?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/internal/pkg/service/stream/storage/level/local/encoding/writesync/sync.go
+++ b/internal/pkg/service/stream/storage/level/local/encoding/writesync/sync.go
@@ -135,7 +135,9 @@ func (s *Syncer) Notifier() *notify.Notifier {
 		return nil
 	}
 
+	ctx := context.Background()
 	s.notifierLock.RLock()
+	s.logger.Debug(ctx, "notifier obtained")
 	notifier := s.notifier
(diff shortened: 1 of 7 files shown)


internal/pkg/service/stream/storage/level/local/encoding/writesync/sync.go
```diff
@@ -135,7 +135,9 @@ func (s *Syncer) Notifier() *notify.Notifier {
 		return nil
 	}
 
+	ctx := context.Background()
```
enumag, line 138: Should add ctx parameter instead.

jachym-tousek-keboola (approved):

hosekpeter (approved):
LGTM

PR #2115, opened 2024-10-22
fix: Dependabot grouping

jachym-tousek-keboola:
Looks like dependabot created several PRs again meaning my configuration doesn't work correctly. Let's try another one.


hosekpeter (approved):
LGTM

Matovidlo (approved):
I would do only dependa bot, looks like there are new objects in telemetry due to increasing version

PR #2121, opened 2024-10-23
fix: Dependabot branches not triggering CI

jachym-tousek-keboola:
Okay one more thing. The dependabot PRs are not triggering workflows. I thought it was a temporary GitHub Actions issue at first but it's actually because the branches contain a `/` character.

https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#filter-pattern-cheat-sheet

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -4,7 +4,7 @@ on:
   push:
     # Avoid tags
     branches:
-      - "*"
+      - "**"
 
 jobs:
   check-license:


Matovidlo (approved):
LGTM. We had discussion on that previously (with Michal) but we did not make this change that you suggest before. I think it is quite improvement for dependabot PRs :)

hosekpeter (approved):

PR #2123, opened 2024-10-23
feat: Implement storage job limiter

Matovidlo:
Jira: XXXXX

**Changes:**
- ...

-----------

### DIFF
--- a/internal/pkg/service/stream/sink/type/tablesink/keboola/bridge/bridge.go
+++ b/internal/pkg/service/stream/sink/type/tablesink/keboola/bridge/bridge.go
@@ -9,6 +9,7 @@ import (
 	"github.com/keboola/keboola-as-code/internal/pkg/service/common/distlock"
 	"github.com/keboola/keboola-as-code/internal/pkg/service/common/etcdop/serde"
 	"github.com/keboola/keboola-as-code/internal/pkg/service/stream/definition"
+	definitionRepo "github.com/keboola/keboola-as-code/internal/pkg/service/stream/definition/repository"
 	"github.com/keboola/keboola-as-code/internal/pkg/service/stream/plugin"
 	keboolasink "github.com/keboola/keboola-as-code/internal/pkg/service/stream/sink/type/tablesink/keboola"
 	"github.com/keboola/keboola-as-code/internal/pkg/service/stream/sink/type/tablesink/keboola/bridge/schema"
@@ -31,15 +32,16 @@ const (
 )
 
 type Bridge struct {
-	logger            log.Logger
-	config            keboolasink.Config
-	client            etcd.KV
-	schema            schema.Schema
-	plugins           *plugin.Plugins
-	publicAPI         *keboola.PublicAPI
-	apiProvider       apiProvider
-	storageRepository *storageRepo.Repository
-	locks             *distlock.Provider
+	logger               log.Logger
+	config               keboolasink.Config
+	client               etcd.KV
+	schema               schema.Schema
+	plugins              *plugin.Plugins
+	publicAPI            *keboola.PublicAPI
+	apiProvider          apiProvider
+	definitionRepository *definitionRepo.Repository
+	storageRepository    *storageRepo.Repository
+	locks                *distlock.Provider
 
 	getBucketOnce    *singleflight.Group
 	createBucketOnce *singleflight.Group
@@ -51,23 +53,25 @@ type dependencies interface {
 	EtcdSerde() *serde.Serde
 	Plugins() *plugin.Plugins
 	KeboolaPublicAPI() *keboola.PublicAPI
+	DefinitionRepository() *definitionRepo.Repository
 	StorageRepository() *storageRepo.Repository
 	DistributedLockProvider() *distlock.Provider
 }
 
 func New(d dependencies, apiProvider apiProvider, config keboolasink.Config) *Bridge {
 	b := &Bridge{
-		logger:            d.Logger().WithComponent("keboola.bridge"),
-		config:            config,
-		client:            d.EtcdClient(),
-		schema:            schema.New(d.EtcdSerde()),
-		plugins:           d.Plugins(),
-		publicAPI:         d.KeboolaPublicAPI(),
-		apiProvider:       apiProvider,
-		storageRepository: d.StorageRepository(),
-		locks:             d.DistributedLockProvider(),
-		getBucketOnce:     &singleflight.Group{},
-		createBucketOnce:  &singleflight.Group{},
+		logger:               d.Logger().WithComponent("keboola.bridge"),
+		config:               config,
+		client:               d.EtcdClient(),
+		schema:               schema.New(d.EtcdSerde()),
+		plugins:              d.Plugins(),
+		publicAPI:            d.KeboolaPublicAPI(),
+		apiProvider:          apiProvider,
+		definitionRepository: d.DefinitionRepository(),
+		storageRepository:    d.StorageRepository(),
+		locks:                d.DistributedLockProvider(),
+		getBucketOnce:        &singleflight.Group{},
+		createBucketOnce:     &singleflight.Group{},
 	}
 
 	b.setupOnFileOpen()
--- a/internal/pkg/service/stream/sink/type/tablesink/keboola/bridge/file.go
+++ b/internal/pkg/service/stream/sink/type/tablesink/keboola/bridge/file.go
@@ -61,8 +61,9 @@ func (b *Bridge) setupOnFileOpen() {
 
 			// Update file entity
 			file.Mapping = sink.Table.Mapping
-			file.StagingStorage.Provider = stagingFileProvider // staging file is provided by the Keboola
-			file.TargetStorage.Provider = targetProvider       // destination is a Keboola table
+			file.StagingStorage.Provider = stagingFileProvider      // staging file is provided by the Keboola
+			file.TargetStorage.Provider = targetProvider            // destination is a Keboola table
+			file.TargetStorage.Import.SinkLimit = b.config.JobLimit // specific sink condition for keboola provider
 			file.StagingStorage.Expiration = utctime.From(keboolaFile.UploadCredentials.CredentialsExpiration())
 		}
 
@@ -203,15 +204,18 @@ func (b *Bridge) importFile(ctx context.Context, file plugin.File, stats statist
 			return err
 		}
 
-		// Save job ID to etcd
-		keboolaFile.StorageJobID = &job.ID
-
-		b.logger.With(attribute.String("job.id", keboolaFile.StorageJobID.String())).Infof(ctx, "created new storage job for file")
-
 		err = b.schema.File().ForFile(file.FileKey).Put(b.client, keboolaFile).Do(ctx).Err()
 		if err != nil {
 			return err
 		}
+
+		// Save job ID to etcd
+		err = b.createJob(ctx, start, file, job)
+		if err != nil {
+			return err
+		}
+
+		b.logger.Info(ctx, "created staging file")
 	}
 
 	// Wait for job to complete
--- a/internal/pkg/service/stream/sink/type/tablesink/keboola/bridge/job.go
+++ b/internal/pkg/service/stream/sink/type/tablesink/keboola/bridge/job.go
@@ -1,0 +1,39 @@
+package bridge
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	"github.com/keboola/go-client/pkg/keboola"
+	"github.com/keboola/keboola-as-code/internal/pkg/service/stream/definition"
+	"github.com/keboola/keboola-as-code/internal/pkg/service/stream/definition/key"
+	"github.com/keboola/keboola-as-code/internal/pkg/service/stream/plugin"
+)
+
+func (b *Bridge) createJob(ctx context.Context, start time.Time, file plugin.File, storageJob *keboola.StorageJob) error {
+	job := definition.Job{
+		JobKey: key.JobKey{
+			SinkKey: file.SinkKey,
+			JobID:   key.JobID(storageJob.ID.String()),
+		},
+		StorageJobID: storageJob.ID,
+	}
+	lock := b.locks.NewMutex(fmt.Sprintf("api.source.sink.jobs.%s", file.SinkKey))
+	if err := lock.Lock(ctx); err != nil {
+		return err
+	}
+	defer func() {
+		if err := lock.Unlock(ctx); err != nil {
+			b.logger.Warnf(ctx, "cannot unlock lock %q: %s", lock.Key(), err)
+		}
+	}()
+
+	op := b.definitionRepository.Job().Create(start, &job)
+	op = op.RequireLock(lock)
+	if err := op.Do(ctx).Err(); err != nil {
+		return err
+	}
+
+	return nil
+}
--- a/internal/pkg/service/stream/sink/type/tablesink/keboola/config.go
+++ b/internal/pkg/service/stream/sink/type/tablesink/keboola/config.go
@@ -8,10 +8,12 @@ import (
 type Config struct {
 	// EventSendTimeout is a timeout to perform slice upload event or file import event.
 	EventSendTimeout time.Duration `configKey:"eventSendTimeout" configUsage:"Timeout to perform upload send event of slice or import event of file"`
+	JobLimit         int           `configKey:"jobLimit" configUsage:"Number of import jobs running in keboola"`
 }
 
 func NewConfig() Config {
 	return Config{
 		EventSendTimeout: 30 * time.Second,
+		JobLimit:         10,
 	}
 }
--- a/internal/pkg/service/stream/storage/level/target/config/config.go
+++ b/internal/pkg/service/stream/storage/level/target/config/config.go
@@ -52,6 +52,7 @@ type OperatorConfig struct {
 
 // ImportConfig configures the file import.
 type ImportConfig struct {
+	SinkLimit   int               `json:"sinkLimit" configKey:"sinkLimit" configUsage:"Specifies limit of sink that is representable by number"`
 	MinInterval duration.Duration `json:"minInterval" configKey:"minInterval" configUsage:"Min duration from the last import to trigger the next, takes precedence over other settings." validate:"required,minDuration=30s,maxDuration=24h"`
 	Trigger     ImportTrigger     `json:"trigger" configKey:"trigger"`
 }
--- a/internal/pkg/service/stream/storage/node/coordinator/filerotation/conditions.go
+++ b/internal/pkg/service/stream/storage/node/coordinator/filerotation/conditions.go
@@ -21,6 +21,7 @@ const (
 	recordCountThreshold
 	sizeThreshold
 	timeThreshold
+	sinkConditionUnsatisfied
 )
 
 func newConditionResult(result int, message string) fileRotationConditionResult {
@@ -31,7 +32,7 @@ func newConditionResult(result int, message string) fileRotationConditionResult 
 }
 
 func (c fileRotationConditionResult) ShouldImport() bool {
-	return c.result != noConditionMet
+	return c.result != noConditionMet && c.result != sinkConditionUnsatisfied
 }
 
 func (c fileRotationConditionResult) Cause() string {
@@ -52,12 +53,14 @@ func (c fileRotationConditionResult) String() string {
 		return "size"
 	case timeThreshold:
 		return "time"
+	case sinkConditionUnsatisfied:
+		return "sinkConditionUnsatisfied"
 	default:
 		return "unknown"
 	}
 }
 
-func shouldImport(cfg targetConfig.ImportConfig, now, openedAt, expiration time.Time, stats statistics.Value) fileRotationConditionResult {
+func shouldImport(cfg targetConfig.ImportConfig, now, openedAt, expiration time.Time, stats statistics.Value, sinkLimit int) fileRotationConditionResult {
 	sinceOpened := now.Sub(openedAt).Truncate(time.Second)
 	if threshold := cfg.MinInterval.Duration(); sinceOpened < threshold {
 		// Min interval settings take precedence over other settings.
@@ -87,6 +90,10 @@ func shouldImport(cfg targetConfig.ImportConfig, now, openedAt, expiration time.
 
 	if threshold := cfg.Trigger.Interval.Duration(); sinceOpened >= threshold {
 		return newConditionResult(timeThreshold, fmt.Sprintf("time threshold met, opened at: %s, passed: %s threshold: %s", openedAt.Format(utctime.TimeFormat), sinceOpened.String(), threshold.String()))
+	}
+
+	if threshold := cfg.SinkLimit; threshold > 0 && sinkLimit >= threshold {
+		return newConditionResult(sinkConditionUnsatisfied, "sink conditions were not satisfied")
 	}
 
 	return newConditionResult(noConditionMet, "no condition met")
--- a/internal/pkg/service/stream/definition/job_base.go
+++ b/internal/pkg/service/stream/definition/job_base.go
@@ -1,0 +1,13 @@
+package definition
+
+import (
+	"github.com/keboola/go-client/pkg/keboola"
+	"github.com/keboola/keboola-as-code/internal/pkg/service/stream/definition/key"
+)
+
+// Job contains all Keboola-specific data we need for polling jobs.
+type Job struct {
+	key.JobKey
+	SoftDeletable
+	StorageJobID keboola.StorageJobID
+}
--- a/internal/pkg/service/stream/definition/key/job.go
+++ b/internal/pkg/service/stream/definition/key/job.go
@@ -1,0 +1,30 @@
+package key
+
+import (
+	"github.com/keboola/keboola-as-code/internal/pkg/utils/errors"
+	"go.opentelemetry.io/otel/attribute"
+)
+
+type JobID string
+
+type JobKey struct {
+	SinkKey
+	JobID JobID `json:"jobId" validate:"required,min=1,max=48"`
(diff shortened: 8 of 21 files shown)


internal/pkg/service/stream/sink/type/tablesink/keboola/bridge/file.go
```diff
@@ -61,8 +61,9 @@ func (b *Bridge) setupOnFileOpen() {
 
 			// Update file entity
 			file.Mapping = sink.Table.Mapping
-			file.StagingStorage.Provider = stagingFileProvider // staging file is provided by the Keboola
-			file.TargetStorage.Provider = targetProvider       // destination is a Keboola table
+			file.StagingStorage.Provider = stagingFileProvider      // staging file is provided by the Keboola
+			file.TargetStorage.Provider = targetProvider            // destination is a Keboola table
+			file.TargetStorage.Import.SinkLimit = b.config.JobLimit // specific sink condition for keboola provider
```
Matovidlo, line 66: This is moment, when bridge `jobLimit` is injected into file `sinkLimit`. For another provider it would be done in similar way, but it will be new configuration name.

internal/pkg/service/stream/plugin/collection.go
```diff
@@ -207,6 +207,15 @@ func (c *Collection) OnSinkModification(fn onSinkSaveFn) {
 	})
 }
 
+func (c *Collection) OnSinkDelete(fn onSinkSaveFn) {
```
Matovidlo, line 210: New plugin

internal/pkg/service/stream/sink/type/tablesink/keboola/bridge/file.go
```diff
@@ -203,15 +204,18 @@ func (b *Bridge) importFile(ctx context.Context, file plugin.File, stats statist
 			return err
 		}
 
-		// Save job ID to etcd
-		keboolaFile.StorageJobID = &job.ID
```
jachym-tousek-keboola, line 207: If I understand correctly you're tracking jobs as separate etcd records now. Do we still need this field on File then? If yes, where is it assigned now?

internal/pkg/service/stream/sink/type/tablesink/keboola/bridge/job.go
```diff
@@ -0,0 +1,72 @@
...
+}
+
+func (b *Bridge) createJob(ctx context.Context, token string, file plugin.File, storageJob *keboola.StorageJob) error {
+	modelJob := model.Job{
+		JobKey: key.JobKey{
+			SinkKey: file.SinkKey,
+			JobID:   key.JobID(storageJob.ID.String()),
+		},
+	}
+	b.logger.Debugf(ctx, "creating job: %v", modelJob)
```
jachym-tousek-keboola, line 40: Shouldn't the job be an attribute rather than part of the message?

internal/pkg/service/stream/sink/type/tablesink/keboola/bridge/job.go
```diff
@@ -0,0 +1,72 @@
...
+
+	keboolaJob := keboolasink.Job{
+		JobKey:        modelJob.JobKey,
+		StorageJobKey: storageJob.StorageJobKey,
+		Token:         token,
+	}
+	b.logger.Infof(ctx, "creating storage job: %v", keboolaJob)
```
jachym-tousek-keboola, line 47: Shouldn't the job be an attribute rather than part of the message?

internal/pkg/service/stream/sink/type/tablesink/keboola/bridge/job.go
```diff
@@ -0,0 +1,72 @@
...
+
+	lock := b.locks.NewMutex(fmt.Sprintf("api.source.sink.jobs.%s", file.SinkKey))
```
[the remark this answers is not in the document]
jachym-tousek-keboola: Agreed.

internal/pkg/service/stream/sink/type/tablesink/keboola/bridge/test/table.go
```diff
@@ -110,7 +110,7 @@ func MockTableStorageAPICalls(tb testing.TB, transport *httpmock.MockTransport)
 	// Create table job - ok
 	transport.RegisterResponder(
 		http.MethodGet,
-		`=~/v2/storage/jobs/.+$`,
+		`=~/v2/storage/jobs/1.*`,
```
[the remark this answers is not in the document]
jachym-tousek-keboola: Wait why the `1`? What if the job begins with `2`?

internal/pkg/service/stream/storage/model/job.go
```diff
@@ -0,0 +1,11 @@
+package model
+
+import (
+	"github.com/keboola/keboola-as-code/internal/pkg/service/stream/definition/key"
+)
+
+// Job contains all Keboola-specific data we need for polling jobs.
```
jachym-tousek-keboola, line 7: If it's keboola specific, why is it here rather than in the keboola bridge?

internal/pkg/service/stream/storage/model/repository/job/job_list.go
```diff
@@ -0,0 +1,19 @@
+package job
+
+import (
+	"github.com/keboola/keboola-as-code/internal/pkg/service/common/etcdop/iterator"
+	"github.com/keboola/keboola-as-code/internal/pkg/service/stream/storage/model"
+	"github.com/keboola/keboola-as-code/internal/pkg/service/stream/storage/model/repository/job/schema"
+)
+
+func (r *Repository) ListAll() iterator.DefinitionT[model.Job] {
```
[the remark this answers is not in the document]
Matovidlo: Ofc

internal/pkg/service/stream/storage/level/target/config/config.go
```diff
@@ -52,6 +53,7 @@ type OperatorConfig struct {
 
 // ImportConfig configures the file import.
 type ImportConfig struct {
+	SinkLimit   int               `json:"sinkLimit" configKey:"sinkLimit" configUsage:"Specifies limit of sink that is representable by number"`
```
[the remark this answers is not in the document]
Matovidlo: Basically the import is on top of `sink`. I would like to say somehow that we have some limitation on `sink` level that the import has to respect nevertheless the import conditions were satisfied.

internal/pkg/service/stream/sink/type/tablesink/keboola/bridge/file.go
Matovidlo: I think this should be still there.
Basically, if we import file and check it again, we don't wan to trigger the StorageJob again. I'll add it back.

internal/pkg/service/stream/sink/type/tablesink/keboola/bridge/job.go
Matovidlo: I added it into `ctxattr`

internal/pkg/service/stream/sink/type/tablesink/keboola/bridge/job.go
Matovidlo: I added it into `ctxattr`

internal/pkg/service/stream/storage/model/job.go
Matovidlo: Fixed commentary. It was copied from `keboolasink/bridge/schema/job.go`

internal/pkg/service/stream/storage/metacleanup/metacleanup.go
```diff
@@ -205,3 +245,72 @@ func (n *Node) isFileExpired(file model.File, age time.Duration) bool {
...
+		n.logger.Error(ctx, err.Error())
+		return err, false
+	}
+
+	// Purge the job
+	if err = n.storageRepository.Job().Purge(&job).RequireLock(mutex).Do(ctx).Err(); err != nil {
+		err = errors.PrefixErrorf(err, `cannot delete finished job "%s"`, job.JobKey)
+		n.logger.Error(ctx, err.Error())
+		return err, false
+	}
```
Matovidlo, line 310: This will cause error when we soft delete `sink` and we do not delete `storage/keboola/jobs`. Should we suppress the error or leave it fail? @jachym-tousek-keboola

hosekpeter (approved):
LGTM

internal/pkg/service/stream/storage/node/coordinator/filerotation/conditions.go
```diff
@@ -52,18 +57,25 @@ func (c fileRotationConditionResult) String() string {
...
 		return "time"
+	case sinkThrottled:
+		return "sinkThrottled"
 	default:
 		return "unknown"
 	}
 }
 
-func shouldImport(cfg targetConfig.ImportConfig, now, openedAt, expiration time.Time, stats statistics.Value) fileRotationConditionResult {
+func shouldImport(cfg targetConfig.ImportConfig, now, openedAt, expiration time.Time, stats statistics.Value, jobLimit int) fileRotationConditionResult {
```
jachym-tousek-keboola, line 67: `jobLimit` is incorrect name for this parameter, should be `runningJobs` probably... 

that said I wanna suggest a different solution, see the comment below

internal/pkg/service/stream/storage/node/coordinator/filerotation/conditions.go
```diff
@@ -52,18 +57,25 @@ func (c fileRotationConditionResult) String() string {
...
 	sinceOpened := now.Sub(openedAt).Truncate(time.Second)
 	if threshold := cfg.MinInterval.Duration(); sinceOpened < threshold {
 		// Min interval settings take precedence over other settings.
 		return newConditionResult(noConditionMet, "min interval between imports is not met")
 	}
 
+	if threshold := cfg.JobLimit; jobLimit >= threshold {
```
jachym-tousek-keboola, line 74: My current thought process is that instead of this we could add a new plugin similar to `UploadSlice` and `ImportFile`. Something like `IsThrottling` which would let the plugin say whether it's ready to accept a new upload / import. Could even be separate `CanAcceptNewUpload` and `CanAcceptNewImport` and let the bridge make that decision. Keboola doesn't have a limit on uploads I think but a different plugin might so Keboola bridge would always return true for `CanAcceptNewUpload` and `CanAcceptNewImport` would be decided based on the jobs.

If we let the bridge decide then another question is how should it make the decision. Should we track the current jobs as you implemented it? Or should we make an api call to keboola to ask for currently running jobs and let the bridge return false if keboola is busy? (Result should be cached for a configurable amount of time so that the plugin can give the same answer if asked repeatedly in a short amount of time about the same sink.)

What do you think? And probably a better quesiton, what would Michal think? :joy:

Matovidlo: The idea seems leggit, as it would be plugin. But basically we always evaluate plugins, so we need to enforce each provider to implement the plugin which returns by default `false` which would not throttle or set it as default value, when not implemented.

The problem lays in effective implementation. It is possible to move the implementation of storage job under keboola bridge and use mirroring. But, I am not sure whether it is correct step to have mirror in `bridge` component.
I would not want to use caching of HTTP requests and rather use ETCD as source of truth as it can go from multiple pod's and every pod would have different cache entry etc, so it would not be cached as someone would expect.

Matovidlo: 61160429a36a923505b729900ad9094c072ebc7e Moved the repository under `bridge`. Remove duplicate `job` model. Mirror is on top of bridge. I made it configurable in another commit. It has exported method to start mirroring as separate process. It is called under `coordinator/filerotation` module.
Created new plugin and attach it into bridge and evaluate it in `coordinator/filerotation`.
I hope it is now better as
* `+` I removed the second model, which fits more only `sink/type/tablesink/keboola` provider.
* `+` Other providers can register the plugin and evaluate, by default it is not throttled (`false`)
* `-` Bridge is now more complicated and contains repository, mirror and seems that it is more like component than library with plugins because of mirroring

internal/pkg/service/stream/sink/type/tablesink/keboola/bridge/model/job.go
```diff
@@ -0,0 +1,13 @@
...
+import (
+	"github.com/keboola/keboola-as-code/internal/pkg/service/stream/definition/key"
+)
+
+// Job contains all Keboola-specific data we need for polling jobs.
+// At the end throttle the sink, so it does not overloads the service.
+type Job struct {
+	key.JobKey
+	Deleted bool   `json:"-"` // internal field to mark the entity for deletion, there is no soft delete
+	Token   string `json:"token"`
```
jachym-tousek-keboola, line 12: I don't think Token should be copied here. You have the sink key meaning you can get retrieve the token for the sink when you need it.

```go
	// Get authorization token
	token, err := b.schema.Token().ForSink(file.SinkKey).GetOrErr(b.client).Do(ctx).ResultOrErr()
	if err != nil {
		return err
	}
```

internal/pkg/service/stream/sink/type/tablesink/keboola/bridge/model/repository/job/job_create.go
```diff
@@ -0,0 +1,33 @@
...
+		Read(func(ctx context.Context) op.Op {
+			return r.MustNotExist(k)
+		}).
+		// Create
+		Write(func(ctx context.Context) op.Op {
+			// Create
+			created = deepcopy.Copy(*input).(model.Job)
+			return r.save(&created)
+		}).
+		// Update the input entity, it the operation is successful
```
jachym-tousek-keboola, line 29: Looks like a typo? Should be `if` instead of `it`. I assume you copied this from elsewhere though...

internal/pkg/service/stream/sink/type/tablesink/keboola/bridge/test/table.go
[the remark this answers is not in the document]
jachym-tousek-keboola: I don't understand what you're talking about. I'll need some clarification on this... Though it seems to be only for a test so I guess it's not all that important.

internal/pkg/service/stream/sink/type/tablesink/keboola/config.go
```diff
@@ -8,10 +8,12 @@ import (
 type Config struct {
 	// EventSendTimeout is a timeout to perform slice upload event or file import event.
 	EventSendTimeout time.Duration `configKey:"eventSendTimeout" configUsage:"Timeout to perform upload send event of slice or import event of file"`
+	JobLimit         int           `configKey:"jobLimit" configUsage:"Number of import jobs running in keboola for single sink"`
 }
 
 func NewConfig() Config {
 	return Config{
 		EventSendTimeout: 30 * time.Second,
+		JobLimit:         2,
```
jachym-tousek-keboola, line 17: Why 2? :thinking:

internal/pkg/service/stream/storage/node/coordinator/coordinator.go
```diff
@@ -17,6 +17,11 @@ import (
 func Start(ctx context.Context, d dependencies.CoordinatorScope, cfg config.Config) error {
 	logger := d.Logger().WithComponent("storage.node.coordinator")
 	logger.Info(ctx, `starting storage coordinator node`)
+	// Change default bridge behaviour to use Mirror as it is required for sink throttling
+	err := d.KeboolaSinkBridge().MirrorJobs(d)
```
jachym-tousek-keboola, line 21: Shouldn't you pass `ctx` here and use `withoutCancel` inside `MirrorJobs` instead of `context.Background()`?

jachym-tousek-keboola (changes requested):

internal/pkg/service/stream/plugin/throttle.go
```diff
@@ -0,0 +1,27 @@
...
+type Throttle struct {
+	SinkKey  key.SinkKey
+	Provider targetModel.Provider
+}
+
+func (p *Plugins) RegisterThrottle(provider targetModel.Provider, fn throttleFn) {
+	p.throttle[provider] = fn
+}
+
+func (p *Plugins) IsThrottled(ctx context.Context, provider targetModel.Provider, sinkKey key.SinkKey) bool {
```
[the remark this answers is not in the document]
Matovidlo: I think you can use it in both, why would you do from something generic more specific thing? You can use it anywhere you would like to evaluate it. I don't uderstand why I would do `CanAcceptNewFileRotation` and than change it back to `IsThrottled` because I used it in `sliceRotation` for example.

internal/pkg/service/stream/sink/type/tablesink/keboola/bridge/model/job.go
Matovidlo: I use the token in `metacleanup.go`. I cannot access internal `schema` and I think it is easier to use single access to `job` instead of 2 calls to etcd because I did not store the Token within the `job`.

internal/pkg/service/stream/storage/node/coordinator/coordinator.go
Matovidlo: 6cd186a0915f6f6fb59febd0e09a3815958f9018

internal/pkg/service/stream/sink/type/tablesink/keboola/bridge/model/repository/job/job_create.go
Matovidlo: defdd2ca90ceeb940fad97242b874c87b9625a08

internal/pkg/service/stream/sink/type/tablesink/keboola/bridge/bridge.go
```diff
@@ -105,10 +105,10 @@ func (b *Bridge) isKeboolaStagingFile(file *model.File) bool {
 	return file.StagingStorage.Provider == stagingFileProvider
 }
 
-func (b *Bridge) MirrorJobs(d dependencies) error {
+func (b *Bridge) MirrorJobs(ctx context.Context, d dependencies) error {
 	// Mirror jobs
 	wg := &sync.WaitGroup{}
-	ctx, cancel := context.WithCancel(context.Background())
+	ctx, cancel := context.WithCancel(ctx)
```
jachym-tousek-keboola, line 111: Shouldn't this be like this? :thinking: I'm not sure because context cancellation is a topic I'm not very experienced with.

```go
ctx, cancel := context.WithCancel(context.WithoutCancel(ctx))
```

internal/pkg/service/stream/plugin/throttle.go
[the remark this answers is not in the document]
jachym-tousek-keboola: The point is that if slice upload also cared about this, we'd need to use a separate method anyway so that it can be decided independently. The fact that the bridge is unable to accept another file import does not imply it is unable to accept a slice upload (and vice-versa).

internal/pkg/service/stream/sink/type/tablesink/keboola/bridge/test/table.go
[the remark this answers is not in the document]
jachym-tousek-keboola: Okay I think I understand this slightly better now. But where is the id for a table job decided? How do you know it will always begin with a `1`? I was unable to find that.

internal/pkg/service/stream/sink/type/tablesink/keboola/bridge/bridge.go
Matovidlo: Once the `coordinator` node is canceled, you want to tear down the `mirror` in the bridge. It does not have totally separate livespan. Using the `coordinator` context, once it is canceled you can teardown the bridge `mirror`. You do not need to wait for termination. It can be done upfront.

If you use thing that you post, the `mirror` can be canceled only when process shutdown signal is received even though rest of program is already in teardown state . Either way we miss the dependency that the `mirror` should be canceled before the `coordinator`.

If you would like to have dependency between these two instances who is served when it has to be done somehow else. Either way it is not bulletproof that the `mirror` is stopped or `mirror` is updated after `coordinator` teardown

internal/pkg/service/stream/sink/type/tablesink/keboola/config.go
Matovidlo: I think that 1 is too low as there could be issues on cleanup side and it takes some time to cleanup the previous job runs as it is not checked frequently
0332f7eaaaae021286eaace6d1d9ac3a69fecb9c

internal/pkg/service/stream/plugin/throttle.go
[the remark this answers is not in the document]
Matovidlo: 577e20d055b0c7249651d3067ff12e376fce763e

internal/pkg/service/stream/sink/type/tablesink/keboola/config.go
jachym-tousek-keboola: Ah I see. Keboola can accept more jobs but it won't run more than one at a time. So by using 2 here we're allowing to queue one extra one and then we start to throttle. Sounds about right?

internal/pkg/service/stream/plugin/throttle.go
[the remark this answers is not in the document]
jachym-tousek-keboola: Much better though I'd drop the Rotation part... `CanAcceptNewFileRotation` => `CanAcceptNewFile`. The plugin doesn't care about rotations, it just decides whether it can process another file.

internal/pkg/service/stream/sink/type/tablesink/keboola/bridge/bridge.go
jachym-tousek-keboola: I see. Thanks for explaining.

internal/pkg/service/stream/sink/type/tablesink/keboola/bridge/model/job.go
jachym-tousek-keboola: Are you planning to make some changes regarding this or not? For the file upload I believe the credentials are actually different for each file and need to be as they're. This token is not different for each job, only for each sink.

internal/pkg/service/stream/plugin/throttle.go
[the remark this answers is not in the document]
Matovidlo: 0ef0afedab52a89f2926daa5da154d8502c6477c

internal/pkg/service/stream/sink/type/tablesink/keboola/bridge/model/job.go
Matovidlo: 41c0053ff2be13d37daae46cd7cfbc793cf485b9

internal/pkg/service/stream/sink/type/tablesink/keboola/bridge/job.go
```diff
@@ -10,13 +10,17 @@ import (
 	"github.com/keboola/keboola-as-code/internal/pkg/service/common/ctxattr"
 	"github.com/keboola/keboola-as-code/internal/pkg/service/stream/definition/key"
 	"github.com/keboola/keboola-as-code/internal/pkg/service/stream/plugin"
+	keboolasink "github.com/keboola/keboola-as-code/internal/pkg/service/stream/sink/type/tablesink/keboola"
 	"github.com/keboola/keboola-as-code/internal/pkg/service/stream/sink/type/tablesink/keboola/bridge/model"
 )
 
-func (b *Bridge) createJob(ctx context.Context, token string, file plugin.File, storageJob *keboola.StorageJob) error {
+func (b *Bridge) SinkToken(ctx context.Context, sinkKey key.SinkKey) (keboolasink.Token, error) {
```
jachym-tousek-keboola, line 17: Mhmm... I see why you didn't like this. I don't like it either.

What do you think about adding a `Cleanup` method to the bridge instead? I'm suggesting moving this entire block + cleanupJob into the bridge and simply calling bridge cleanup from `metacleanup.go`.

https://github.com/keboola/keboola-as-code/blob/a376096489a627100480d43bb6bdc64711af8fbb/internal/pkg/service/stream/storage/metacleanup/metacleanup.go#L174-L192

jachym-tousek-keboola (changes requested):

internal/pkg/service/stream/sink/type/tablesink/keboola/bridge/job.go
jachym-tousek-keboola: Possible second step is to use the plugin system for cleanup and simply call `n.plugins.Cleanup(ctx)` from `metacleanup.go`.

internal/pkg/service/stream/storage/metacleanup/metacleanup.go
```diff
@@ -176,10 +173,24 @@ func (n *Node) cleanMetadata(ctx context.Context) (err error) {
 		ListAll().
 		ForEach(func(job keboolaBridgeModel.Job, _ *iterator.Header) error {
```
jachym-tousek-keboola, line 174: Why don't we put this whole block into the bridge as I suggested? :thinking: I'd like to avoid using `keboolaBridgeRepository` here in `metacleanup`.

Matovidlo: You will face the thing, that part of cleaning is spans/traces inside `bridge` which missleads, because real Process/component that reports the cleanup is `Metadata` component.
The most important is this one 
```
// There can be several cleanup nodes, each node processes an own part.
if !n.dist.MustCheckIsOwner(job.ProjectID.String()) {
	return nil
}
```
This says that there is only 1 coordinator node that cleans up the metadata, so there are not race conditions or other issues due to duplicate cleanups. This cannot be done on bridge level because it is not a process but a library.

jachym-tousek-keboola: I see. I'll see if I can improve this using range over functions but that can be done separately. :)

jachym-tousek-keboola (approved):

PR #2126, opened 2024-10-29
fix: cleanup test do not wait exit

Matovidlo:
**Changes:**
- split log check so the behaviour of task is always checked before shudown 

-----------


hosekpeter:
this change is also addressed in this commit [678514d](https://github.com/keboola/keboola-as-code/pull/2128/commits/678514dd2bbc3d613c0f3dff82158f905e24b5bc)

jachym-tousek-keboola:
Fixed in https://github.com/keboola/keboola-as-code/pull/2128

PR #2130, opened 2024-10-31
feat: Add source undelete event

jachym-tousek-keboola:
Jira: https://keboola.atlassian.net/browse/PSGO-916

**Changes:**
- Added source undelete event
- Added missing admin checks

-----------


internal/pkg/service/stream/sink/type/tablesink/keboola/bridge/event.go
```diff
@@ -167,7 +167,7 @@ func SendEvent(
 		Message:     msg(err),
 		Type:        "info",
 		Duration:    client.DurationSeconds(duration),
-		Results: map[string]any{
+		Params: map[string]any{
```
jachym-tousek-keboola, line 170: Should be Params instead of Results based on KBC team feedback: https://github.com/keboola/event-schema/pull/22#discussion_r1824305903

internal/pkg/service/stream/api/service/sink.go
```diff
@@ -62,13 +62,13 @@ func (s *service) CreateSink(ctx context.Context, d dependencies.SourceRequestSc
...
-				result = s.mapper.WithTaskOutputs(result, sink.SinkKey)
-				return result
-			} else {
+			if err := op.Do(ctx).Err(); err != nil {
 				return task.ErrResult(err)
 			}
+
+			result := task.OkResult("Sink has been created successfully.")
+			result = s.mapper.WithTaskOutputs(result, sink.SinkKey)
+			return result
```
jachym-tousek-keboola, line 71: In my opinion the intended (happy) path should be left-aligned and errors should be handled in the ifs so I swapped the branches accordingly and removed unnecessary else branches.

internal/pkg/service/stream/api/service/sink.go
```diff
@@ -434,6 +434,12 @@ func (s *service) EnableSink(ctx context.Context, d dependencies.SinkRequestScop
 }
 
 func (s *service) UndeleteSink(ctx context.Context, scope dependencies.SinkRequestScope, payload *api.UndeleteSinkPayload) (res *api.Task, err error) {
+	// If user is not admin deny access for write
+	token := scope.StorageAPIToken()
+	if token.Admin == nil || token.Admin.Role != adminRole {
+		return nil, svcerrors.NewForbiddenError(s.adminError)
+	}
```
jachym-tousek-keboola, line 441: After debating with Martin for a while we concluded it's not worth covering with tests as it would significantly complicate the setup for testing projects. Each developer would need to create a second account, invite it as guest into all his test projects and generate a token for that user so that we could have both admin and guest token in projects.json.

hosekpeter (approved):
LGTM

internal/pkg/service/stream/sink/type/tablesink/keboola/bridge/event.go
Matovidlo: I commented it in the PR. Please bare in mind, the `telemetry` team scrapes the events and take the `results` currently. You have to align with them. Also the Streams telemetry is currently binded to `results` to be correctly billed.

Matovidlo (approved):

PR #2131, opened 2024-10-31
 fix: Preserve state on rollback

jachym-tousek-keboola:
Jira: https://keboola.atlassian.net/browse/PSGO-918

**Changes:**
- Preserve enabled/disabled/deleted/undeleted on rollback


test/templates/api/projects/instance-delete/008-list-instances/expected-http-code
jachym-tousek-keboola, line 1: Unrelated test fix.

https://github.com/keboola/keboola-as-code/tree/992fef0c1c99d4677c4c1ae945653f176e7d6f65/test/templates/api/projects/instance-delete

This e2e test was randomly failing. After checking it I noticed that there are two directories prefixed with 005. The list-instances should be 008 so that it's executed after both polls are finished.

internal/pkg/service/stream/definition/repository/source/source_version.go
```diff
@@ -54,6 +54,15 @@ func (r *Repository) RollbackVersion(k key.SourceKey, now time.Time, by definiti
 			// Prepare the new value
 			versionDescription := fmt.Sprintf(`Rollback to version "%d".`, targetVersion.Version.Number)
 			updated = deepcopy.Copy(*targetVersion).(definition.Source)
+
+			// Preserve Disabled and Deleted states from the latest version.
+			// Without this it would be possible to bypass the Enable/Undelete event and avoid billing.
+			updated.Disabled = deepcopy.Copy(latestVersion.Disabled).(*definition.Disabled)
+			updated.Enabled = deepcopy.Copy(latestVersion.Enabled).(*definition.Enabled)
+			updated.Deleted = deepcopy.Copy(latestVersion.Deleted).(*definition.Deleted)
+			updated.Undeleted = deepcopy.Copy(latestVersion.Undeleted).(*definition.Undeleted)
```
jachym-tousek-keboola, line 63: What do you think about this? The intention is to make it so that the client can't avoid billing by disabling a source and then rollbacking to it's enabled state or vice versa being billed incorrectly after rollbacking from enabled to disabled state. We could alternatively try to detect these changes on rollback and fire the relevant events instead. Which way do you think is better?

internal/pkg/service/stream/definition/repository/sink/sink_enable_test.go
```diff
@@ -106,7 +106,7 @@ func TestSinkRepository_Enable(t *testing.T) {
 		now = now.Add(time.Hour)
 		sink, err := repo.RollbackVersion(sinkKey, now, by, 2).Do(ctx).ResultOrErr()
 		require.NoError(t, err)
-		assert.True(t, sink.IsDisabled())
+		assert.False(t, sink.IsDisabled()) // Rollback intentionally doesn't change disabled/enabled state
```
hosekpeter, line 109: how? I thought that the rollback action changes the state.

hosekpeter (approved):

internal/pkg/service/stream/definition/repository/sink/sink_version.go
```diff
@@ -54,6 +54,15 @@ func (r *Repository) RollbackVersion(k key.SinkKey, now time.Time, by definition
 			// Prepare the new value
 			versionDescription := fmt.Sprintf(`Rollback to version "%d".`, targetVersion.Version.Number)
 			updated = deepcopy.Copy(*targetVersion).(definition.Sink)
+
+			// Preserve Disabled and Deleted states from the latest version.
+			// Without this it would be possible to bypass the Enable/Undelete event and avoid billing.
+			updated.Disabled = deepcopy.Copy(latestVersion.Disabled).(*definition.Disabled)
+			updated.Enabled = deepcopy.Copy(latestVersion.Enabled).(*definition.Enabled)
+			updated.Deleted = deepcopy.Copy(latestVersion.Deleted).(*definition.Deleted)
+			updated.Undeleted = deepcopy.Copy(latestVersion.Undeleted).(*definition.Undeleted)
```
Matovidlo, line 63: Can we do some method for deep copying the objects? I don't like the duplication. If we define that we want to preserve previous states of objects, it should be called as helper as the copy copies only certain fields, not whole objects fields.

internal/pkg/service/stream/definition/repository/source/source_version.go
Matovidlo: I don't mind this approach, just make sure the code is not duplicated that much. Create some "deep-partial-copy" function

Matovidlo (approved):
LGTM. Won't block it now, i'll think about code duplication

PR #2138, opened 2024-11-11
build: Update dependencies

jachym-tousek-keboola:
Jira: https://keboola.atlassian.net/browse/PSGO-940

**Changes:**
- Updated dependencies

-----------


provisioning/stream/kubernetes/templates/api/deployment.yaml
```diff
@@ -92,6 +92,8 @@ spec:
               valueFrom:
                 fieldRef:
                   fieldPath: metadata.labels['tags.datadoghq.com/version']
+            - name: OTEL_METRICS_EXEMPLAR_FILTER
+              value: "always_off"
```
jachym-tousek-keboola, line 96: I found a similar issue here: https://github.com/open-telemetry/opentelemetry-go/issues/5936

Couldn't find a good place where to call `sdkmetric.WithExemplarFilter(exemplar.AlwaysOffFilter)` in our code since we don't use `sdkmetric` directly anywhere but I found [this function](https://github.com/open-telemetry/opentelemetry-go/blob/ac5caeaee88d744b21dc3789ebda79ae97f6cfa5/sdk/metric/config.go#L158-L172) which suggests that we can disable exemplars using this env variable.

Matovidlo (approved):
LGTM

provisioning/stream/docker/k6/Dockerfile
```diff
@@ -2,7 +2,7 @@ FROM grafana/xk6:latest AS build
 
 RUN xk6 build \
     --with github.com/LeonAdato/xk6-output-statsd@latest \
-    --with github.com/domsolutions/xk6-fasthttp@latest
+    --with github.com/domsolutions/xk6-fasthttp@0.1.1
```
jachym-tousek-keboola, line 5: Latest version of `github.com/domsolutions/xk6-fasthttp` requires Go 1.23 but `grafana/xk6` Dockerfile is built from Go `1.22.7`.

PR #2140, opened 2024-11-12
fix: Stream events schema

jachym-tousek-keboola:
Requires https://github.com/keboola/event-schema/pull/23 to be merged and release which is currently blocked by Telemetry team.

Waiting for https://github.com/keboola/connection/pull/5356.

### DIFF
--- a/internal/pkg/service/stream/sink/type/tablesink/keboola/bridge/event.go
+++ b/internal/pkg/service/stream/sink/type/tablesink/keboola/bridge/event.go
@@ -174,6 +174,7 @@ func SendEvent(
 			"streamId":  params.SourceKey.String(),
 			"sinkId":    params.SinkID,
 		},
+		Results: map[string]any{},
 	}
 	if params.SourceName != "" {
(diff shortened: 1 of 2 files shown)


internal/pkg/service/stream/sink/type/tablesink/keboola/bridge/event.go
```diff
@@ -174,15 +174,17 @@ func SendEvent(
...
 		event.Params["sourceName"] = params.SourceName
 	}
 
 	var sErr error
 	defer func() {
-		// BC compatibility, should be removed later.
-		event.Results = event.Params
+		if len(event.Results) == 0 {
+			event.Results = nil
+		}
```
Matovidlo, line 187: Why did we add this?

jachym-tousek-keboola: I initialize the `Results: map[string]any{},` above but when it's empty it can be removed from the event. It's not technically necessary but the results is marked in the schema of our events as optional.

PR #2144, opened 2024-11-15
feat: Add data app template filter

hosekpeter:
Jira: [PSGO-882](https://keboola.atlassian.net/browse/PSGO-882)

**Changes:**
- ...

-----------


api/templates/design.go
```diff
@@ -181,10 +181,11 @@ var _ = Service("templates", func() {
 		Meta("openapi:summary", "List templates in the repository")
 		Description("List all templates  defined in the repository.")
 		Result(Templates)
-		Payload(RepositoryRequest)
+		Payload(TemplatesRequest)
 		HTTP(func() {
 			GET("/repositories/{repository}/templates")
 			Meta("openapi:tag:template")
+			Param("filterBy")
```
Matovidlo, line 188:
```suggestion
			Param("filter")
```
I would expect that I use `/repositories/keboola/templates?filter=keboola.data-apps`. the `filterBy=keboola.data-apps` seems weird but it is just my weird taste :smile:

internal/pkg/service/templates/api/service/mapper.go
```diff
@@ -130,6 +134,25 @@ func TemplatesResponse(ctx context.Context, d dependencies.ProjectRequestScope,
...
 
+		if filterBy != nil && *filterBy != "" {
+			filterString := *filterBy
+
+			t, found := tmpl.DefaultVersion()
+			if !found {
+				continue
+			}
+			switch filterString {
+			case keboola.DataAppsComponentID.String():
```
Matovidlo, line 145: I would rather define new constant here. The `go-client` is not the best place to get constants from. It is keboola specific thing that can change due to some API change and therefore the UI would need to update also templates pages.

internal/pkg/service/templates/api/service/mapper.go
```diff
@@ -130,6 +134,25 @@ func TemplatesResponse(ctx context.Context, d dependencies.ProjectRequestScope,
...
 		}
 
+		if filterBy != nil && *filterBy != "" {
+			filterString := *filterBy
+
+			t, found := tmpl.DefaultVersion()
+			if !found {
+				continue
+			}
+			switch filterString {
```
Matovidlo, line 144:
```suggestion
			t, found := tmpl.DefaultVersion()
			if !found {
				continue
			}
			
			filterString := *filterBy
			switch filterString {
```

test/templates/api/repositories/template-list-filter/002-list-ok-data-apps-only/expected-response.json
```diff
@@ -0,0 +1,65 @@
...
+    "ref": "",
+    "author": {
+      "name": "Example Author",
+      "url": "https://example.com"
+    }
+  },
+  "templates": [
+    {
+      "id": "my-template-id-5",
+      "name": "My Template-snowflake-5",
```
Matovidlo, line 14: How do I know that these templates are for data apps?

test/templates/api/repositories/template-list-filter/004-list-ok-wrong-id/expected-response.json
```diff
@@ -0,0 +1,115 @@
+{
+  "repository": {
```
Matovidlo, line 2: How can wrong ID in `filter` return all components? I would say that it returns empty slice.

api/templates/design.go
hosekpeter: [770aa80](https://github.com/keboola/keboola-as-code/pull/2144/commits/770aa8021ee46f9f69bc9fe881453872b36d593c)

internal/pkg/service/templates/api/service/mapper.go
hosekpeter: [0a285f9](https://github.com/keboola/keboola-as-code/pull/2144/commits/0a285f96b192a0fe48f094e048828c14095fbcc5)

internal/pkg/service/templates/api/service/mapper.go
hosekpeter: [0a285f9](https://github.com/keboola/keboola-as-code/pull/2144/commits/0a285f96b192a0fe48f094e048828c14095fbcc5)

test/templates/api/repositories/template-list-filter/004-list-ok-wrong-id/expected-response.json
hosekpeter: [905b218](https://github.com/keboola/keboola-as-code/pull/2144/commits/905b218b1fa9bfaedd62b73ca09946de26845a6d)

api/templates/design.go
```diff
@@ -567,6 +568,11 @@ var TemplateRequest = Type("TemplateRequest", func() {
 	Required("template")
 })
 
+var TemplatesRequest = Type("TemplatesRequest", func() {
+	Extend(RepositoryRequest)
+	Attribute("filter", String)
```
jachym-tousek-keboola, line 573: Shouldn't it be an enum rather than a string? From the code it seems that values other than `"keboola.data-apps"` and `"keboola.components"` don't do anything.

Matovidlo: Good question, It would be nice that these values are in the API docs and generated by `goa` so we don't need to create separate documentation.
We currently do not know about other values

Matovidlo (approved):
LGTM

api/templates/design.go
hosekpeter: [aa1c10b](https://github.com/keboola/keboola-as-code/pull/2144/commits/aa1c10bc266644f5a140cdf1dee92f7384a3c629) - enum+description

[f1bab1a](https://github.com/keboola/keboola-as-code/pull/2144/commits/f1bab1a7f6069d9ce708967def31d0a63332cdbb) - update test

hosekpeter: @jachym-tousek-keboola it was a good catch 👍

jachym-tousek-keboola (approved):

api/templates/design.go
jachym-tousek-keboola: Much better, thank you!

PR #2145, opened 2024-11-16
fix: Do not log wakeup error when app is in stopping state

Matovidlo:
**Changes:**
- Wait till the app is stopped and perform wakeup again, do not log the error.
- Increase context timeout to sandboxes service to 60s

-----------

### DIFF
--- a/internal/pkg/service/appsproxy/dataapps/wakeup/wakeup.go
+++ b/internal/pkg/service/appsproxy/dataapps/wakeup/wakeup.go
@@ -71,7 +71,11 @@ func (l *Manager) Wakeup(ctx context.Context, appID api.AppID) error {
 	item.nextRequestAfter = now.Add(Interval)
 
 	// Send the notification
-	if _, err := l.api.WakeupApp(appID).Send(ctx); err != nil && err.Error() != wakeupErrorToBeSkipped {
+	_, err := l.api.WakeupApp(appID).Send(ctx)
+	// If it does not succeed but app is currently stopping do not log it as error, log only other errors
+	// Instead of implementing state machine as in sandboxes service, we want to skip valid state that the
+	// pod is dealocating and we want to wait till pod is `stopped` and we can `start` the pod again.
+	if err != nil && err.Error() != wakeupErrorToBeSkipped {
 		l.logger.Errorf(ctx, `failed sending wakeup request to Sandboxes Service about for app "%s": %s`, appID, err.Error())
 		return err
 	}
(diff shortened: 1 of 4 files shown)


hosekpeter (approved):
LGTM

internal/pkg/service/appsproxy/dataapps/wakeup/wakeup.go
```diff
@@ -68,7 +71,7 @@ func (l *Manager) Wakeup(ctx context.Context, appID api.AppID) error {
 	item.nextRequestAfter = now.Add(Interval)
 
 	// Send the notification
-	if _, err := l.api.WakeupApp(appID).Send(ctx); err != nil {
+	if _, err := l.api.WakeupApp(appID).Send(ctx); err != nil && err.Error() != wakeupErrorToBeSkipped {
```
jachym-tousek-keboola, line 74: So you want to just ignore this specific error and behave as though wakeup succeeded? I don't like that tbh. Maybe just use lower log level then error instead? Also add some comment why we're treating this case differently.

jachym-tousek-keboola (changes requested):

internal/pkg/service/appsproxy/dataapps/wakeup/wakeup.go
Matovidlo: c971ed3496ef7dee7a7599cb1778552e53d94ee4

internal/pkg/service/appsproxy/dataapps/wakeup/wakeup.go
```diff
@@ -68,7 +71,9 @@ func (l *Manager) Wakeup(ctx context.Context, appID api.AppID) error {
 	item.nextRequestAfter = now.Add(Interval)
 
 	// Send the notification
-	if _, err := l.api.WakeupApp(appID).Send(ctx); err != nil {
+	_, err := l.api.WakeupApp(appID).Send(ctx)
+	// If it does not succeed but app is currently stopping do not log it as error, log only other errors
```
jachym-tousek-keboola, line 75: I can tell as much from the code. What I want in the comment is the reason why we're doing this.

jachym-tousek-keboola (changes requested):

internal/pkg/service/appsproxy/dataapps/wakeup/wakeup.go
Matovidlo: 3979510edd27e6305524d2e04f4b316a3a8e732f

jachym-tousek-keboola (approved):

PR #2153, opened 2024-11-21
hotfix: Pipeline does not test all code, just that one that it is expected to be released

Matovidlo:
**Changes:**
- CI release should test only top level packages without other service ones
- Possibly extend it to more modules that should be omitted

-----------

### DIFF
--- a/.github/workflows/release-cli.yml
+++ b/.github/workflows/release-cli.yml
@@ -57,6 +57,7 @@ jobs:
     uses: ./.github/workflows/test-unit.yml
     with:
       without-cache: true
+      test-regex: "./internal/pkg/service/stream|./internal/pkg/service/templates|./internal/pkg/service/appsproxy"
 
   test-e2e-cli:
     name: "E2E: CLI"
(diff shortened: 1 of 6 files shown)


.github/workflows/release-cli.yml
```diff
@@ -57,6 +57,7 @@ jobs:
     uses: ./.github/workflows/test-unit.yml
     with:
       without-cache: true
+      test-regex: "./internal/pkg/service/stream|./internal/pkg/service/templates|./internal/pkg/service/appsproxy"
```
jachym-tousek-keboola, line 60: This is quite ugly to be honest.
1. The regex seems to note paths that are ignored which is not what I'd expect from this name.
2. Any time we add a new component we'd have to adjust these regexes everywhere.

Matovidlo: 1. I will come with some better name
2. You do not need to, this is hotfix to add any type of path that you would like to omit from tests.

I am not going to waste more time on hotfix as we need to come up with solution.

Matovidlo: 0b54d17169d23065377c159e3aea22ceaf2688e5

jachym-tousek-keboola (approved):

hosekpeter (approved):

PR #2156, opened 2024-11-26
fix: documentation stream minikube and configuration

Matovidlo:
**Changes:**
- Configuration of metacleanup interval changed to 30s
- Add minikube metrics endpoint so we are able to measure CPU/memory
- Add documentation how to spin off local deployment

-----------


jachym-tousek-keboola (approved):

hosekpeter (approved):
LGTM

PR #2157, opened 2024-11-26
feat: Token encryption

jachym-tousek-keboola:
Jira: https://keboola.atlassian.net/browse/PSGO-911

**Changes:**
- Encrypt keboola tokens in Stream API

-----------


provisioning/stream/local.sh
```diff
@@ -56,6 +56,9 @@ export STREAM_VOLUME_FAST_STORAGE_SIZE="100Mi"
 export STREAM_VOLUME_MEDIUM_STORAGE_SIZE="100Mi"
 export STREAM_VOLUME_SLOW_STORAGE_SIZE="100Mi"
 
+export STREAM_ENCRYPTION_PROVIDER="native"
+export STREAM_ENCRYPTION_NATIVE_SECRET_KEY="00000000000000000000000000000000"
```
jachym-tousek-keboola, line 60: Where to take the secret key for localhost from? I could randomly generate it but that doesn't feel technically correct since it should be preserved between deploys. :thinking:

provisioning/stream/kubernetes/templates/api/deployment.yaml
```diff
@@ -94,6 +94,10 @@ spec:
                   fieldPath: metadata.labels['tags.datadoghq.com/version']
             - name: OTEL_METRICS_EXEMPLAR_FILTER
               value: always_off
+            - name: STREAM_ENCRYPTION_PROVIDER
+              value: "${STREAM_ENCRYPTION_PROVIDER}"
+            - name: STREAM_ENCRYPTION_NATIVE_SECRET_KEY
+              value: "${STREAM_ENCRYPTION_NATIVE_SECRET_KEY}"
```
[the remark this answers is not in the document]
Matovidlo: Why don't we use configuration? At least for provider it would be easier, the secret key can be passed by variable.

provisioning/stream/local.sh
Matovidlo: I would expect similarly to the library. It has some local provisioning. As it is shell script you can use any random number generator basically. E.g get random number from `/dev/random` https://unix.stackexchange.com/questions/268952/using-dev-random-dev-urandom-to-generate-random-data

provisioning/stream/kubernetes/templates/api/deployment.yaml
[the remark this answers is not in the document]
jachym-tousek-keboola: But how? :thinking:

provisioning/stream/local.sh
jachym-tousek-keboola: In case of the library you just need to run tests so it doesn't matter if the generated secret key ends up being different every time. Here however you might want to locally deploy one version, then change something and locally deploy the the fix in which case the key should stay the same.

internal/pkg/service/common/dependencies/encryption.go
```diff
@@ -0,0 +1,47 @@
...
+	BaseScope
+	EtcdClientScope
+}
+
+func NewEncryptionScope(ctx context.Context, cfg encryption.Config, d encryptionScopeDeps) (EncryptionScope, error) {
+	return newEncryptionScope(ctx, cfg, d)
+}
+
+func newEncryptionScope(ctx context.Context, cfg encryption.Config, d encryptionScopeDeps) (v *encryptionScope, err error) {
+	ctx, span := d.Telemetry().Tracer().Start(ctx, "keboola.go.common.dependencies.NewEncryptionScope")
```
Matovidlo, line 27:
```suggestion
	ctx, span := d.Telemetry().Tracer().Start(ctx, "keboola.go.common.dependencies.EncryptionScope")
```

internal/pkg/service/stream/components.go
```diff
@@ -22,6 +23,7 @@ const (
 	ComponentStorageCoordinator Component = "storage-coordinator"
 	ComponentStorageWriter      Component = "storage-writer"
 	ComponentStorageReader      Component = "storage-reader"
+	ComponentMigrator           Component = "migrator"
```
Matovidlo, line 26: We do not want migration component but CLI command. Are you able to drop this after migration easily? You need to write up commits which will be dropped after successful migration.
That's why having a cmd is easier, just drop the `migration` folder and done.

internal/pkg/service/stream/encryption/config.go
```diff
@@ -0,0 +1,52 @@
...
+	if c.Provider != ProviderGCP {
+		c.GCP = nil
+	}
+	if c.Provider != ProviderAWS {
+		c.AWS = nil
+	}
+	if c.Provider != ProviderAzure {
+		c.Azure = nil
+	}
+}
```
Matovidlo, line 52: When you have constructor and skip initialization it has `nil` by design. Why do we want to set it ino `nil` values? `Normalize` by naming means that we set it to somehow "default" values which this unset's the values and does nothing with the values. Cannot we just set desired encryptor as in `NewEncryptor`? Maybe this is similarly implemented in `go-client` but seem's as a hack.

internal/pkg/service/stream/encryption/encryption.go
```diff
@@ -0,0 +1,72 @@
...
+			return nil, err
+		}
+	}
+
+	if config.Provider != ProviderNative {
+		encryptor, err = cloudencrypt.NewDualEncryptor(ctx, encryptor)
+		if err != nil {
+			return nil, err
+		}
+	}
```
Matovidlo, line 52:
```suggestion
        if config.Provider == ProviderNative {
                encryptor, err = cloudencrypt.NewNativeEncryptor([]byte(config.Native.SecretKey))
		if err != nil {
			return nil, err
		}
		
		return encryptor, nil
        }
        
	switch config.Provider {
	case ProviderGCP:
		encryptor, err = cloudencrypt.NewGCPEncryptor(ctx, config.GCP.KMSKeyID)
		if err != nil {
			return nil, err
		}
	case ProviderAWS:
		encryptor, err = cloudencrypt.NewAWSEncryptor(ctx, config.AWS.Region, config.AWS.KMSKeyID)
		if err != nil {
			return nil, err
		}
	case ProviderAzure:
		encryptor, err = cloudencrypt.NewAzureEncryptor(ctx, config.Azure.KeyVaultURL, config.Azure.KeyName)
		if err != nil {
			return nil, err
		}
	}

	encryptor, err = cloudencrypt.NewDualEncryptor(ctx, encryptor)
	if err != nil {
		return nil, err
        }
```
Basically the code want's to say that we do not want to create dual encryptor from native encryptor, so it goes on the top as first decision. When we have cloud specific vendor we do the switch and wrap it up with DualEncryptor always. 
We get rid of the if statement at the end which would someone refactor by putting the code into switch statement which was my first iteration.
After while I realised we can do simple if statement to go with fast path decision and rest of construct remains unchanged without any other decisions.
So we do not duplicate code and cut out the statement from `switch` into early `if` check
For me this is more readable that we have just one if statement at start and do not evaluate if over each `switch` case.

internal/pkg/service/stream/encryption/encryption.go
```diff
@@ -0,0 +1,72 @@
...
+
+	return encryptor, nil
+}
+
+const charset = "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789"
+
+func RandomSecretKey() (string, error) {
```
Matovidlo, line 59: This I would put under `utils/encryption`

internal/pkg/service/stream/sink/type/tablesink/keboola/bridge/slice.go
Matovidlo, line 1: What about this `UploadCredentials` ? It should be encrypted as well. Just asking whether you know about this that this should be encrypted as well.

internal/pkg/service/stream/sink/type/tablesink/keboola/bridge/token.go
```diff
@@ -157,5 +188,67 @@ func (b *Bridge) tokenForSink(ctx context.Context, now time.Time, sink definitio
 	)
 
 	b.logger.Info(ctx, "created token")
-	return newToken, nil
+	return *result, nil
+}
+
+func (b *Bridge) MigrateTokens(ctx context.Context) error {
```
Matovidlo, line 194: I think you have issue, that this migration is run with each restart of program. Can you verify it in test? Then add some protection that migration can be done only once. That's why I would rather do this function inside command line

internal/pkg/service/stream/sink/type/tablesink/keboola/bridge/token_test.go
```diff
@@ -0,0 +1,233 @@
...
+					return bridgeSchema.Token().ForSink(newToken.SinkKey).Put(d.EtcdClient(), newToken)
+				}),
+			).Do(ctx).Err()
+		require.NoError(t, err)
+	}
+
+	// Migrate tokens
+	// -----------------------------------------------------------------------------------------------------------------
+	err := d.KeboolaSinkBridge().MigrateTokens(ctx)
+	require.NoError(t, err)
```
Matovidlo, line 199: Call this multiple times to be sure that migration is done only once

internal/pkg/service/stream/sink/type/tablesink/keboola/bridge/token.go
Matovidlo: Also can we try it saving into another prefix just for testing ? I see the test but be rather sure that we do not override the tokens incorrectly

Matovidlo (changes requested):

internal/pkg/service/common/dependencies/encryption.go
jachym-tousek-keboola: image

internal/pkg/service/stream/components.go
jachym-tousek-keboola: Yeah it's easy to revert.

internal/pkg/service/stream/encryption/config.go
jachym-tousek-keboola: > Why do we want to set it ino nil values?

I assume that assigning the values of the inner object using ENV variables would not work if the objects were nil... I will try it..

internal/pkg/service/stream/encryption/encryption.go
jachym-tousek-keboola: I considered that but I think the current way is better because this way we can easily wrap the encryptor by logging encryptor for instance.

internal/pkg/service/stream/sink/type/tablesink/keboola/bridge/slice.go
jachym-tousek-keboola: I actually disagree. Those are temporary credentials for single file only.

internal/pkg/service/stream/encryption/encryption.go
jachym-tousek-keboola: It's fine here in my opinion. If anything it could be moved into tests.

internal/pkg/service/common/dependencies/encryption.go
Matovidlo: I don't like the `New` prefix. Maybe create jira task for that. Why do we want to trace it like that ?

internal/pkg/service/stream/encryption/encryption.go
Matovidlo: You can do it also with if statement at the top using `defer` function. For me it is logically worse to read the code above because it has some if conditions at the end that we do not want to wrap dual encryptor to basic one, but why do I need to know such detail ? Simply put the if statement on top of that and you do not need if statement at the end or even you can use `defer` for wrapping up the `DualEncryptor`

internal/pkg/service/stream/encryption/encryption.go
Matovidlo: This is pure utility/separate function. Check for example `idgenerator`. I would like to split generic randoms in each module to have it together so we can reuse it. I would be beneficial for us to prevent code duplication and having random strings, numbers etc thrown over monorepository

internal/pkg/service/stream/sink/type/tablesink/keboola/bridge/slice.go
Matovidlo: Can we check what's their expiration? In case they expire (have expiration set) it is fine for me to resolve this comment.

internal/pkg/service/stream/encryption/config.go
```diff
@@ -0,0 +1,52 @@
...
+	return Config{
+		Provider: ProviderNative,
+		Native:   &NativeConfig{},
+		GCP:      &GCPConfig{},
+		AWS:      &AWSConfig{},
+		Azure:    &AzureConfig{},
+	}
+}
+
+func (c *Config) Normalize() {
```
Matovidlo, line 39: Have you check that?

internal/pkg/service/common/dependencies/encryption.go
jachym-tousek-keboola: I believe these spans are just for the creation of the dependencies so it kinda makes sense to me? Anyway this was set up by Michal so I don't really know the answers.

internal/pkg/service/stream/encryption/encryption.go
jachym-tousek-keboola: As you wish. https://github.com/keboola/keboola-as-code/commit/16bfb39a0d06011d971624f2aad6727714b99b46

internal/pkg/service/stream/encryption/config.go
jachym-tousek-keboola: You mean where this method is called? It's part of `ValueWithNormalization` interface and invoked here:

https://github.com/keboola/keboola-as-code/blob/60c9f4f79f093a2a722d933c80028109d0b4a662/internal/pkg/service/common/configmap/bind.go#L245-L248

internal/pkg/service/stream/encryption/config.go
jachym-tousek-keboola: As expected, having GCP, AWS and Azure be nil by default causes a panic when `STREAM_ENCRYPTION_GCP_KMS_KEY_ID` is defined.

image

internal/pkg/service/stream/sink/type/tablesink/keboola/bridge/slice.go
jachym-tousek-keboola: It depends on the cloud provider and on what keboola connection sets there... but yeah they certainly do have expiration. Relevant part from AWS docs for example:

> If you create a presigned URL with the Amazon S3 console, the expiration time can be set between 1 minute and 12 hours. If you use the AWS CLI or AWS SDKs, the expiration time can be set as high as 7 days.

Is this good enough for you or should I ask connection team what the expiration is?

internal/pkg/service/stream/encryption/encryption.go
jachym-tousek-keboola: Removed. https://github.com/keboola/keboola-as-code/commit/e90d3cbb5b6b5978e8ca702aa11f0f6b43b12aa9

internal/pkg/service/stream/sink/type/tablesink/keboola/bridge/bridge.go
```diff
@@ -69,9 +73,23 @@ type dependencies interface {
...
 }
 
-func New(d dependencies, apiProvider apiProvider, config keboolasink.Config) *Bridge {
+func New(d dependencies, apiProvider apiProvider, config keboolasink.Config) (*Bridge, error) {
+	cache, err := ristretto.NewCache(
+		&ristretto.Config[[]byte, []byte]{
+			NumCounters: 1e6,
+			MaxCost:     1 << 20,
+			BufferItems: 64,
+		},
```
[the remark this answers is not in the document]
jachym-tousek-keboola: Well we could of course. I just don't think it's worth the time to be honest. I don't think we'll need to change the values.

internal/pkg/service/stream/encryption/encryption.go
Matovidlo: Still not the best but better. The `return` statement is quite hidden.

internal/pkg/service/stream/sink/type/tablesink/keboola/bridge/slice.go
Matovidlo: It's ok, We can ask anyway just to be sure as we just receive it from the `connection`

jachym-tousek-keboola: I changed my mind and implemented it.

Matovidlo (approved):
LGTM

internal/pkg/service/stream/sink/type/tablesink/keboola/bridge/job.go
```diff
@@ -73,15 +74,29 @@ func (b *Bridge) CleanJob(ctx context.Context, job model.Job) (err error, delete
...
+	// Decrypt token
+	var token keboola.Token
+	if existingToken.EncryptedToken != nil {
+		token, err = b.tokenEncryptor.Decrypt(ctx, existingToken.EncryptedToken, metadata)
+		if err != nil {
+			return err, false
+		}
+	} else {
+		token = *existingToken.Token
+	}
```
hosekpeter, line 95: Couldn't we move this logic into a separate function? It's repeated in several places.

internal/pkg/service/stream/encryption/encryption.go
```diff
@@ -0,0 +1,63 @@
...
+		return nil, nil
+	case ProviderNative:
+		encryptor, err = cloudencrypt.NewNativeEncryptor(config.Native.SecretKey)
+		if err != nil {
+			return nil, err
+		}
+
+		return encryptor, nil
+	case ProviderGCP:
+		encryptor, err = NewGCPEncryptor(ctx, config.GCP.KMSKeyID, logger)
```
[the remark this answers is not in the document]
jachym-tousek-keboola: yeah, that was just debug, it's already reverted

internal/pkg/service/stream/sink/type/tablesink/keboola/bridge/token_test.go
```diff
@@ -0,0 +1,265 @@
...
+			).Do(ctx).Err()
+		require.NoError(t, err)
+	}
+
+	// Migrate tokens
+	// -----------------------------------------------------------------------------------------------------------------
+	{
+		err := d.KeboolaSinkBridge().MigrateTokens(ctx)
+		require.NoError(t, err)
+	}
```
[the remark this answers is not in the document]
jachym-tousek-keboola: It's covered by e2e tests, not quite sure how to do it otherwise to be honest. The `TestBridge_MigrateTokens` creates dummy tokens, then calls migration and then tries to decrypt them. If I write a similar test without the migration then both encryption and decryption would be inside the test which would be pointless.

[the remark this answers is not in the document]
Matovidlo: But we had issue with this particular code.
In the library all the tests were passing but in production it somehow did not work. I am not sure whether this is good approach leave it like this.

Matovidlo (approved):
LGTM

PR #2163, opened 2024-11-28
refactor: Keboola bridge

jachym-tousek-keboola:
1. Apparently we forgot to put the FileKey to File object. All other objects saved to etcd have their own key identifier so it should be here as well. We don't need a migration, the old files where it's `nil` will be deleted by cleanup after they're processed. Then we can make FileKey mandatory and remove SinkKey since FileKey contains it.
2. JobKey was left outside of keboola bridge by accident so I moved it.


hosekpeter (approved):
LGTM

Matovidlo (approved):

PR #2165, opened 2024-12-02
fix: Simplify Dockerfiles

jachym-tousek-keboola:
@Matovidlo You missed what I wanted so here you go. ;-)

### DIFF
--- a/provisioning/stream/docker/service/race/Dockerfile
+++ b/provisioning/stream/docker/service/race/Dockerfile
@@ -1,8 +1,7 @@
 # Build container
 FROM golang:1.23.3-alpine3.20 AS buildContainer
-RUN apk add -U --no-cache bash make curl
-# This is needed for CGO_ENABLED racer
-RUN apk add -U --no-cache gcc g++
+# gcc and g++ are needed for CGO_ENABLED racer
+RUN apk add --update --no-cache bash make curl gcc g++
 WORKDIR /app
 
 COPY Makefile Makefile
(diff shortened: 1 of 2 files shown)


provisioning/stream/docker/service/race/Dockerfile
```diff
@@ -1,8 +1,7 @@
 # Build container
 FROM golang:1.23.3-alpine3.20 AS buildContainer
-RUN apk add -U --no-cache bash make curl
-# This is needed for CGO_ENABLED racer
-RUN apk add -U --no-cache gcc g++
+# gcc and g++ are needed for CGO_ENABLED racer
+RUN apk add --update --no-cache bash make curl gcc g++
```
Matovidlo, line 4: I mind a bit that we install some dependencies for the container runtime and some for the compilation using CGO. How can you know  what to omit in case we would like to disable CGO?

jachym-tousek-keboola: Hmm? I mean I adjusted the comment to specify that gcc and g++ are for GCO...

Matovidlo: Ok then, but I think it is same as you would separate it and just remove lines. I'll resolve it

Matovidlo (approved):

provisioning/stream/docker/service/race/Dockerfile
jachym-tousek-keboola: It isn't. Splitting the lines creates another layer. It's a good practice to minimize the layers unless you have a good reason not to.

PR #2176, opened 2024-12-09
fix: Fix templates tests

hosekpeter:
Jira: [PSGO-898](https://keboola.atlassian.net/browse/PSGO-898)

**Changes:**
- 

-----------
After running a local test we got this. We can see there, that dynamic IDs don't match. Therefore I am ignoring this IDs for tests.
image

[PSGO-898]: https://keboola.atlassian.net/browse/PSGO-898?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/internal/pkg/utils/testhelper/diff.go
+++ b/internal/pkg/utils/testhelper/diff.go
@@ -4,6 +4,7 @@ package testhelper
 import (
 	"context"
 	"fmt"
+	"regexp"
 	"strings"
 
 	"github.com/keboola/go-utils/pkg/wildcards"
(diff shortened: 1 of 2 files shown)


Matovidlo (changes requested):
Can we add test what actually changed? I see normalization of string but I don't see the tests associated with this change

Matovidlo (approved):
LGTM

PR #2177, opened 2024-12-10
fix: unit test CI no space left on device

Matovidlo:
**Changes:**
- Remove more things from Ubuntu image due to no space left on device.
- Add new logs into CI to show how much space is left after unit tests.
- Remove custom gRPC metrics that we do not use to save money on DD.

-----------


hosekpeter (approved):
LGTM

PR #2178, opened 2024-12-10
fix: Use kbc-stacks trigger locally

wokoman:
Public repos can't use GitHub Actions from private repos, so we're "forking" the kbc-stacks trigger here. The workflow itself still runs in kbc-stacks, it's just triggered from someplace else now 🤷 

Worked here: https://github.com/keboola/keboola-as-code/actions/runs/12256767310
Which triggered this (it is fine it's failed): https://github.com/keboola/kbc-stacks/actions/runs/12256770547/job/34192752960

You would need to presumably delete [this tag](https://github.com/keboola/keboola-as-code/tree/production-stream-v0.3.23) and create new one :( Sowwy.

### DIFF
--- a/.github/actions/trigger-image-tag-update/action.yaml
+++ b/.github/actions/trigger-image-tag-update/action.yaml
@@ -1,0 +1,66 @@
+name: Trigger image tag update
+description: 'Trigger image tag update workflow'
+inputs:
+  helm-chart:
+    description: "The Helm chart to update"
+    required: true
+  image-tag:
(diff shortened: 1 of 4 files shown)


kacurez (approved):
lgtm
len dodam ze sme sa o tom bavili tu https://keboolaglobal.slack.com/archives/C055K9CRFDX/p1733834074491109

Matovidlo (approved):
LGTM

PR #2184, opened 2024-12-13
fix: The generic errors were missing status codes

Matovidlo:
JIRA: [PSGO-971](https://keboola.atlassian.net/browse/PSGO-971)

**Changes:**
- Return appropriate status codes instead of 500.
- Stops reporting it as critical issue in DD
- Adjust message a little more that the project and template are misaligned.

-----------

[PSGO-971]: https://keboola.atlassian.net/browse/PSGO-971?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/internal/pkg/service/templates/api/service/service.go
+++ b/internal/pkg/service/templates/api/service/service.go
@@ -609,8 +609,9 @@ func repositoryRef(d dependencies.ProjectRequestScope, name string) (model.Templ
 		return repo, nil
 	} else {
 		return model.TemplateRepository{}, &GenericError{
-			Name:    "templates.repositoryNotFound",
-			Message: fmt.Sprintf(`Repository "%s" not found.`, name),
+			StatusCode: http.StatusNotFound,
+			Name:       "templates.repositoryNotFound",
+			Message:    fmt.Sprintf(`Repository "%s" not found.`, name),
 		}
 	}
 }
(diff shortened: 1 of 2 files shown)


internal/pkg/service/templates/api/service/service.go
```diff
@@ -609,8 +609,9 @@ func repositoryRef(d dependencies.ProjectRequestScope, name string) (model.Templ
 		return repo, nil
 	} else {
 		return model.TemplateRepository{}, &GenericError{
-			Name:    "templates.repositoryNotFound",
-			Message: fmt.Sprintf(`Repository "%s" not found.`, name),
+			StatusCode: http.StatusNotFound,
```
jachym-tousek-keboola, line 612: So you're changing status codes in some cases. Are these status codes correctly documented in `design.go` for the relevant endpoints?

jachym-tousek-keboola (approved):

internal/pkg/service/templates/api/service/service.go
Matovidlo: I see only `RepositoryIndexTemplatesRepositoryNotFoundResponseBody` but that's body for Repository not found for `RepositoryIndex` api endpoint. I see that these functions return `GenericError`. I see in docs that the endpoint can return `404` but we returned `500`. Maybe this is just mistake in our codebase?

jachym-tousek-keboola: Probably, yeah.

jachym-tousek-keboola (approved):

PR #2185, opened 2024-12-13
feat: Add configID as task result when preview endpoint succeed

Matovidlo:
Jira: [PSGO-970](https://keboola.atlassian.net/browse/PSGO-970)

**Changes:**
- ConfigID is returned by async task for possible redirect on UI

-----------

[PSGO-970]: https://keboola.atlassian.net/browse/PSGO-970?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


pkg/lib/operation/project/local/template/use/operation.go
```diff
@@ -206,6 +210,26 @@ func PrepareTemplate(ctx context.Context, d dependencies, o ExtendedOptions) (pl
 		existingObjects[objectState.Key()] = true
 		plan.modified = append(plan.modified, ModifiedObject{ObjectState: objectState, OpMark: opMark})
 
+		var configID uint64
+		if objectState.Kind().IsConfig() {
+			// Change config name that is used in the template instead of meta.json one
+			config := objectState.(*model.ConfigState)
+			config.Local.Name = plan.options.ConfigName
```
Matovidlo, line 217: hotfix

internal/pkg/service/templates/api/service/mapper.go
```diff
@@ -84,6 +84,19 @@ func (m Mapper) TaskPayload(model task.Task) (r *Task) {
 				InstanceID: &v,
 			}
 		}
+
+		if v, ok := model.Outputs["configId"].(float64); ok {
```
jachym-tousek-keboola, line 88: why the float handling here?

pkg/lib/operation/project/local/template/use/operation.go
```diff
@@ -273,7 +299,12 @@ func (p *TemplatePlan) Invoke(ctx context.Context) (*Result, error) {
 		logger.Warn(ctx, `Push operation is only possible when project is valid.`)
 	}
 
-	result := &Result{InstanceID: p.options.InstanceID}
+	result := p.result
+	if result == nil {
+		result = &Result{InstanceID: p.options.InstanceID}
+	} else {
+		result.InstanceID = p.options.InstanceID
+	}
```
jachym-tousek-keboola, line 307: What is this change for?

Matovidlo: It could be omited as the template use creates `InstanceID` which UI uses. I wanted to add `ConfigID` of the template created juts for info but it will not be used.

internal/pkg/service/templates/api/service/mapper.go
Matovidlo: I'll check it, it is due to some internal parsing to `uint64`->`any`, it converts it back to `float64` for some unknown reason.

Matovidlo: https://keboola.atlassian.net/browse/PSGO-973

internal/pkg/service/templates/api/service/mapper.go
```diff
@@ -84,9 +85,19 @@ func (m Mapper) TaskPayload(model task.Task) (r *Task) {
 				InstanceID: &v,
 			}
 		}
+
+		if v, ok := model.Outputs["configId"].(string); ok {
```
jachym-tousek-keboola, line 89: This way it's still internally a float and here it's cast to string and then parsed, isn't it?
I'd rather just use type string for it instead of uint.

pkg/lib/operation/project/local/template/use/operation.go
jachym-tousek-keboola: I don't get it. How is InstanceID related to ConfigID?

Matovidlo: It's not related, it is in returned values.

internal/pkg/service/templates/api/service/mapper.go
jachym-tousek-keboola: My point is that I don't see any value in typing IDs as int or uint considering they could become uuids in the future. Having IDs typed as string is more future proof imo.

pkg/lib/operation/project/local/template/use/operation.go
jachym-tousek-keboola: I checked the code again, I think I understand now. Let's keep this change. :+1:

jachym-tousek-keboola (approved):

internal/pkg/service/templates/api/service/mapper.go
Matovidlo: I can check with UI how they work with such values.

Matovidlo: Changed to `string`

PR #2195, opened 2025-01-02
build: Replace clock library

jachym-tousek-keboola:
Jira: https://keboola.atlassian.net/browse/PSGO-406

### DIFF
--- a/.github/workflows/test-codeql-analysis.yml
+++ b/.github/workflows/test-codeql-analysis.yml
@@ -37,6 +37,10 @@ jobs:
         # Learn more about CodeQL language support at https://aka.ms/codeql-docs/language-support
 
     steps:
+    # Removing this directory helps to prevent a "no space left on device" error.
+    - name: Remove useless data
+      run: rm -rf ./opt/hostedtoolcache/CodeQL
+
     - name: Checkout repository
       uses: actions/checkout@v4
 
--- a/internal/pkg/service/stream/storage/level/local/encoding/count/counter.go
+++ b/internal/pkg/service/stream/storage/level/local/encoding/count/counter.go
@@ -158,6 +158,10 @@ func (c *CounterWithBackup) SyncBackup() error {
 	c.lock.RLock()
 	defer c.lock.RUnlock()
 
+	if c.backup == nil {
+		return nil
+	}
+
 	// Seek to the beginning of the file
 	// The size counter can only grow, so it guarantees that the entire file will be overwritten.
 	if _, err := c.backup.Seek(0, io.SeekStart); err != nil {
@@ -189,9 +193,14 @@ func (c *CounterWithBackup) Close() error {
 		return err
 	}
 
+	c.lock.Lock()
+	defer c.lock.Unlock()
+
 	if err := c.backup.Close(); err != nil {
 		return errors.Errorf(`cannot close the backup file: %w`, err)
 	}
+
+	c.backup = nil
 
 	return nil
 }
--- a/internal/pkg/service/stream/storage/level/local/encoding/count/counter.go
+++ b/internal/pkg/service/stream/storage/level/local/encoding/count/counter.go
@@ -11,7 +11,7 @@ import (
 	"sync"
 	"time"
 
-	"github.com/benbjohnson/clock"
+	"github.com/jonboulle/clockwork"
 
 	"github.com/keboola/keboola-as-code/internal/pkg/log"
 	"github.com/keboola/keboola-as-code/internal/pkg/service/common/utctime"
@@ -32,7 +32,7 @@ type Counter struct {
 type CounterWithBackup struct {
 	*Counter
 	backup       backup
-	backupTicker *clock.Ticker
+	backupTicker clockwork.Ticker
 }
 
 // backup interface contains used methods from *os.File.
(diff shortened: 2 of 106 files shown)


internal/pkg/service/stream/storage/level/local/encoding/count/counter_test.go
```diff
@@ -152,9 +152,11 @@ func TestCounterWithBackup_SyncBackupPeriodically(t *testing.T) {
...
 	// Sync backup by clock
 	clk.Advance(backupInterval)
-	content, err := os.ReadFile(backupPath)
-	require.NoError(t, err)
-	assert.Equal(t, "5,2001-01-01T00:00:00.000Z,2002-01-01T00:00:00.000Z", string(content))
+	assert.EventuallyWithT(t, func(collect *assert.CollectT) {
+		content, err := os.ReadFile(backupPath)
+		require.NoError(t, err)
+		assert.Equal(t, "5,2001-01-01T00:00:00.000Z,2002-01-01T00:00:00.000Z", string(content))
+	}, 5*time.Second, 10*time.Millisecond)
```
jachym-tousek-keboola, line 159: Not quite sure if this is the best way to fix the test.

internal/pkg/service/stream/storage/level/local/encoding/size/size_test.go
```diff
@@ -185,9 +185,11 @@ func TestMeterWithBackup_SyncBackupPeriodically(t *testing.T) {
...
 	// Sync backup by clock
 	clk.Advance(backupInterval)
-	content, err := os.ReadFile(backupPath)
-	require.NoError(t, err)
-	assert.Equal(t, "6", string(content))
+	assert.EventuallyWithT(t, func(collect *assert.CollectT) {
+		content, err := os.ReadFile(backupPath)
+		require.NoError(t, err)
+		assert.Equal(t, "6", string(content))
+	}, 5*time.Second, 10*time.Millisecond)
```
jachym-tousek-keboola, line 192: Similarly here, I'm unsure if this is the best way to fix it.

internal/pkg/service/stream/storage/level/local/encoding/size/size.go
```diff
@@ -31,6 +31,7 @@ type MeterWithBackup struct {
 	*Meter
 	backup       backup
 	backupTicker clockwork.Ticker
+	lock         sync.Mutex
```
jachym-tousek-keboola, line 34: Trying to fix this problem:

image

I don't like using a lock here but couldn't figure out a better solution.

internal/pkg/service/stream/storage/level/local/encoding/count/counter.go
```diff
@@ -189,9 +193,14 @@ func (c *CounterWithBackup) Close() error {
 		return err
 	}
 
+	c.lock.Lock()
+	defer c.lock.Unlock()
```
jachym-tousek-keboola, line 197: Similar issue in `TestCounterWithBackup_SyncBackupPeriodically.

image

.github/workflows/test-codeql-analysis.yml
```diff
@@ -37,6 +37,10 @@ jobs:
         # Learn more about CodeQL language support at https://aka.ms/codeql-docs/language-support
 
     steps:
+    # Removing this directory helps to prevent a "no space left on device" error.
+    - name: Remove useless data
+      run: rm -rf ./opt/hostedtoolcache/CodeQL
```
jachym-tousek-keboola, line 42: I ran into "no space left on device" issue on linter job again today. So instead of just cleaning up the cache on AWS I ran some experiments with

```yaml
      - working-directory: /
        run: du | sort -rn | head -n100 -$worstOffenders
```

This directory is pretty high among the worst offenders, see:

Screenshot from 2025-01-07 15-24-36

Note: it took me a while to realize that the numbers are in blocks of 1024, meaning kilobytes. So the directory takes 3.5 GB. Looks like a good candidate to get rid of in my opinion.

Matovidlo:
@jachym-tousek-keboola I will skip the review as there is still failing E2E test for streams https://github.com/keboola/keboola-as-code/actions/runs/12653776006/job/35297084560?pr=2195
Get back on that once working to prevent multiple went through the code :slightly_smiling_face:

internal/pkg/service/stream/storage/level/local/encoding/count/counter_test.go
Matovidlo: If you advance for `backupInterval+1` it won't work? I am not sure but I would expect that when you advance the time the test should generate the path. I don't know why it does not create it after advance. Maybe there is some skew in clock? 
For me we can leave it like this

Matovidlo (approved):
LGTM. Let's see how the CI will work

internal/pkg/service/stream/storage/level/local/encoding/count/counter_test.go
jachym-tousek-keboola: > If you advance for backupInterval+1 it won't work? 
Unfortunately no. This seems to work but not quite sure what's better:

```
	clk.Advance(backupInterval)
	time.Sleep(time.Millisecond)
```

Matovidlo: better is `EventuyallyWithT`, but I mind that it does work in old version.

.github/workflows/test-codeql-analysis.yml
Matovidlo: let's try this

hosekpeter (approved):

PR #2203, opened 2025-01-10
feat: Add decryption-encryption test on bridge

Matovidlo:
**Changes:**
- Add missing test for decryption/encryption of token
- Add testing provider which is the same as in cloudencrypt package. The only thing is that we Seal GCM with static bytes.

-----------

### DIFF
--- a/internal/pkg/service/stream/sink/type/tablesink/keboola/bridge/fixtures/encrypt_sink_snapshot.txt
+++ b/internal/pkg/service/stream/sink/type/tablesink/keboola/bridge/fixtures/encrypt_sink_snapshot.txt
@@ -1,0 +1,54 @@
+<<<<<
+storage/keboola/file/123/456/my-source/my-sink/2000-01-01T01:00:00.000Z
+-----
+{
+  "FileKey": {
+    "projectId": 123,
+    "branchId": 456,
+    "sourceId": "my-source",
+    "sinkId": "my-sink",
+    "fileOpenedAt": "2000-01-01T01:00:00.000Z"
+  },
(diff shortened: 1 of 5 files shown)


internal/pkg/service/stream/sink/type/tablesink/keboola/bridge/fixtures/encrypt_sink_snapshot.txt
```diff
@@ -0,0 +1,54 @@
...
+storage/keboola/secret/token/123/456/my-source/my-sink
+-----
+{
+  "projectId": 123,
+  "branchId": 456,
+  "sourceId": "my-source",
+  "sinkId": "my-sink",
+  "token": null,
+  "tokenId": "1001",
+  "encryptedToken": "AAECAwQFBgcICQoL%s"
```
jachym-tousek-keboola, line 42: I don't see a point in verifying the encrypted value. Imo we should just use native, check that we can decrypt successfully and simply use %s here.

Matovidlo: In case you change the library functionality and the encryptedToken will be prefixed or so, we need to do migration. How we are able to say that upon updating our library the code is not broken on both sides?

jachym-tousek-keboola (changes requested):
Please rebase on top of main since https://github.com/keboola/keboola-as-code/pull/2208 is merged now.

jachym-tousek-keboola (approved):

PR #2208, opened 2025-01-16
feat: Prefix encrypted values

jachym-tousek-keboola:


.github/workflows/test-lint.yml
```diff
@@ -13,6 +13,10 @@ jobs:
     name: lint
     runs-on: ubuntu-latest
     steps:
+      # Removing this directory helps to prevent a "no space left on device" error.
+      - name: Remove useless data
+        run: rm -rf /opt/hostedtoolcache
```
jachym-tousek-keboola, line 18: Somehow I put this into wrong file? How did that even happen? :exploding_head: 

Note for myself: I used this to debug the space issues:
```
      - working-directory: /
        if: always()
        run: du | sort -rn | head -n100 -$worstOffenders
```

internal/pkg/service/stream/encryption/encryption.go
```diff
@@ -47,7 +45,22 @@ func NewEncryptor(ctx context.Context, config Config) (cloudencrypt.Encryptor, e
...
+			return nil, err
+		}
+	}
+
+	encryptor, err = cloudencrypt.NewBase64Encryptor(encryptor)
+	if err != nil {
+		return nil, err
+	}
+
+	encryptor, err = cloudencrypt.NewPrefixEncryptor(encryptor, []byte(prefix))
```
Matovidlo, line 63: Just a question. Does it matter if I put `Base64` encryptor after `PrefixEncryptor` ? The chain can be sorted only in this way?

Matovidlo (approved):
LGTM just question

hosekpeter (approved):
LGTM

PR #2209, opened 2025-01-16
fix: Proxy wakeup stability of tests

Matovidlo:
Jira: [PSGO-768](https://keboola.atlassian.net/browse/PSGO-768)

**Changes:**
- Add port manager util package.
- Save AAAA record instead of A when having IPv6 address.
- Add better error reporting when saving <nil> IPv4/IPv6 address into DNS record.
- Auto create AAAA record instead of A record on IPv6 address.
- Remove `httptest.SetupUnstartedServer` and use `PortManager` instead.
- Add new env variable `TEST_KBC_TMP_DIR` to set tmp dir where tmp things are saved. Used for CI

-----------

[PSGO-768]: https://keboola.atlassian.net/browse/PSGO-768?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


internal/pkg/service/appsproxy/proxy/testutil/api.go
```diff
@@ -79,8 +82,17 @@ func StartDataAppsAPI(t *testing.T) *DataAppsAPI {
 		}
 	})
 
-	ts := httptest.NewUnstartedServer(mux)
-	ts.EnableHTTP2 = true
+	port := pm.GetFreePort()
+	l, err := net.Listen("tcp", "127.0.0.1:"+strconv.FormatInt(int64(port), 10))
```
jachym-tousek-keboola, line 86: These IP + port strings would be more readable with fmt.Sprintf imo.

internal/pkg/service/appsproxy/proxy/testutil/dns.go
```diff
@@ -31,7 +32,14 @@ func AddAppDNSRecord(t *testing.T, appServer *AppServer, dnsServer *dnsmock.Serv
 	require.NoError(t, err)
 
 	appHost := "app.local"
-	dnsServer.AddARecord(dns.Fqdn(appHost), net.ParseIP(ip))
+	err = dnsServer.AddARecord(dns.Fqdn(appHost), net.ParseIP(ip))
+	var derr *dnsmock.DNSRecordError
+	if err != nil && errors.As(err, &derr) {
+		err = dnsServer.AddAAAARecord(dns.Fqdn(appHost), net.ParseIP(ip))
```
jachym-tousek-keboola, line 38: I'd suggest detecting whether the IP is v4 or v6 and calling either A or AAAA based on that instead of this fallbacking. Or just remove this if it isn't necessary.

jachym-tousek-keboola (changes requested):

internal/pkg/utils/server/port.go
```diff
@@ -0,0 +1,94 @@
...
+	defer p.mu.Unlock()
+	randomPort := random.Intn(len(p.ports))
+	port := p.ports[randomPort]
+	dir := path.Join(p.dir, strconv.FormatInt(int64(port), 10))
+	for _, err := os.Open(dir); err == nil; {
+		randomPort = random.Intn(len(p.ports))
+		port = p.ports[randomPort]
+		dir = path.Join(p.dir, strconv.FormatInt(int64(port), 10))
+		_, err = os.Open(dir)
+	}
```
hosekpeter: perhaps, I would use this magic in a separate function 🤔 .

hosekpeter (approved):
LGTM 🎉

internal/pkg/utils/server/port.go
```diff
@@ -74,17 +74,8 @@ func (p portManager) GetFreePort() int {
...
-		if len(p.ports) == 0 {
-			return 0
-		}
 
-		randomPort = p.random.IntN(len(p.ports))
-		port = p.ports[randomPort]
-		dir = path.Join(p.dir, strconv.FormatInt(int64(port), 10))
-		_, err = os.Open(dir) // nolint:forbidigo
-	}
+	port = p.retryOnOccupiedPort(dir, port)
```
hosekpeter, line 78: I think, it looks better 👍 🎉

internal/pkg/utils/server/port.go
Matovidlo: 5413d832b

internal/pkg/service/appsproxy/proxy/testutil/dns.go
Matovidlo: 76efb49b9a87b04f6b0a44a036801fd8ad7fe920 I switched it into IPv6 addresses with client resolution.

internal/pkg/service/appsproxy/proxy/testutil/api.go
Matovidlo: 76efb49b9a87b04f6b0a44a036801fd8ad7fe920

internal/pkg/service/appsproxy/proxy/testutil/dns.go
jachym-tousek-keboola: You didn't address what I dislike here. I don't like that you call `AddARecord` always and `AddAAAARecord` as fallback when it fails. You should check the IP and detect which one of them you should call in my opinion.

jachym-tousek-keboola (changes requested):
Unadressed comment.

jachym-tousek-keboola (approved):

PR #2215, opened 2025-01-21
feat: Add missing snowflake writers

hosekpeter:
Jira: [PSGO-951](https://keboola.atlassian.net/browse/PSGO-951)

**Changes:**
- added missing snowflake writer for ComponentsResponse() and SnowflakeWriterComponentID
- added test
-----------

[PSGO-951]: https://keboola.atlassian.net/browse/PSGO-951?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/internal/pkg/template/jsonnet/function/function.go
+++ b/internal/pkg/template/jsonnet/function/function.go
@@ -168,6 +168,8 @@ func SnowflakeWriterComponentID(components *model.ComponentsMap) *jsonnet.Native
 				return SnowflakeWriterIDAzure.String(), nil
 			} else if _, found := components.Get(SnowflakeWriterIDGCPS3); found {
 				return SnowflakeWriterIDGCPS3.String(), nil
+			} else if _, found := components.Get(SnowflakeWriterIDGCP); found {
+				return SnowflakeWriterIDGCP.String(), nil
 			} else {
 				return nil, errors.New("no Snowflake Writer component found")
 			}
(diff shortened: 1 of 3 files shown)


internal/pkg/template/jsonnet/function/function.go
```diff
@@ -168,6 +168,8 @@ func SnowflakeWriterComponentID(components *model.ComponentsMap) *jsonnet.Native
 				return SnowflakeWriterIDAzure.String(), nil
 			} else if _, found := components.Get(SnowflakeWriterIDGCPS3); found {
 				return SnowflakeWriterIDGCPS3.String(), nil
+			} else if _, found := components.Get(SnowflakeWriterIDGCP); found {
```
Matovidlo, line 171: I miss test where this component is returned.
In general we should rework this function to take into account
`backend` that project runs on
`storage` that the project runs on

As this function is static (we do not take into account any project attributes) it will always fail in future if the correct components are set on stack.

hosekpeter: This test: https://github.com/keboola/keboola-as-code/blob/c88fb0e48fb8ee93ad833e9fab0c05028668b047/internal/pkg/service/templates/api/service/mapper_test.go#L153

Matovidlo (approved):
LGTM let's test it on dev template and with both cases.

PR #2217, opened 2025-01-23
feat: Log operations for statistics clear

jachym-tousek-keboola:


internal/pkg/service/common/etcdop/op/txn_result.go
```diff
@@ -16,6 +19,16 @@ func newErrorTxnResult[R any](err error) *TxnResult[R] {
...
 }
 
+func (v *TxnResult[R]) addTotals(totalIfs, totalThenOps, totalElseOps int) {
+	v.totalIfs = totalIfs
+	v.totalThenOps = totalThenOps
+	v.totalElseOps = totalElseOps
+}
+
+func (v *TxnResult[R]) TotalOps() int {
+	return v.totalIfs + v.totalThenOps + v.totalElseOps
```
jachym-tousek-keboola, line 29: Not quite sure about this to be honest. Do ifs count to the limit? Are both thenOps and elseOps counted or only those that end up being used?

internal/pkg/service/stream/api/service/sink.go
```diff
@@ -364,7 +364,10 @@ func (s *service) SinkStatisticsClear(ctx context.Context, d dependencies.SinkRe
 		return err
 	}
 
-	return d.StatisticsRepository().ResetSinkStats(d.SinkKey()).Do(ctx).Err()
+	result := d.StatisticsRepository().ResetSinkStats(d.SinkKey()).Do(ctx)
+	s.logger.Infof(ctx, `Statistics clear for sink "%s" used %d operations`, d.SinkKey().String(), result.TotalOps())
```
jachym-tousek-keboola, line 368: TxnResult and AtomicResult now contain ops count so that we can log it here.

provisioning/common/etcd/values.yaml
```diff
@@ -61,7 +61,7 @@ extraEnvVars:
     value: "256MiB"
   # Increase limit of operations in transaction.
   - name: ETCD_MAX_TXN_OPS
-    value: "1000"
+    value: "1024"
```
jachym-tousek-keboola, line 64: I think it's better to use these "binary" numbers here so that we can use count on statistics reset to be able to process 1000 files + slices and not worry about the few ops overhead.

internal/pkg/service/common/etcdop/op/txn_result.go
Matovidlo: That's up to the debugging of this MAX_TXN option. If it is not written in the documentation, simply create such transaction that you are able to investigate. I don't see a problem to have in `dev` container only `64` or so transactions and let the code fall (return error which you assert)

internal/pkg/service/common/etcdop/op/txn.go
```diff
@@ -219,12 +219,41 @@ func (v *TxnOp[R]) OnSucceeded(fn func(result *TxnResult[R])) *TxnOp[R] {
...
+		result.setMaxOps(v.maxOps(lowLevel.op().Op))
+		return result
 	} else {
 		return newErrorTxnResult[R](err)
 	}
 }
 
+// This code doesn't make sense but should replicate etcd behavior.
+// https://github.com/etcd-io/etcd/blob/e289ba30780208c1a464ea18f2f08259c75ac23a/server/etcdserver/api/v3rpc/key.go#L150-L179
+func (v *TxnOp[R]) maxOps(op etcd.Op) int {
```
jachym-tousek-keboola, line 232: I wanna note that I absolutely hate adding this method.

internal/pkg/service/stream/storage/statistics/repository/stats_delete_test.go
```diff
@@ -405,3 +405,142 @@ func TestRepository_RollupStatisticsOnFileDelete_LevelTarget(t *testing.T) {
 		assert.False(t, stats.Total.LastRecordAt.IsZero())
 	}
 }
+
+// TestRepository_RollupStatisticsOnFileDelete_LevelTarget_TxnLimit tests statistics reset near the transaction limit.
+func TestRepository_RollupStatisticsOnFileDelete_LevelTarget_TxnLimit(t *testing.T) {
```
jachym-tousek-keboola, line 410: hmm I'm unsure about including the test... on CI it is several times slower than locally and often times out even with increased time limit

internal/pkg/service/common/etcdop/op/txn.go
```diff
@@ -219,12 +219,41 @@ func (v *TxnOp[R]) OnSucceeded(fn func(result *TxnResult[R])) *TxnOp[R] {
...
-		return lowLevel.Do(ctx, opts...)
+		result := lowLevel.Do(ctx, opts...)
+		result.setMaxOps(v.maxOps(lowLevel.op().Op))
+		return result
 	} else {
 		return newErrorTxnResult[R](err)
 	}
 }
 
+// This code doesn't make sense but should replicate etcd behavior.
```
Matovidlo, line 230:
```suggestion
```

internal/pkg/service/stream/storage/statistics/repository/stats_delete_test.go
```diff
@@ -405,3 +405,142 @@ func TestRepository_RollupStatisticsOnFileDelete_LevelTarget(t *testing.T) {
...
+	}
+
+	// Reset statistics
+	// -----------------------------------------------------------------------------------------------------------------
+	{
+		clk.Advance(time.Hour)
+		require.NoError(t, statsRepo.ResetAllSinksStats([]key.SinkKey{sinkKey}).Do(ctx).Err())
+
+		stats, err := statsRepo.SinkStats(ctx, sinkKey)
+		require.NoError(t, err)
```
Matovidlo, line 537: I miss asserting the error `etcdserver: too many operations in txn request` This is what accutally happens when txn limit is reached and the logger outputs that. Maybe put it into e2e test?

Matovidlo (changes requested):

internal/pkg/service/stream/api/service/sink.go
```diff
@@ -364,7 +364,10 @@ func (s *service) SinkStatisticsClear(ctx context.Context, d dependencies.SinkRe
 		return err
 	}
 
-	return d.StatisticsRepository().ResetSinkStats(d.SinkKey()).Do(ctx).Err()
+	result := d.StatisticsRepository().ResetSinkStats(d.SinkKey()).Do(ctx)
+	s.logger.Infof(ctx, `Statistics clear for sink "%s" used %d operations`, d.SinkKey().String(), result.MaxOps())
```
Matovidlo, line 368: I don't see this asserted anywhere in code

internal/pkg/service/stream/api/service/source.go
```diff
@@ -343,7 +343,10 @@ func (s *service) SourceStatisticsClear(ctx context.Context, d dependencies.Sour
 		sinkKeys = append(sinkKeys, sink.SinkKey)
 	}
 
-	return d.StatisticsRepository().ResetAllSinksStats(ctx, sinkKeys)
+	result := d.StatisticsRepository().ResetAllSinksStats(sinkKeys).Do(ctx)
+	s.logger.Infof(ctx, `Statistics clear for source "%s" used %d operations`, d.SourceKey().String(), result.MaxOps())
```
Matovidlo, line 347: I don't see this asserted anywhere in code

internal/pkg/service/stream/api/service/sink.go
jachym-tousek-keboola: There is no test where it could be asserted, this part is afaik only covered by e2e tests.

internal/pkg/service/common/etcdop/op/txn.go
jachym-tousek-keboola: :shrug:

internal/pkg/service/stream/storage/statistics/repository/stats_delete_test.go
jachym-tousek-keboola: The test is running near the limit, not above it... should I change it?

internal/pkg/service/stream/api/service/sink.go
Matovidlo: It can be asserted anywhere using the same assertions as in e2e tests. Sample in unit test asserting logs 
https://github.com/keboola/keboola-as-code/blob/main/internal/pkg/service/stream/storage/node/coordinator/slicerotation/operator_test.go#L145

internal/pkg/service/stream/storage/statistics/repository/stats_delete_test.go
Matovidlo: I would have covered both cases to be sure what is running on the `edge` and beyond the `edge`. So simply copy and assert this log. This was present in DD log output so we know that it is covered

internal/pkg/service/stream/api/service/sink.go
jachym-tousek-keboola: I don't understand, we don't have a place to write such code in e2e tests since they're just a directory with files like requests and responses but generally no code.

Matovidlo (changes requested):
I still miss tests for e2e logs and reaching limit and checking specific string `etcdserver: too many operations in txn request` which should this PR check.
In subsequent PR we should change the default production environment to not reach the limit based on the metrics produced from this PR.

internal/pkg/service/stream/storage/statistics/repository/stats_delete_test.go
jachym-tousek-keboola: I'd have to copy the whole test for that... I don't have a way to assert both on edge and beyond edge in the same test.

Matovidlo: Yes that's why I wrote that you duplicate the test but do the beyond edge part there as this is below edge test

hosekpeter (approved):

Matovidlo (approved):
LGTM

PR #2220, opened 2025-01-27
fix: Preserve decrypted tokens on migration

jachym-tousek-keboola:
Migration now preserves the decrypted tokens. The script still tries to use and decrypt the encrypted token if it exists but if decryption fails it will just log an error and use the decrypted token.


hosekpeter (approved):
lgtm

PR #2222, opened 2025-01-28
feat: Add a snowflake writer placeholder in the jsonnet input

hosekpeter:
Jira: XXXXX

**Changes:**
- added a snowflake writer placeholder to the inputs.jsonnet
-----------

Tested on `dev` `/inputs`:
Bigquery backend:
image

Snowflake backend: 
image

### DIFF
--- a/internal/pkg/service/templates/api/service/mapper.go
+++ b/internal/pkg/service/templates/api/service/mapper.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"slices"
 	"sort"
+	"strings"
 	"time"
 
 	"github.com/spf13/cast"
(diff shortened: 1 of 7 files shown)


internal/pkg/service/templates/api/service/mapper.go
```diff
@@ -299,6 +300,29 @@ func ComponentsResponse(d dependencies.ProjectRequestScope, in []string) (out []
 	return out
 }
 
+// ReplacePlaceholders replaces the Snowflake writer component ID placeholder with the appropriate backend component ID.
+func ReplacePlaceholders(d dependencies.ProjectRequestScope, value string) string {
```
hosekpeter, line 304: Maybe a better naming? 🤔

Matovidlo: Namig seems fine. I would do that in generic way, so remove Snowflake writer. It can basically replace any placeholder defined in `manifest.go` module

Matovidlo (approved):
LGTM just change a bit description of function to be generic

internal/pkg/service/templates/api/service/mapper.go
hosekpeter: [cece3eb](https://github.com/keboola/keboola-as-code/pull/2222/commits/cece3eb00eec06ad10b3a886002a2406f6299a49)

jachym-tousek-keboola (approved):

PR #2228, opened 2025-01-30
fix: Disk writer cannot send remaining data after etcd shutdown

Matovidlo:
Jira: [PSGO-993](https://keboola.atlassian.net/browse/PSGO-993)

**Changes:**
- Delete after write does not happens in `watch` module.
  - Statistics seems to be working correctly. [PSGO-810](https://keboola.atlassian.net/browse/PSGO-810)
  - Diskwriter seems to be working correctly. [PSGO-993](https://keboola.atlassian.net/browse/PSGO-993)
  - Probablly solves this issue as well [PSGO-830](https://keboola.atlassian.net/browse/PSGO-830)
  - Probablly solves this issue as well [PSGO-855](https://keboola.atlassian.net/browse/PSGO-855) 

Issues
- Removed sorting in ETCD transaction/compact operation to prevent Delete after Create, test included.
- Do not change distribution nodes when no changes were made
- Do not change volumes within connection when no changes were made. Closing and recreating of transport is extensive operation, we save some CPU/MEM during ETCD shutdown.
- Open connection in dialer section. Should prevent issues that `connection.volume` has no old reference but always up to date reference from set of connection volumes  
- One small race condition on waitgroup. Use w.file.Sync() instead of Sync() method in Close as waitgroup is not longer needed during close of writer
-----------

[PSGO-993]: https://keboola.atlassian.net/browse/PSGO-993?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[PSGO-810]: https://keboola.atlassian.net/browse/PSGO-810?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[PSGO-830]: https://keboola.atlassian.net/browse/PSGO-830?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/internal/pkg/service/common/distribution/node.go
+++ b/internal/pkg/service/common/distribution/node.go
@@ -3,7 +3,6 @@ package distribution
 import (
 	"context"
 	"fmt"
-	"maps"
 	"sync"
 	"time"
 
@@ -251,12 +250,6 @@ func (n *GroupNode) updateNodesFrom(ctx context.Context, events []etcdop.WatchEv
 		}
 	}
 
-	// Make sure that nodes are updated only when there is maximum nodes
-	if len(n.duplicateNodes) > len(n.sameNodes) {
-		n.logger.Debugf(ctx, "more duplicate nodes than same nodes, skipping update")
-		return Events{}
-	}
-
 	for key, nodeChange := range n.sameNodes {
 		nodeChange.f(key)
 		if nodeChange.eventType == EventNodeRemoved {
@@ -265,6 +258,5 @@ func (n *GroupNode) updateNodesFrom(ctx context.Context, events []etcdop.WatchEv
 		n.logger.Infof(ctx, nodeChange.message)
 	}
 
-	maps.Copy(n.duplicateNodes, n.sameNodes)
 	return out
 }
--- a/internal/pkg/service/common/distribution/node.go
+++ b/internal/pkg/service/common/distribution/node.go
@@ -3,6 +3,7 @@ package distribution
 import (
 	"context"
 	"fmt"
+	"maps"
 	"sync"
 	"time"
 
@@ -26,6 +27,11 @@ type Node struct {
 	groups       map[string]*GroupNode
 }
 
+type nodeChange struct {
+	eventType EventType
+	f         func(nodeID string)
+}
+
 // GroupNode is created within each node in the group.
 // One physical cluster node can be present in multiple independent distribution groups.
 //
@@ -41,6 +47,9 @@ type GroupNode struct {
 	client      *etcd.Client
 	groupPrefix etcdop.Prefix
 	listeners   *listeners
+
+	sameNodes      map[string]nodeChange
+	duplicateNodes map[string]nodeChange
 }
 
 type assigner = Assigner
@@ -87,11 +96,13 @@ func newGroupMember(nodeID, groupID string, cfg Config, d dependencies) (*GroupN
 
 	// Create instance
 	g := &GroupNode{
-		assigner:    newAssigner(nodeID),
-		logger:      d.Logger().WithComponent("distribution"),
-		config:      cfg,
-		client:      d.EtcdClient(),
-		groupPrefix: etcdop.NewPrefix(fmt.Sprintf("runtime/distribution/group/%s/nodes", groupID)),
+		assigner:       newAssigner(nodeID),
+		logger:         d.Logger().WithComponent("distribution"),
+		config:         cfg,
+		client:         d.EtcdClient(),
+		groupPrefix:    etcdop.NewPrefix(fmt.Sprintf("runtime/distribution/group/%s/nodes", groupID)),
+		sameNodes:      make(map[string]nodeChange),
+		duplicateNodes: make(map[string]nodeChange),
 	}
 
 	// Graceful shutdown
@@ -220,19 +231,39 @@ func (n *GroupNode) updateNodesFrom(ctx context.Context, events []etcdop.WatchEv
 		switch rawEvent.Type {
 		case etcdop.CreateEvent, etcdop.UpdateEvent:
 			nodeID := string(rawEvent.Kv.Value)
+			if _, ok := n.sameNodes[nodeID]; ok {
+				continue
+			}
 			event := Event{Type: EventNodeAdded, NodeID: nodeID, Message: fmt.Sprintf(`found a new node "%s"`, nodeID)}
 			out = append(out, event)
-			n.assigner.addNode(nodeID)
+			n.sameNodes[nodeID] = nodeChange{eventType: EventNodeAdded, f: n.assigner.addNode}
 			n.logger.Infof(ctx, event.Message)
 		case etcdop.DeleteEvent:
 			nodeID := string(rawEvent.PrevKv.Value)
+			if _, ok := n.sameNodes[nodeID]; !ok {
+				continue
+			}
 			event := Event{Type: EventNodeRemoved, NodeID: nodeID, Message: fmt.Sprintf(`the node "%s" gone`, nodeID)}
 			out = append(out, event)
-			n.assigner.removeNode(nodeID)
+			n.sameNodes[nodeID] = nodeChange{eventType: EventNodeRemoved, f: n.assigner.removeNode}
 			n.logger.Infof(ctx, event.Message)
 		default:
 			panic(errors.Errorf(`unexpected event type "%s"`, rawEvent.Type.String()))
 		}
 	}
+
+	// Make sure that nodes are updated only when there is maximum nodes
+	if len(n.duplicateNodes) > len(n.sameNodes) {
+		return Events{}
+	}
+
+	for key, nodeChange := range n.sameNodes {
+		nodeChange.f(key)
+		if nodeChange.eventType == EventNodeRemoved {
+			delete(n.sameNodes, key)
+		}
+	}
+
+	n.duplicateNodes = maps.Clone(n.sameNodes)
 	return out
 }
--- a/internal/pkg/service/common/distribution/node.go
+++ b/internal/pkg/service/common/distribution/node.go
@@ -30,6 +30,7 @@ type Node struct {
 type nodeChange struct {
 	eventType EventType
 	f         func(nodeID string)
+	message   string
 }
 
 // GroupNode is created within each node in the group.
@@ -236,8 +237,7 @@ func (n *GroupNode) updateNodesFrom(ctx context.Context, events []etcdop.WatchEv
 			}
 			event := Event{Type: EventNodeAdded, NodeID: nodeID, Message: fmt.Sprintf(`found a new node "%s"`, nodeID)}
 			out = append(out, event)
-			n.sameNodes[nodeID] = nodeChange{eventType: EventNodeAdded, f: n.assigner.addNode}
-			n.logger.Infof(ctx, event.Message)
+			n.sameNodes[nodeID] = nodeChange{eventType: EventNodeAdded, f: n.assigner.addNode, message: event.Message}
 		case etcdop.DeleteEvent:
 			nodeID := string(rawEvent.PrevKv.Value)
 			if _, ok := n.sameNodes[nodeID]; !ok {
@@ -245,8 +245,7 @@ func (n *GroupNode) updateNodesFrom(ctx context.Context, events []etcdop.WatchEv
 			}
 			event := Event{Type: EventNodeRemoved, NodeID: nodeID, Message: fmt.Sprintf(`the node "%s" gone`, nodeID)}
 			out = append(out, event)
-			n.sameNodes[nodeID] = nodeChange{eventType: EventNodeRemoved, f: n.assigner.removeNode}
-			n.logger.Infof(ctx, event.Message)
+			n.sameNodes[nodeID] = nodeChange{eventType: EventNodeRemoved, f: n.assigner.removeNode, message: event.Message}
 		default:
 			panic(errors.Errorf(`unexpected event type "%s"`, rawEvent.Type.String()))
 		}
@@ -254,6 +253,7 @@ func (n *GroupNode) updateNodesFrom(ctx context.Context, events []etcdop.WatchEv
 
 	// Make sure that nodes are updated only when there is maximum nodes
 	if len(n.duplicateNodes) > len(n.sameNodes) {
+		n.logger.Debugf(ctx, "more duplicate nodes than same nodes, skipping update")
 		return Events{}
 	}
 
@@ -262,8 +262,9 @@ func (n *GroupNode) updateNodesFrom(ctx context.Context, events []etcdop.WatchEv
 		if nodeChange.eventType == EventNodeRemoved {
 			delete(n.sameNodes, key)
 		}
+		n.logger.Infof(ctx, nodeChange.message)
 	}
 
-	n.duplicateNodes = maps.Clone(n.sameNodes)
+	maps.Copy(n.duplicateNodes, n.sameNodes)
 	return out
 }
--- a/internal/pkg/service/common/etcdop/watch.go
+++ b/internal/pkg/service/common/etcdop/watch.go
@@ -1,9 +1,7 @@
 package etcdop
 
 import (
-	"bytes"
 	"context"
-	"sort"
 	"time"
 
 	"go.etcd.io/etcd/api/v3/mvccpb"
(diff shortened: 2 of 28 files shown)


internal/pkg/service/common/distribution/node.go
```diff
@@ -220,19 +231,39 @@ func (n *GroupNode) updateNodesFrom(ctx context.Context, events []etcdop.WatchEv
...
+
+	// Make sure that nodes are updated only when there is maximum nodes
+	if len(n.duplicateNodes) > len(n.sameNodes) {
+		return Events{}
+	}
+
+	for key, nodeChange := range n.sameNodes {
+		nodeChange.f(key)
+		if nodeChange.eventType == EventNodeRemoved {
+			delete(n.sameNodes, key)
```
Matovidlo, line 263: We have to delete old nodes. Duplicates does not need to be handled as it is clone of the `sameNodes`.

internal/pkg/service/common/distribution/node.go
```diff
@@ -220,19 +231,39 @@ func (n *GroupNode) updateNodesFrom(ctx context.Context, events []etcdop.WatchEv
...
 		default:
 			panic(errors.Errorf(`unexpected event type "%s"`, rawEvent.Type.String()))
 		}
 	}
+
+	// Make sure that nodes are updated only when there is maximum nodes
+	if len(n.duplicateNodes) > len(n.sameNodes) {
+		return Events{}
+	}
+
```
Matovidlo, line 259: Miss the comparison of `map` keys `duplicate` vs `same`

internal/pkg/service/stream/storage/level/local/diskwriter/network/rpc/client.go
```diff
@@ -33,11 +35,31 @@ type networkFile struct {
...
+		conn, found := connections.ConnectionToVolume(sliceKey.VolumeID)
+		if !found || !conn.IsConnected() {
+			return nil, errors.Errorf("no connection to the volume %q", sliceKey.VolumeID.String())
+		}
+
+		ctx = ctxattr.ContextWith(
+			ctx,
+			attribute.String("writerNodeId", conn.RemoteNodeID()),
+			attribute.String("writerNodeAddress", conn.RemoteAddr()),
+		)
```
Matovidlo, line 61: This can be maybe moved to the previous location.
This `dialer` function is used with each changes in underlying connection.
Problem is when `conn` was received as argument and not cheked here.
The downtime of ETCD causes the connection to be already closed and this change effectivly asks volume for exact connection during dial up

internal/pkg/service/stream/storage/level/local/diskwriter/network/rpc/server.go
```diff
@@ -160,6 +160,7 @@ func (s *NetworkFileServer) KeepAliveStream(req *pb.KeepAliveStreamRequest, stre
 
 	case <-stream.Context().Done():
 		// The client is gone, remove reference for the writer, and do graceful close
+		s.logger.Infof(stream.Context(), "client is gone, file id %d", req.FileId)
```
Matovidlo:
```suggestion
```

internal/pkg/service/stream/storage/level/local/diskwriter/writer.go
```diff
@@ -146,6 +146,10 @@ func (w *writer) Write(ctx context.Context, aligned bool, p []byte) (n int, err
 }
 
 func (w *writer) Sync(ctx context.Context) error {
+	if w.isClosed() {
+		return nil
+	}
```
Matovidlo, line 151: `race` condition on `waitgroup`

internal/pkg/service/common/etcdop/watch.go
```diff
@@ -206,7 +206,14 @@ func (v Prefix) WatchWithoutRestart(ctx context.Context, client etcd.Watcher, op
 			}
 
 			// Map event type
+			var newRevision int64
 			for _, rawEvent := range rawResp.Events {
+				// Skip events that have modification revision lower than the last newly created revision
+				// This resolves delete after create issue.
+				if rawEvent.Kv.ModRevision < newRevision {
+					continue
+				}
```
Matovidlo, line 215: This caused delete after create events to be performed. Checking revision skips the events.
I am wondering how it is possible for ETCD to emit such event from `Watch` method :thinking:

michaljurecko: @Matovidlo 

Note from a passer-by:
- `ModRevision` is pseudo-timestamp of the last modification of **the key**
- `CreateRevision` s pseudo-timestamp of **the key** creation

You are watching a prefix, **so here you will have multiple/different keys**, in the for loop.

I don't think you should compare Create/Mod revisions of **different keys**:
- I don't know what you're trying to do, but this doesn't look valid.

----------------------

https://etcd.io/docs/v3.5/learning/api_guarantees/#watch-apis

```
Watches make guarantees about events:
Ordered - events are ordered by revision. An event will never appear on a watch if it precedes an event in time that has already been posted.
....
```
--------

So, in a watch stream, there should be never a modification from sooner revision after a later modification.

Situation may be more complicated on a stream restart (for example a network outage occurred):
- There is no effort implemented to continue where it left off.
  - I don't think it would be possible to make it 100% bulletproof.
- Instead, on stream reset, the `reset=true` flag is emitted.
- You should reset also all local in-memory states (start from the beginning).
- The `mirror` utils implements it for example here:

https://github.com/keboola/keboola-as-code/blob/abd2d93b017c2d15af5362c2e89785fc1121c7e2/internal/pkg/service/common/etcdop/watch_mirror_map.go#L84-L87

--------------

If you would like some _free_ consultation/discussion, write to me.

My motivation is a deep interest in etcd and watch area.

--------------

BTW, in the code above is 

`SORT: 1. By type, PUT (CREATE/UPDATE) before DELETE`, I don't know if this is related to your issue.

- IMHO there shouldn't be both, create and update event, for the same key, in the same events batch.
- But maybe I am wrong, and the sort should be removed.

Matovidlo: > I don't think you should compare Create/Mod revisions of **different keys**:
> - I don't know what you're trying to do, but this doesn't look valid.

I understand that, I will check on that one more time but this change made a significant improvement on 2-3 issues that we were facing for about 4 months.

> `SORT: 1. By type, PUT (CREATE/UPDATE) before DELETE`, I don't know if this is related to your issue.

It seems to be the issue. I would add there another sorting options to be sorted by the `ModRevision` and `CreateRevision` respectivelly. E.g newer event is after older event. Sort of priority queue based on the `revision`.

> So, in a watch stream, there should be never a modification from sooner revision after a later modification.

That's nice but the actual restart of ETCD (probablly leader election happend again) causes that watch return the events out of the order or the sort mentioned above makes this issue. 

>- Instead, on stream reset, the `reset=true` flag is emitted.

This is never emited after outage.

> If you would like some free consultation/discussion, write to me.

I'll write you :+1:

internal/pkg/service/common/distribution/node.go
```diff
@@ -220,19 +231,39 @@ func (n *GroupNode) updateNodesFrom(ctx context.Context, events []etcdop.WatchEv
...
 			n.logger.Infof(ctx, event.Message)
 		default:
 			panic(errors.Errorf(`unexpected event type "%s"`, rawEvent.Type.String()))
 		}
 	}
+
+	// Make sure that nodes are updated only when there is maximum nodes
+	if len(n.duplicateNodes) > len(n.sameNodes) {
+		return Events{}
+	}
```
jachym-tousek-keboola, line 258: Okay I don't understand this. If I understand correctly from the call you want to suppress the events if a node was both deleted and added, right?

Isn't this too aggressive then? It looks like you suppress all events if there are fewer of them than previously?

internal/pkg/service/common/distribution/node.go
jachym-tousek-keboola: Why are the IDs not important?

internal/pkg/service/common/distribution/node.go
```diff
@@ -87,11 +96,13 @@ func newGroupMember(nodeID, groupID string, cfg Config, d dependencies) (*GroupN
...
-		config:      cfg,
-		client:      d.EtcdClient(),
-		groupPrefix: etcdop.NewPrefix(fmt.Sprintf("runtime/distribution/group/%s/nodes", groupID)),
+		assigner:       newAssigner(nodeID),
+		logger:         d.Logger().WithComponent("distribution"),
+		config:         cfg,
+		client:         d.EtcdClient(),
+		groupPrefix:    etcdop.NewPrefix(fmt.Sprintf("runtime/distribution/group/%s/nodes", groupID)),
+		sameNodes:      make(map[string]nodeChange),
+		duplicateNodes: make(map[string]nodeChange),
```
jachym-tousek-keboola, line 105: Can we improve `distribution/node_test.go` to test these changes?

internal/pkg/service/stream/storage/level/local/diskwriter/network/connection/connection.go
```diff
@@ -143,17 +147,27 @@ func (m *Manager) ConnectionsCount() int {
 }
 
 func (m *Manager) updateConnections(ctx context.Context) {
-	m.logger.Infof(ctx, `the list of volumes has changed, updating connections`)
-
 	activeNodes := m.writerNodes()
+	if len(m.oldVolumes) > len(activeNodes) {
+		activeNodes = m.oldVolumes
+	}
```
jachym-tousek-keboola, line 153: Similarly as the distribution, I don't understand this. I would expect `!=` but `>` is surprising.

internal/pkg/service/stream/storage/level/local/diskwriter/network/connection/connection.go
```diff
@@ -143,17 +147,27 @@ func (m *Manager) ConnectionsCount() int {
...
 
+	if maps.Equal(m.oldVolumes, activeNodes) {
+		m.logger.Infof(ctx, "no changes in volumes, skipping connection update")
+		return
```
jachym-tousek-keboola, line 157: I'd expect some new test that would expect this log line.

internal/pkg/service/stream/storage/level/local/diskwriter/network/rpc/client.go
```diff
@@ -121,6 +144,11 @@ func OpenNetworkFile(ctx context.Context, logger log.Logger, telemetry telemetry
 			}
 
 			if s, ok := status.FromError(err); !ok || s.Code() != codes.Canceled {
+				if s.Code() == codes.Unavailable && errors.Is(s.Err(), io.EOF) {
+					onServerTermination(ctx, "remote server shutdown")
+					return
+				}
```
jachym-tousek-keboola, line 150: Can you explain this block please?

internal/pkg/service/stream/storage/level/local/diskwriter/writer.go
```diff
@@ -159,14 +163,13 @@ func (w *writer) Close(ctx context.Context) error {
...
+	errs := errors.NewMultiError()
 	if w.isClosed() {
 		return errors.New(`writer is already closed`)
 	}
-	close(w.closed)
-
-	errs := errors.NewMultiError()
 
 	// Wait for running writes
+	close(w.closed)
```
jachym-tousek-keboola, line 172: Looking closer at this, I think these changes don't do anything? Why would it matter when exactly is `errs := errors.NewMultiError()` created? You aren't using it sooner or anything... I'd recommend ammending the commit to remove this change as it's just confusing but doesn't do anything as far as I can tell.

jachym-tousek-keboola (changes requested):
Overall I feel like this PR is too big. Not code wise, each of the changes are relatively small. But they're difficult to understand and not very related so it gets a bit confusing when they're mixed up in one PR like this. It would be better if this PR was split in my opinion so that we can discuss each fix separately.

internal/pkg/service/common/distribution/node.go
Matovidlo: Because they are present in map[`key`].

internal/pkg/service/stream/storage/level/local/diskwriter/network/connection/connection.go
Matovidlo: I would like to encourage you to test it on your behalf instead of asking.
Simply the operations in watchMirror are streamlined 1 by 1. So if I restart a watch it goes like this 
```
DEL vol1 modRev 3
DEL vol2 modeRev 4
PUT vol1 modRev 5
PUT vol2 modRev 6
```
So based on your comparison I would effectively go through rest of code.
This code has one issue and it is in downsizing the number of volumes. E.g 2->1. It requires a restart. I think it cannot be safelly done only by restarting whole deployment.

internal/pkg/service/stream/storage/level/local/diskwriter/network/connection/connection.go
Matovidlo: Please help mi with that and adjust the test based on this branch. I have 4 more issues on table

internal/pkg/service/stream/storage/level/local/diskwriter/network/rpc/client.go
Matovidlo: I think it is better to go throughout how gRPC and network things works. I think that's where we still have problem in streaming at fast pace and this will be probably reworked if we want to achieve higher throughput

internal/pkg/service/stream/storage/level/local/diskwriter/writer.go
Matovidlo: I am more convinient with variable declaration at top of method instead of doing that throughout the function.

Matovidlo:
@jachym-tousek-keboola I understand your concert about PR but this PR fixes 4 related things on Disk Writer communication. It is a set of maybe unrelated things but left overs that were not tested. I think that there is no real reason to split this because all `4` issues have to be address otherwise the disk writer is not working properly.

internal/pkg/service/common/distribution/node.go
Matovidlo: The issue is only when you downsize the cluster e.g from 4 nodes -> 2 nodes. It requires restart of all services. It cannot be done otherwise. I will try to write a test that shows that but cannot do it with any other approach.
I think this is very important topic that you need to understand in depth (ETCD mirror, watch operation and restarts of ETCD) otherwise the PRs and CRs will not make sense.

Matovidlo: 8edb62e5068dc008578004fee1b364820a623e46

internal/pkg/service/common/distribution/node.go
Matovidlo: 8edb62e5068dc008578004fee1b364820a623e46

internal/pkg/service/stream/storage/level/local/diskwriter/writer.go
jachym-tousek-keboola: I'm not since this way the variable is declared even when not needed (when the writer is closed).

internal/pkg/service/stream/storage/level/local/diskwriter/network/connection/connection.go
jachym-tousek-keboola: > I would like to encourage you to test it on your behalf instead of asking.

I don't know what the scenario is.

Anyway, I have been looking at the code for a while and comparing it with the old code and your changes still don't make any sense to me... We have a mirrored map of volumes and when they change we need to open connections to the new volumes and close the connections to the old volumes. right? At least that's what the old code did.

Now if I understand the code correctly `updateConnections` is called on every revision, right? And the goal is to prevent disconnecting a node if it is readded soon after? And it also looks like your code assumes that each revision has a maximum of one volume change? How is that guaranteed?

jachym-tousek-keboola: Recap from calls: The relevant deletes and puts seem to be within a single revision. The code in `distribution/node.go` processed every event independently so it wasn't able to catch no-ops when a DELETE and PUT appeared together. In `connection.go` however this is already the case due to how MirrorMap works internally - the `WithOnUpdate` callback is only called once at the end of the revision instead of for each event independently. This is evident when comparing `WithForEach` and `for ... range events {` in `node.go` vs `watch_mirror_map.go`. Because of that I believe that we actually only need the distribution changes but not the connection changes.

Martin said that the distribution changes alone were not sufficient though so he'll repeat the testing scenario to gather more information what was wrong in that case.

PR #2230, opened 2025-01-31
WIP

hosekpeter:
Jira: XXXXX

**Changes:**
- ...

-----------


internal/pkg/service/common/etcdop/watch_mirror_map.go
```diff
@@ -248,3 +261,41 @@ func (m *MirrorMap[T, K, V]) ForEach(fn func(K, V) (stop bool)) {
...
 	}
 }
+
+func (m *MirrorMap[T, K, V]) recordTelemetry(ctx context.Context, tel telemetry.Telemetry) {
+	tel.Meter().
+		IntCounter(
+			"keboola.go.mirror.map.num.keys",
+			"Number of keys in the mirror map.",
+			"count",
+		).
```
jachym-tousek-keboola, line 271: You shoud save this to a variable / property in order to avoid calling it in every iteration.

internal/pkg/service/common/etcdop/watch_mirror_map.go
```diff
@@ -248,3 +261,41 @@ func (m *MirrorMap[T, K, V]) ForEach(fn func(K, V) (stop bool)) {
...
+	}
+
+	// Emit metrics for the current map memory
+	meter := tel.Meter()
+
+	// Track memory consumed by the map itself
+	meter.IntCounter(
+		"keboola.go.mirror.map.memory.usage.bytes",
+		"Memory consumed by the MirrorMap, including keys and values.",
+		"bytes",
```
jachym-tousek-keboola, line 299: Same here.

internal/pkg/service/common/etcdop/watch_mirror_map.go
jachym-tousek-keboola, line 271: You shoud save this to a variable / property in order to avoid calling it in every iteration. Same goes for `recordMemoryTelemetry` and `MirrorTree`.

jachym-tousek-keboola (changes requested):

internal/pkg/utils/memory/memory.go
```diff
@@ -0,0 +1,67 @@
...
+	"reflect"
+	"unsafe"
+)
+
+func Size(v interface{}) uintptr {
+	return calculateSize(reflect.ValueOf(v), make(map[uintptr]bool))
+}
+
+// calculateSize calculates the memory size of a given value, including nested or referenced elements.
+func calculateSize(v reflect.Value, visited map[uintptr]bool) uintptr {
```
jachym-tousek-keboola, line 13: How did you come up with this? Does some library exist for this?

internal/pkg/service/common/etcdop/watch_mirror_map.go
```diff
@@ -248,3 +261,41 @@ func (m *MirrorMap[T, K, V]) ForEach(fn func(K, V) (stop bool)) {
...
+	// Variable to track memory consumed
+	var memoryConsumed int64
+
+	// Measure base memory usage of the map structure
+	memoryConsumed += int64(unsafe.Sizeof(*m))
+
+	// Lock the map to measure memory usage of its elements
+	for k, v := range m.mapData {
+		memoryConsumed += int64(unsafe.Sizeof(k)) // Add key size
+		memoryConsumed += int64(memory.Size(v))   // Add value size
```
jachym-tousek-keboola, line 289: Bit worried about performance. MirrorMap and MirrorTree are fairly busy and this could slow them down.

internal/pkg/service/common/etcdop/watch_mirror_map.go
```diff
@@ -20,6 +24,7 @@ import (
 //
 // MirrorMap is ideal for quick single key access or iteration over all keys.
 // To get all keys from a common prefix, use MirrorTree instead.
+
```
jachym-tousek-keboola, line 27:
```suggestion
```

internal/pkg/service/common/etcdop/watch_mirror_map.go
hosekpeter: it will be solved in this [PR](https://github.com/keboola/keboola-as-code/pull/2235). I set a 30 seconds ticker for testing, then it won't called in every iteration.

internal/pkg/utils/memory/memory.go
hosekpeter: I found out that the original golang is `unsafe`. I used it, but it returned some strange values, so following Martin's recommendation I used `reflect`. I didn't look for it in any library.

internal/pkg/service/common/etcdop/watch_mirror_map.go
Matovidlo: I think once we introduce that this is calculated once per e.g 5 minutes it think that's not such issue. We want to measure it to make sure that the representation that we save is efficient and consuming adequate memory.

internal/pkg/service/common/etcdop/watch_mirror_map_test.go
```diff
@@ -321,7 +331,7 @@ func TestMirrorMap_WithOnChanges(t *testing.T) {
 		Deleted: []MirrorKVPair[testUserFullName, int]{
 			{
 				Key:   "Paul Green",
-				Value: 20,
+				Value: (20),
```
Matovidlo, line 334: What are the brackets for?

internal/pkg/utils/memory/memory.go
Matovidlo: https://github.com/DmitriyVTitov/size
Long description here golang.org/x/tools/go/pointer

Why would we use that `size` library? I don't know it is finished thing, I would rather put this into utils instead of sending PRs. On the other hand we want the measurement to be approximative, not exact, which this utils functions does.

Matovidlo (approved):
LGTM

internal/pkg/utils/memory/memory.go
jachym-tousek-keboola: ok

internal/pkg/service/common/etcdop/watch_mirror_map.go
jachym-tousek-keboola: Even then I'd still store the meter somewhere if you don't mind. Similar to https://github.com/keboola/keboola-as-code/blob/20013c8973ab940023891387a210dbe9f5f36e8a/internal/pkg/service/stream/storage/node/trace.go#L23-L24.

Matovidlo: It would be nice but I we do not have `watch.go` module but `etcdop` and I am not sure whether all telemetry should be under single root. It will be bunch of unrelated telemetry stuff for each etcdop thing.
I would think through in another PR whether we would like to split `watch` module to the separate package

internal/pkg/service/common/etcdop/watch_mirror_map_test.go
hosekpeter: i don't know, it a relic in from the past :)

jachym-tousek-keboola (approved):

Matovidlo (approved):
LGTM

PR #2231, opened 2025-01-31
feat: GitLab provider

jachym-tousek-keboola:
Jira: https://keboola.atlassian.net/browse/PSGO-991

**Changes:**
- Added GitLab provider for apps proxy

-----------

### DIFF
--- a/internal/pkg/service/appsproxy/dataapps/auth/provider/gitlab.go
+++ b/internal/pkg/service/appsproxy/dataapps/auth/provider/gitlab.go
@@ -1,0 +1,32 @@
+package provider
+
+import proxyOptions "github.com/oauth2-proxy/oauth2-proxy/v7/pkg/apis/options"
+
+type GitLab struct {
+	OIDC
+	Groups   []string `json:"groups"`
+	Projects []string `json:"projects"`
+}
+
+func (v GitLab) ProxyProviderOptions() (proxyOptions.Provider, error) {
+	p, err := v.OIDC.ProxyProviderOptions()
+	if err != nil {
+		return proxyOptions.Provider{}, err
+	}
+
+	p.Type = proxyOptions.GitLabProvider
+	p.LoginURLParameters = []proxyOptions.LoginURLParameter{
+		{
+			// GitLab doesn't support "select_account" prompt
+			// https://gitlab.com/gitlab-org/gitlab/-/issues/377368
+			Name:    "prompt",
+			Default: []string{"login"},
+		},
+	}
+	p.GitLabConfig = proxyOptions.GitLabOptions{
+		Group:    v.Groups,
+		Projects: v.Projects,
+	}
+
+	return p, nil
+}
--- a/internal/pkg/service/appsproxy/dataapps/auth/provider/gitlab_test.go
+++ b/internal/pkg/service/appsproxy/dataapps/auth/provider/gitlab_test.go
@@ -1,0 +1,86 @@
+package provider
+
+import (
+	"encoding/json"
+	"testing"
+
+	proxyOptions "github.com/oauth2-proxy/oauth2-proxy/v7/pkg/apis/options"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestGitLab(t *testing.T) {
+	t.Parallel()
+
+	// Mock part of the API response
+	providerJSON := `
+{
+  "id": "my-id",
+  "name": "My Name",
+  "type": "gitlab",
+  "clientId": "6779ef20e75817b79602",
+  "clientSecret": "f2a1ed52710d4533bde25be6da03b6e3",
+  "issuerUrl": "https://gitlab.com",
+  "allowedRoles": ["admin"],
+  "groups": ["group1", "group2"],
+  "projects": ["project1", "project2"]
+}
+`
+
+	// Unmarshal, detect the target struct
+	var providers Providers
+	require.NoError(t, json.Unmarshal([]byte("["+providerJSON+"]"), &providers))
+	require.Len(t, providers, 1)
+
+	// Decoded content
+	provider := providers[0]
+	assert.Equal(t, GitLab{
+		OIDC: OIDC{
+			Base: Base{
+				Info: Info{
+					ID:   "my-id",
+					Name: "My Name",
+					Type: TypeGitLab,
+				},
+			},
+			ClientID:     "6779ef20e75817b79602",
+			ClientSecret: "f2a1ed52710d4533bde25be6da03b6e3",
+			IssuerURL:    "https://gitlab.com",
+			AllowedRoles: &[]string{"admin"},
+		},
+		Groups:   []string{"group1", "group2"},
+		Projects: []string{"project1", "project2"},
+	}, provider)
+
+	// OAuth2Proxy configuration
+	oAuth2ProxyProvider, ok := provider.(GitLab)
+	require.True(t, ok)
+	proxyOpts, err := oAuth2ProxyProvider.ProxyProviderOptions()
+	require.NoError(t, err)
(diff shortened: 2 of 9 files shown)


internal/pkg/service/appsproxy/dataapps/auth/provider/gitlab.go
```diff
@@ -0,0 +1,32 @@
+package provider
+
+import proxyOptions "github.com/oauth2-proxy/oauth2-proxy/v7/pkg/apis/options"
+
+type GitLab struct {
+	OIDC
+	Groups   []string `json:"groups"`
+	Projects []string `json:"projects"`
+}
```
jachym-tousek-keboola, line 9: Separate provider struct in order to support the gitlab-specific options.

internal/pkg/service/appsproxy/dataapps/auth/provider/gitlab.go
```diff
@@ -0,0 +1,32 @@
...
+
+	p.Type = proxyOptions.GitLabProvider
+	p.LoginURLParameters = []proxyOptions.LoginURLParameter{
+		{
+			// GitLab doesn't support "select_account" prompt
+			// https://gitlab.com/gitlab-org/gitlab/-/issues/377368
+			Name:    "prompt",
+			Default: []string{"login"},
+		},
+	}
```
jachym-tousek-keboola, line 25: This was the main issue. Our oidc provider adds prompt=select_account which GitLab doesn't support and says the configuration is invalid.

internal/pkg/service/appsproxy/dataapps/auth/provider/gitlab_test.go
```diff
@@ -0,0 +1,86 @@
...
+	assert.Equal(t, proxyOptions.Provider{
+		ID:                  "my-id",
+		Type:                "gitlab",
+		Name:                "My Name",
+		CodeChallengeMethod: "S256",
+		ClientID:            "6779ef20e75817b79602",
+		ClientSecret:        "f2a1ed52710d4533bde25be6da03b6e3",
+		AllowedGroups:       []string{"admin"},
+		OIDCConfig: proxyOptions.OIDCOptions{
+			IssuerURL:      "https://gitlab.com",
```
jachym-tousek-keboola, line 69: The issuer url is just `https://gitlab.com`, oauth2-proxy adds the `/.well-known/openid-configuration` on its own. Note that it also cannot contain the trailing slash - `https://gitlab.com/` is invalid.

internal/pkg/service/appsproxy/dataapps/auth/provider/oidc.go
```diff
@@ -7,6 +7,13 @@ import (
 	"github.com/keboola/keboola-as-code/internal/pkg/utils/errors"
 )
 
+type OIDCProvider interface {
```
jachym-tousek-keboola, line 10: New interface for the OIDC providers.

internal/pkg/service/appsproxy/proxy/apphandler/authproxy/manager.go
```diff
@@ -49,7 +49,8 @@ func (m *Manager) NewHandlers(app api.AppConfig, upstream chain.Handler) map[pro
 		switch p := auth.(type) {
 		case provider.OIDC:
 			authHandlers[auth.ID()] = oidcproxy.NewHandler(m.logger, m.config, m.providerSelector, m.pageWriter, app, p, upstream)
-
+		case provider.GitLab:
+			authHandlers[auth.ID()] = oidcproxy.NewHandler(m.logger, m.config, m.providerSelector, m.pageWriter, app, p, upstream)
```
jachym-tousek-keboola, line 53: Can't combine these into one case even though the code is the same.

Matovidlo: Why? Just move the `provider.GitLab` under `provider.OIDC` :thinking:

internal/pkg/service/appsproxy/dataapps/auth/provider/gitlab_test.go
```diff
@@ -0,0 +1,86 @@
...
+	t.Parallel()
+
+	// Mock part of the API response
+	providerJSON := `
+{
+  "id": "my-id",
+  "name": "My Name",
+  "type": "gitlab",
+  "clientId": "6779ef20e75817b79602",
+  "clientSecret": "f2a1ed52710d4533bde25be6da03b6e3",
```
Matovidlo, line 22: Hopefully this is just an example

internal/pkg/service/appsproxy/dataapps/auth/provider/gitlab_test.go
Matovidlo: I know about this. But we do not want to adjust the forked oauth2 proxy library to prevent future issues with receiving updates from upstream

internal/pkg/service/appsproxy/proxy/apphandler/authproxy/manager.go
jachym-tousek-keboola: I don't know. It doesn't pass.

internal/pkg/service/appsproxy/dataapps/auth/provider/gitlab_test.go
jachym-tousek-keboola: yup, I just generated a random string, real gitlab secrets have a prefix

internal/pkg/service/appsproxy/dataapps/auth/provider/gitlab_test.go
jachym-tousek-keboola: We could in theory have the proxy cut it without adjusting the fork but I'm unsure whether it's safe. I can't say for sure that some other provider won't work the other way and require the trailing slash.

internal/pkg/service/appsproxy/proxy/apphandler/authproxy/manager.go
jachym-tousek-keboola: https://github.com/keboola/keboola-as-code/actions/runs/13109459311/job/36570014284

jachym-tousek-keboola: https://go.dev/ref/spec#Type_switches

> The TypeSwitchGuard may include a [short variable declaration](https://go.dev/ref/spec#Short_variable_declarations). When that form is used, the variable is declared at the end of the TypeSwitchCase in the [implicit block](https://go.dev/ref/spec#Blocks) of each clause. In clauses with a case listing exactly one type, the variable has that type; otherwise, the variable has the type of the expression in the TypeSwitchGuard.

jachym-tousek-keboola: But wait, maybe if I do it like this? https://github.com/keboola/keboola-as-code/commit/d5524742451b4af896bdda6b712936e84ca67449

jachym-tousek-keboola: Okay that works. :)

internal/pkg/service/appsproxy/dataapps/auth/provider/gitlab_test.go
Matovidlo: It will be in the docs so I don't think that we need to adjust it if someone goes step by step by the documentation

internal/pkg/service/appsproxy/proxy/apphandler/authproxy/manager.go
Matovidlo: It's because it does not implement the interface. The type switch need the interface type to get proper Handlers :)

Matovidlo (approved):
LGTM

hosekpeter (approved):
LGTM

PR #2234, opened 2025-02-05
feat: Add backup file writer

Matovidlo:
Jira: [PSGO-860](https://keboola.atlassian.net/browse/PSGO-860), [PSGO-1015](https://keboola.atlassian.net/browse/PSGO-1015)

**Changes:**
- Add backup file writer.
- Move file from tmp to regular file before uploading phase.
- Once the restart of program happens and the file is corrupted, do not change it to import that and leave it as hidden.

-----------

[PSGO-860]: https://keboola.atlassian.net/browse/PSGO-860?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[PSGO-1015]: https://keboola.atlassian.net/browse/PSGO-1015?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


internal/pkg/service/stream/storage/level/local/diskwriter/file.go
```diff
@@ -3,13 +3,31 @@ package diskwriter
...
+
+	"github.com/keboola/keboola-as-code/internal/pkg/service/stream/storage/level/local/model"
 )
 
 const (
 	sliceFileFlags = os.O_CREATE | os.O_WRONLY
 	sliceFilePerm  = 0o640
 )
 
+func NewFileWithBackupFile(
```
hosekpeter, line 15: Why doesn't this function return error?

internal/pkg/service/stream/storage/level/local/diskwriter/file.go
```diff
@@ -3,13 +3,31 @@ package diskwriter
...
+	backupPath := slice.FileNameWithBackup(volumePath, sourceNodeID)
+	backupFile, err := os.OpenFile(backupPath, os.O_CREATE|os.O_RDWR, 0o640)
+	if err == nil {
+		backupFile.Close()
+		return backupPath
+	}
+
+	filePath := slice.FileName(volumePath, sourceNodeID)
+	return filePath
+}
```
hosekpeter, line 29:
```suggestion
	if err != nil {
		filePath := slice.FileName(volumePath, sourceNodeID)
	        return filePath
	}

	defer backupFile.Close()
	return backupPath
}
```

internal/pkg/service/stream/storage/level/local/diskreader/reader.go
```diff
@@ -239,3 +305,24 @@ func openFileAndWrite(ctx context.Context, logger log.Logger, opener FileOpener,
...
 	}
 }
+
+func checkGZIP(
+	localCompression compression.Config,
+	file File,
+	filePath string,
+) (err error) {
+	// Try to decompress with local compression
+	reader, err := compressionReader.New(file, localCompression)
```
jachym-tousek-keboola, line 315: This looks like it just assumes the file is compressed when the backup file is used? I didn't see any condition for the file actually being compressed.

Since the configuration is separate rather than tied to compression it may not be.

Additionally `checkGZIP` doesn't feel correct, there is no condition to check whether we're using gzip or a different compression here. And this code probably works regardless of compression algorithm.

internal/pkg/service/stream/storage/level/local/diskreader/reader.go
```diff
@@ -229,6 +238,63 @@ func openFileAndWrite(ctx context.Context, logger log.Logger, opener FileOpener,
 		return
 	}
 
+	// Check if the file is hidden (has "." prefix)
+	if strings.HasPrefix(path.Base(filePath), ".") {
```
jachym-tousek-keboola, line 242: Can you try to extract the block inside this if (not the if itself) into a separate function? I think it could simplify the error handling.

jachym-tousek-keboola (changes requested):

internal/pkg/service/stream/storage/level/local/diskreader/reader_test.go
```diff
@@ -273,14 +276,20 @@ func (tc *compressionTestCase) TestReadError(t *testing.T) {
 		return f, nil
 	})
 
-	// Create reader
+	// Create reader, which inspects the file when compressed
+	// Make sure that the file is compressed correctly
 	r, err := rtc.NewReader(tc.DisableValidation)
+	if rtc.Config.UseBackupReader && err != nil {
+		require.Contains(t, err.Error(), readError.Error())
```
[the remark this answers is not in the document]
Matovidlo: In new implementation we do `Read()` function within constructing NewReader. It now throws readError. In previous implementation it just constructed the instance and the error was never given. When we use backup reader we read the file, otherwise the function should work as before

internal/pkg/service/stream/storage/level/local/diskreader/reader.go
Matovidlo: It looks like that but default reader for non compressed file is io.NopCloser
https://github.com/keboola/keboola-as-code/blob/e5aa3393351cf2dfed5e0c520c0c2aedf389a8ce/internal/pkg/service/stream/storage/level/local/encoding/compression/reader/reader.go#L21

I would only find more convinient to call this function `canBeRead` on slice. Yep the code will work regarled compression even though we use compressionReader package

Matovidlo: 3b067c5139b3229d6d7b0f2c50140134a2e759ac

internal/pkg/service/stream/storage/level/local/diskreader/reader.go
Matovidlo: a13850e77f97828c4c856cc47795b94d75b7a8f7

internal/pkg/service/stream/storage/level/local/diskwriter/writer.go
```diff
@@ -80,7 +81,16 @@ func newWriter(
...
 	// Open file
+
 	filePath := slice.FileName(volumePath, w.writerKey.SourceNodeID)
+	if withBackup {
+		filePath = NewFileWithBackupFile(
+			slice,
+			volumePath,
+			w.writerKey.SourceNodeID,
+		)
+	}
```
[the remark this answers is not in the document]
Matovidlo: 63d49656cf4469422ae8a5aaa39ff921c5340703

internal/pkg/service/stream/storage/level/local/diskwriter/file.go
Matovidlo: 63d49656cf4469422ae8a5aaa39ff921c5340703

internal/pkg/service/stream/storage/level/local/diskwriter/file.go
Matovidlo: 63d49656cf4469422ae8a5aaa39ff921c5340703

internal/pkg/service/stream/storage/level/local/diskreader/reader.go
```diff
@@ -221,21 +230,118 @@ func openFileAndWrite(ctx context.Context, logger log.Logger, opener FileOpener,
...
+
+	// Reopen file with new path
+	logger.Debugf(ctx, `moved hidden file "%s" to "%s"`, filePath, visiblePath)
+	newFile, err := opener.OpenFile(visiblePath)
+	if err != nil {
+		logger.Errorf(ctx, `cannot open file "%s": %s`, visiblePath, err)
+		return nil, "", err
+	}
+
+	return newFile, visiblePath, nil
```
jachym-tousek-keboola, line 302: Wdyt about moving this block outside of this function in order to avoid returning three values? It feels a bit odd to return both the path and the file handle. I don't think it's necessary.

jachym-tousek-keboola (approved):

internal/pkg/service/stream/storage/level/local/diskreader/reader.go
Matovidlo: 2e89529253af090e25bd3c0a578e9f494f129e34

hosekpeter (approved):

PR #2236, opened 2025-02-06
test: Add test for limitbuffer

jachym-tousek-keboola:
Jira: https://keboola.atlassian.net/browse/PSGO-683

**Changes:**
- Added missing tests for limitbuffer
- Fixed minor limitbuffer issues

-----------

### DIFF
--- a/internal/pkg/service/stream/storage/level/local/encoding/limitbuffer/buffer_test.go
+++ b/internal/pkg/service/stream/storage/level/local/encoding/limitbuffer/buffer_test.go
@@ -1,0 +1,113 @@
+package limitbuffer
+
+import (
+	"bytes"
+	"sync"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestBuffer_SmallChunks(t *testing.T) {
+	t.Parallel()
+
+	var out bytes.Buffer
+	maxSize := 10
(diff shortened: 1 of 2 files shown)


internal/pkg/service/stream/storage/level/local/encoding/limitbuffer/buffer_test.go
```diff
@@ -0,0 +1,113 @@
...
+
+	var out bytes.Buffer
+	maxSize := 10
+	buf := New(&out, maxSize)
+
+	var wg sync.WaitGroup
+	numGoroutines := 10
+	data := []byte("12345")
+
+	for range numGoroutines {
```
Matovidlo, line 97: Why do you do it like this? Each write in separate go routine?
It can be without go routine and wou will get same results because you have `wg.Wait()` right after.

jachym-tousek-keboola: It's in order to test that the lock works.

hosekpeter (approved):
LGTM

internal/pkg/service/stream/storage/level/local/encoding/limitbuffer/buffer_test.go
Matovidlo: Ok I did not know that you wanted to test that the limit buffer is concurrently safe.

PR #2237, opened 2025-02-08
Add configurable telemetry interval for data collection

hosekpeter:
Jira: XXXXX

**Changes:**
- Introduced a feature to make the telemetry interval configurable, allowing users to adjust the frequency of data collection as per their needs.
- Added corresponding configuration options to support this feature.


Matovidlo (approved):
LGTM

internal/pkg/service/stream/components.go
```diff
@@ -36,7 +36,6 @@ func StartComponents(ctx context.Context, serviceScp dependencies.ServiceScope,
 	for _, c := range components {
 		componentsMap[c] = true
 	}
-
```
jachym-tousek-keboola, line 39: revert

jachym-tousek-keboola (approved):
I dislike the cases where you removed blank lines like the one below but otherwise it's fine.

internal/pkg/service/stream/components.go
hosekpeter: [b362e3e](https://github.com/keboola/keboola-as-code/pull/2237/commits/b362e3ea77d664f7c2766b200592525cdf786c97)

hosekpeter: just check :)

jachym-tousek-keboola: :+1:

PR #2238, opened 2025-02-08
feat: Open file processing chunks

Matovidlo:
Jira: [PSGO-993](https://keboola.atlassian.net/browse/PSGO-993)

**Changes:**
- Fix deadlock within pipeline creation.
- Fix issue with pipeline "duplicates" creation.
- Open network file when closed during processing chunks.
- Delete after write does not happens in `watch` module.
  - Diskwriter seems to be working correctly. [PSGO-993](https://keboola.atlassian.net/browse/PSGO-993)

Issues
- Removed sorting in ETCD transaction/compact operation to prevent Delete after Create, test included.
- Open connection in dialer section. Should prevent issues that `connection.volume` has no old reference but always up to date reference from set of connection volumes  
- One small race condition on waitgroup. Use w.file.Sync() instead of Sync() method in Close as waitgroup is not longer needed during close of writer
-----------

[PSGO-993]: https://keboola.atlassian.net/browse/PSGO-993?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[PSGO-810]: https://keboola.atlassian.net/browse/PSGO-810?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[PSGO-830]: https://keboola.atlassian.net/browse/PSGO-830?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/internal/pkg/service/stream/storage/level/local/diskwriter/network/router/pipeline/pipeline_sink.go
+++ b/internal/pkg/service/stream/storage/level/local/diskwriter/network/router/pipeline/pipeline_sink.go
@@ -93,7 +93,6 @@ func (p *SinkPipeline) ReopenOnSinkModification() bool {
 }
 
 func (p *SinkPipeline) WriteRecord(c recordctx.Context) (pipelinePkg.WriteResult, error) {
-
 	// Make a local copy of pipelines while holding the read lock
 	p.writeLock.RLock()
 	pipelines := make([]balancer.SlicePipeline, len(p.pipelines))
--- a/internal/pkg/service/stream/storage/level/local/diskwriter/network/router/pipeline/pipeline_sink.go
+++ b/internal/pkg/service/stream/storage/level/local/diskwriter/network/router/pipeline/pipeline_sink.go
@@ -93,9 +93,14 @@ func (p *SinkPipeline) ReopenOnSinkModification() bool {
 }
 
 func (p *SinkPipeline) WriteRecord(c recordctx.Context) (pipelinePkg.WriteResult, error) {
+
+	// Make a local copy of pipelines while holding the read lock
 	p.writeLock.RLock()
-	defer p.writeLock.RUnlock()
-	return p.balancer.WriteRecord(c, p.pipelines)
+	pipelines := make([]balancer.SlicePipeline, len(p.pipelines))
+	copy(pipelines, p.pipelines)
+	p.writeLock.RUnlock()
+
+	return p.balancer.WriteRecord(c, pipelines)
 }
 
 // UpdateSlicePipelines reacts on slices changes - closes old pipelines and open new pipelines.
--- a/internal/pkg/service/stream/storage/level/local/diskwriter/network/router/pipeline/pipeline_slice.go
+++ b/internal/pkg/service/stream/storage/level/local/diskwriter/network/router/pipeline/pipeline_slice.go
@@ -6,7 +6,6 @@ import (
 	"time"
 
 	"github.com/cenkalti/backoff/v4"
-	"go.opentelemetry.io/otel/attribute"
 
 	"github.com/keboola/keboola-as-code/internal/pkg/log"
 	"github.com/keboola/keboola-as-code/internal/pkg/service/common/ctxattr"
@@ -160,21 +159,9 @@ func (p *SlicePipeline) tryOpen() error {
 
 	ctx := p.ctx
 
-	// Get connection
-	conn, found := p.connections.ConnectionToVolume(p.slice.SliceKey.VolumeID)
-	if !found || !conn.IsConnected() {
-		return errors.Errorf("no connection to the volume %q", p.slice.SliceKey.VolumeID.String())
-	}
-
-	ctx = ctxattr.ContextWith(
-		ctx,
-		attribute.String("writerNodeId", conn.RemoteNodeID()),
-		attribute.String("writerNodeAddress", conn.RemoteAddr()),
-	)
-
 	// Open remote RPC file
 	// The disk writer node can notify us of its termination. In that case, we have to gracefully close the pipeline, see Close method.
-	remoteFile, err := rpc.OpenNetworkFile(ctx, p.logger, p.telemetry, p.connections.NodeID(), conn, p.slice.SliceKey, p.slice.LocalStorage, p.Close)
+	remoteFile, err := rpc.OpenNetworkFile(ctx, p.logger, p.telemetry, p.connections, p.slice.SliceKey, p.slice.LocalStorage, p.Close)
 	if err != nil {
 		return errors.PrefixErrorf(err, "cannot open network file for new slice pipeline")
 	}
--- a/internal/pkg/service/stream/storage/level/local/diskwriter/network/router/pipeline/pipeline_slice.go
+++ b/internal/pkg/service/stream/storage/level/local/diskwriter/network/router/pipeline/pipeline_slice.go
@@ -14,7 +14,6 @@ import (
 	pipelinePkg "github.com/keboola/keboola-as-code/internal/pkg/service/stream/sink/pipeline"
 	"github.com/keboola/keboola-as-code/internal/pkg/service/stream/storage/level/local/diskwriter/network/connection"
 	"github.com/keboola/keboola-as-code/internal/pkg/service/stream/storage/level/local/diskwriter/network/router/balancer"
-	"github.com/keboola/keboola-as-code/internal/pkg/service/stream/storage/level/local/diskwriter/network/rpc"
 	"github.com/keboola/keboola-as-code/internal/pkg/service/stream/storage/level/local/encoding"
 	encodingCfg "github.com/keboola/keboola-as-code/internal/pkg/service/stream/storage/level/local/encoding/config"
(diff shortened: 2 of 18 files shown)


internal/pkg/service/stream/storage/level/local/diskwriter/network/router/pipeline/pipeline_sink.go
```diff
@@ -93,9 +93,13 @@ func (p *SinkPipeline) ReopenOnSinkModification() bool {
...
 func (p *SinkPipeline) WriteRecord(c recordctx.Context) (pipelinePkg.WriteResult, error) {
+	// Make a local copy of pipelines while holding the read lock
 	p.writeLock.RLock()
-	defer p.writeLock.RUnlock()
-	return p.balancer.WriteRecord(c, p.pipelines)
+	pipelines := make([]balancer.SlicePipeline, len(p.pipelines))
+	copy(pipelines, p.pipelines)
+	p.writeLock.RUnlock()
+
+	return p.balancer.WriteRecord(c, pipelines)
```
Matovidlo, line 102: Deadlock. As from now we will deliver the records always into old pipeline. This have 2 consequences
1. Data loss (currently we have evidence for about 0.04% data loss)
2. Maybe we are not able to switch the pipelines on the fly.

It means that we always need to proceed the write for old pipeline even though there is new pipeline already created. It would be nice this code to be functioning as atomic swap (hot swap).
As we have agreed this is above the SLA of the service which means I will work on other things and have evidence of this issue. https://keboola.atlassian.net/browse/PSGO-1014

internal/pkg/service/stream/storage/level/local/diskwriter/network/router/pipeline/pipeline_slice.go
```diff
@@ -160,29 +167,20 @@ func (p *SlicePipeline) tryOpen() error {
 
 	ctx := p.ctx
 
-	// Get connection
-	conn, found := p.connections.ConnectionToVolume(p.slice.SliceKey.VolumeID)
-	if !found || !conn.IsConnected() {
-		return errors.Errorf("no connection to the volume %q", p.slice.SliceKey.VolumeID.String())
-	}
-
```
Matovidlo, line 168: Connection is created inside encoding pipeline

internal/pkg/service/stream/storage/level/local/diskwriter/network/router/pipeline/pipeline_slice.go
```diff
@@ -160,29 +167,20 @@ func (p *SlicePipeline) tryOpen() error {
...
-		attribute.String("writerNodeId", conn.RemoteNodeID()),
-		attribute.String("writerNodeAddress", conn.RemoteAddr()),
+		p.slice.SliceKey,
+		p.telemetry,
+		p.connections,
+		p.slice.Mapping,
+		p.slice.Encoding,
+		p.slice.LocalStorage,
+		p.Close,
+		nil, // Do not override network output
```
Matovidlo, line 181: In production we want to have actual network output. This argument is for tests

internal/pkg/service/stream/storage/level/local/encoding/encoder/factory_test.go
```diff
@@ -57,7 +67,17 @@ func TestDefaultFactory_FileTypeInvalid(t *testing.T) {
...
+	_, err := d.EncodingManager().OpenPipeline(
+		ctx,
+		slice.SliceKey,
+		d.Telemetry(),
+		d.ConnectionManager(),
+		slice.Mapping,
+		slice.Encoding,
+		slice.LocalStorage,
+		func(ctx context.Context, cause string) {},
+		discardOutput{},
```
Matovidlo, line 79: Fake network output

internal/pkg/service/stream/storage/level/local/encoding/pipeline.go
```diff
@@ -434,12 +456,32 @@ func (p *pipeline) processChunks(ctx context.Context, clk clockwork.Clock, encod
...
+					// Open remote RPC file
+					p.network, err = rpc.OpenNetworkFile(
+						ctx,
+						p.logger,
+						p.telemetry,
+						p.connections,
+						p.sliceKey,
+						p.localStorage,
+						p.closeFunc,
+					)
```
Matovidlo, line 479: Reopen the file when it is not possible to write into the file because it was closed. It can happend in 2 cases
1. Diskwriter is closed (shutdown). This means that the connection is disconnected and chunks are not processed. Once Diskwriter will come alive, the chunks can never be processed.
2. ETCD shutdown. As the ETCD shutdowns, the layer below is dismissed due to `DELETE` event on volume, which has many implications in which it results into abandoning the network output and file is closed. Once the ETCD is up it streams `PUT` for volume which creates new network. But as we have reference as `p.network` in previous code it caused that chunk would never be processed because of old network layer reference. With this adjustment the chunks are processed and data transfered correctly.

File on the diskwriter side is set into append mode which should cause 0 issues.

internal/pkg/service/stream/storage/level/local/encoding/manager.go
```diff
@@ -80,16 +84,40 @@ func (m *Manager) Pipelines() (out []Pipeline) {
...
+		connections,
+		mappingCfg,
+		encodingCfg,
+		localStorage,
+		m.events,
+		closeFunc,
+		network,
+	)
 	if err != nil {
+		m.removePipeline(sliceKey)
```
Matovidlo, line 120: This was missing and caused issue with "duplicate" pipeline creation. The issue is that this was registered only as hook afterwards but not removed when the constructor fails

Matovidlo (commented):
I did not cover everything with tests. It would be nice to do so by any of you

internal/pkg/service/stream/storage/level/local/diskwriter/writer.go
```diff
@@ -159,14 +163,13 @@ func (w *writer) Close(ctx context.Context) error {
...
+	errs := errors.NewMultiError()
 	if w.isClosed() {
 		return errors.New(`writer is already closed`)
 	}
-	close(w.closed)
-
-	errs := errors.NewMultiError()
 
 	// Wait for running writes
+	close(w.closed)
```
jachym-tousek-keboola, line 172: I still think this change should be reverted. There is no point in creating errs prematurely and you moved `close` out of the block that has the `// Close only once` comment.

internal/pkg/service/stream/storage/level/local/encoding/pipeline_test.go
```diff
@@ -41,9 +43,18 @@ func TestEncodingPipeline_Basic(t *testing.T) {
...
+		d.Telemetry(),
+		d.ConnectionManager(),
+		slice.Mapping,
+		slice.Encoding,
+		slice.LocalStorage,
+		func(ctx context.Context, cause string) {},
+		output,
+	)
 	require.NoError(t, err)
-
```
jachym-tousek-keboola, line 46: I'd keep this line tbh.

internal/pkg/service/stream/storage/level/local/encoding/pipeline_test.go
```diff
@@ -81,9 +92,18 @@ func TestEncodingPipeline_FlushError(t *testing.T) {
...
+		d.Telemetry(),
+		d.ConnectionManager(),
+		slice.Mapping,
+		slice.Encoding,
+		slice.LocalStorage,
+		func(ctx context.Context, cause string) {},
+		nil,
+	)
 	require.NoError(t, err)
-
```
jachym-tousek-keboola, line 86: ditto

jachym-tousek-keboola (approved):

internal/pkg/service/stream/storage/level/local/diskwriter/writer.go
Matovidlo: It was just cherry picked

hosekpeter (approved):
LGTM

internal/pkg/service/stream/storage/level/local/diskwriter/writer.go
Matovidlo: 8c92ed034827f0d9d0371b808ba8f0bbfd582e55 I put it closer to the first append. It had no reason to be between `close` and `wg.Wait()` which should be together

internal/pkg/service/stream/storage/level/local/encoding/pipeline_test.go
Matovidlo: 384002ee0e176a9cdf4d3931a93951a174829bd1

internal/pkg/service/stream/storage/level/local/encoding/pipeline_test.go
Matovidlo: 384002ee0e176a9cdf4d3931a93951a174829bd1

PR #2241, opened 2025-02-11
test: Temporary pipeline error

jachym-tousek-keboola:
Jira: https://keboola.atlassian.net/browse/PSGO-1004

**Changes:**
- Added a pipeline test for when NetworkFile is temporarily returning an error on write.

-----------

### DIFF
--- a/internal/pkg/service/stream/storage/level/local/encoding/pipeline_test.go
+++ b/internal/pkg/service/stream/storage/level/local/encoding/pipeline_test.go
@@ -505,6 +505,118 @@ foo4
 `)
 }
 
+func TestEncodingPipeline_TemporaryError(t *testing.T) {
+	t.Parallel()
+
+	ctx := context.Background()
+	tc := newEncodingTestCase(t)
+	tc.Slice.Encoding.Sync.Mode = writesync.ModeDisk
+	tc.Slice.Encoding.Sync.Wait = true
+
+	w, err := tc.OpenPipeline()
+	require.NoError(t, err)
+
+	// Writes are BLOCKING, each write is waiting for the next sync
+
+	// Write two rows and trigger sync
+	wg := &sync.WaitGroup{}
+	wg.Add(2)
+	go func() {
+		defer wg.Done()
+		n, err := w.WriteRecord(tc.TestRecord("foo1"))
+		require.NoError(t, err)
+		assert.Equal(t, 5, n)
+		tc.Logger.Infof(ctx, "TEST: write unblocked")
+	}()
+	go func() {
+		defer wg.Done()
+		n, err := w.WriteRecord(tc.TestRecord("foo1"))
+		require.NoError(t, err)
+		assert.Equal(t, 5, n)
+		tc.Logger.Infof(ctx, "TEST: write unblocked")
+	}()
+	tc.ExpectWritesCount(t, 2)
+	// When trigger sync is executed, it creates new notifier, old notifier is close and unblocks 2 writes above.
+	// New notifier is handovered into encoder, so the next write is blocked on new notifier until new trigger sync is executed
+	tc.TriggerSync(t)
+	wg.Wait()
+
+	// Let the next write fail
+	tc.Output.WriteError = errors.New("some error")
+
+	// Write one row and trigger sync
+	wg.Add(1)
+	go func() {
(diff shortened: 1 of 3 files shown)


internal/pkg/service/stream/storage/level/local/encoding/pipeline_test.go
```diff
@@ -505,6 +505,118 @@ foo4
...
+{"level":"debug","message":"writers flushed"}
+{"level":"debug","message":"sync to disk done"}
+{"level":"info","message":"TEST: write unblocked"}
+{"level":"info","message":"TEST: write unblocked"}
+{"level":"debug","message":"notifier obtained"}
+{"level":"debug","message":"starting sync to disk"}
+{"level":"debug","message":"flushing writers"}
+{"level":"debug","message":"chunk completed, aligned = true, size = \"5B\""}
+{"level":"debug","message":"writers flushed"}
+{"level":"warn","message":"chunks write failed: some error, waiting %s, chunks count = 1"}
```
Matovidlo, line 604: I don't see log where chunks were written after fail

Matovidlo (commented):
I think the test misses check for complete writing by logs

internal/pkg/service/stream/storage/level/local/encoding/pipeline_test.go
jachym-tousek-keboola: The `chunk completed` message didn't appear in the logs. I'm not quite sure why to be honest because I don't understand how chunk writer works. The message is logged in `swapChunks` which I'm guessing wasn't necessary to be called in this case? However the data did appear in the stream output and I do assert that.

Matovidlo: I see that we assert the data but it is weird, we receive `chunk write failed` and being in backoff and do not see it write after backoff interval is reached. See `encoding/pipeline.go:493`. The operation is again retried and chunks are processed

jachym-tousek-keboola: I think you made a typo? `encoding/pipeline.go` only has 480 lines.

Matovidlo: https://github.com/keboola/keboola-as-code/blob/main/internal/pkg/service/stream/storage/level/local/encoding/pipeline.go#L462

Matovidlo: I think the log is supposed to be there so you need to spend more time on that why it is missing. E.g deadlock occured, didn't wait enough or any other issue

hosekpeter (approved):
LGTM

internal/pkg/service/stream/storage/level/local/encoding/pipeline_test.go
jachym-tousek-keboola: It is actually missing on the old tests from Michal as well... Apparently the chunk is still active until you send the next record so it only gets completed when you do send the next one. But I added assertions for "chunk written" which is logged immediately.

Matovidlo: So if it is missing in old tests why don't we add this into them? I understand that the chunks can be processed retrospectivelly but I would expect test to cover it fully. Or add commentary somewhere, maybe into processChunks metohd

Matovidlo (changes requested):
I am not sure whether I understand it correctly. You said that there are some tests that has similar issue but we fixed only our test either
1. Add comment to the test that we always receive 1 previous chunk write completed
2. Adjust tests

internal/pkg/service/stream/storage/level/local/encoding/pipeline_test.go
jachym-tousek-keboola: > So if it is missing in old tests why don't we add this into them?

You don't understand. The assert for the last `chunk completed` is missing because that log does not happen. The last chunk is written but not completed.

Matovidlo (approved):

jachym-tousek-keboola:
> You said that there are some tests that has similar issue

I'm saying there are other tests having the same **non-issue**.

PR #2250, opened 2025-02-21
fix: File import operator

jachym-tousek-keboola:
If the file is near expiration then throttling should be ignored. In today's incident stream was incorrectly tracking some jobs as unfinished, therefore it wasn't rotating the files and they eventually expired.

Of course we need to also do something about the underlying issue with job tracking. I'm unsure how to fix that yet but imo this change should be made regardless.


Matovidlo (approved):
LGTM

hosekpeter (approved):
LGTM

PR #2251, opened 2025-02-24
fix: Use WithCancelCause and WithTimeoutCause

jachym-tousek-keboola:
Jira: https://keboola.atlassian.net/browse/PSGO-1018

**Changes:**
- Removing existing usages of `WithCancel` and `WithTimeout`.
- Disallowing these methods for the future.

-----------

### DIFF
--- a/build/ci/golangci.yml
+++ b/build/ci/golangci.yml
@@ -71,6 +71,9 @@ linters-settings:
       # No "gonanoid" package, use "internal/pkg/idgenerator" package instead
       - p: ^gonanoid\..*$
         msg: Use "internal/pkg/idgenerator" package instead.
+      # No context.WithCancel(), context.WithTimeout() or context.WithDeadline(), use the cause variants instead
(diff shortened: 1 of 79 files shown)


build/ci/golangci.yml
```diff
@@ -244,6 +247,10 @@ issues:
       text: "use of `os.Std(out|err)` forbidden"
       linters:
         - forbidigo
+    - path: _test\.go
+      text: "use of `context.WithTimeout` forbidden"
```
Matovidlo, line 251: Why it is forbidden to use WithTimeout in tests? Do we need to enforce it also in tests? Just asking

Matovidlo (approved):
LGTM

hosekpeter (approved):
LGTM

build/ci/golangci.yml
jachym-tousek-keboola: It's not, this is under `exclude-rules:`. It didn't feel worth the time to fix this issue in all tests (~100 cases). But I can do it if you want.

PR #2257, opened 2025-02-25
fix: Remote fixtures contains empty.json file when not exists

Matovidlo:
Jira: [PSGO-986](https://keboola.atlassian.net/browse/PSGO-986)

**Changes:**
- Add fixture into local file system into tmp directory when missing due to not cloning the project

-----------


internal/pkg/state/load_remote_test.go
```diff
@@ -442,9 +443,65 @@ func loadRemoteState(t *testing.T, m *manifest.Manifest, projectStateFile string
...
+        "name": "Bar",
+        "description": "Bar branch",
+        "isDefault": false
+      },
+      "configs": [
+        "without-rows"
+      ]
+    }
+  ]
+}`))
```
Matovidlo, line 499: @hosekpeter due to usage of AferoFS in memory I have to write file into fs and `SetState`. It is little bit unconvinient.
Should I Rework FS for local tests (within repository) to use the `fixtures` files?
I think it could be more readable that way?

jachym-tousek-keboola (approved):

hosekpeter (approved):
LGTM

PR #2259, opened 2025-02-27
fix: Improve logs

jachym-tousek-keboola:
As discussed on call, we want to log the no-op on file import retry if a successful job already exists.

### DIFF
--- a/internal/pkg/service/stream/sink/type/tablesink/keboola/bridge/bridge_test.go
+++ b/internal/pkg/service/stream/sink/type/tablesink/keboola/bridge/bridge_test.go
@@ -131,8 +131,8 @@ func TestBridge_FullWorkflow(t *testing.T) {
 {"level":"info","message":"created token","token.bucketID":"in.c-bucket","token.name":"[_internal] Stream Sink my-source/my-sink"}
 {"level":"info","message":"creating table","table.key":"456/in.c-bucket.my-table"}
 {"level":"info","message":"created table","table.key":"456/in.c-bucket.my-table"}
-{"level":"info","message":"creating staging file","token.ID":"1001","file.name":"my-source_my-sink_20000101010000","file.id":"2000-01-01T01:00:00.000Z"}
-{"level":"info","message":"created staging file","token.ID":"1001","file.resourceID":"1001","file.name":"my-source_my-sink_20000101010000","file.id":"2000-01-01T01:00:00.000Z"}
+{"level":"debug","message":"creating staging file","token.ID":"1001","file.name":"my-source_my-sink_20000101010000","file.id":"2000-01-01T01:00:00.000Z"}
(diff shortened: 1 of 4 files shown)


hosekpeter (approved):

Matovidlo (changes requested):
I would like to go through all Info logs (in DD) and remove some of the redudant ones. It is about 325K info logs per single project. Maybe more streams more logs. I think that's too much noise. I see 1M logs a day which is totally unnecessary at this stage. It helps sometimes debugging but I would like to see less noisy logs that we just skip.
Here is sample https://app.datadoghq.eu/logs?query=service%3Astream%20%40project.id%3A10837&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&event=AwAAAZVEhw4siKfqfgAAABhBWlZFaHh4RkFBQUUzZURkY1NCS2hBQUcAAAAkMDE5NTQ0OGQtM2E3Ny00MjI0LWFhNjUtZGY5YTgxOTIyZDgzAAvjow&fromUser=true&messageDisplay=inline&refresh_mode=paused&storage=hot&stream_sort=desc&viz=stream&from_ts=1740524594545&to_ts=1740611456577&live=false

Matovidlo:
@kacurez what is the cost for our logs currently? I think we need to support this logging reduction by cost of our bill. It would be nice to get some improvements in this way :) 
@jachym-tousek-keboola couple of adepts
```
b.logger.Info(ctx, `creating staging file`)
b.logger.Info(ctx, `created staging file`)
```
These logs are covered fully in unsuccess case by error log which is enough. The info logs are never used in case of error case being handled. Whole stage is lock free where the all workflow or none is executed.
I think similar applies for logs e.g `rotating file` `rotated file` and `rotating slice` `rotated slice`. In case error is there you never check the info log and it does not lead you any further.

If you never used that log then log is redundant in my opinion and there is no real reason to leave it there. We can put it at least into debug mode if needed also for tests to be passed.

jachym-tousek-keboola:
> I think similar applies for logs e.g rotating file rotated file and rotating slice rotated slice. In case error is there you never check the info log and it does not lead you any further.

Incorrect. I specifically used these logs while looking into the incident in order to find out what happened to which slice/file at what time.

Matovidlo:
Still I would like to reduce the logs. it costs us something and it is useless to have it in info level. Where is problem to put it into debug level and enable it through change of configuration and reload of the pod ?

jachym-tousek-keboola:
> Still I would like to reduce the logs. it costs us something and it is useless to have it in info level. Where is problem to put it into debug level and enable it through change of configuration and reload of the pod ?

That would be a big issue for debugging an incident like we had last week since we'd be missing the logs for the time period of the incident which is when they're crucial to have. The logs like "rotating file" are necessary to track the "lifecycle" of a file - what state it was for in any given time, when retries happens etc.

Matovidlo (approved):

PR #2260, opened 2025-02-27
feat: Cleanup storage jobs that return 404

jachym-tousek-keboola:
Jira: https://keboola.atlassian.net/browse/PSGO-1020

**Changes:**
- Cleanup storage jobs that return 404

-----------

### DIFF
--- a/internal/pkg/service/stream/sink/type/tablesink/keboola/bridge/job.go
+++ b/internal/pkg/service/stream/sink/type/tablesink/keboola/bridge/job.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"math"
+	"net/http"
 	"strconv"
 	"time"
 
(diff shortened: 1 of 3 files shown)


jachym-tousek-keboola:
@Matovidlo The task title is "Delete job when 404/cancel received" - what did you mean by "cancel"? There is no such status for storage jobs according to https://github.com/keboola/go-client/blob/82a91f7fbb0ea18160f04f4a27545e651f0cfccf/pkg/keboola/storage_job.go#L21-L26.

internal/pkg/service/stream/sink/type/tablesink/keboola/bridge/job.go
```diff
@@ -94,17 +95,23 @@ func (b *Bridge) CleanJob(ctx context.Context, job model.Job) (err error, delete
...
-	}
-
-	attr := attribute.String("job.state", jobStatus.Status)
-	ctx = ctxattr.ContextWith(ctx, attr)
-	// Check status of storage job
-	if jobStatus.Status == keboola.StorageJobStatusProcessing || jobStatus.Status == keboola.StorageJobStatusWaiting {
-		b.logger.Debugf(ctx, "cannot remove storage job, job status: %s", jobStatus.Status)
-		return nil, false
+	jobStatus, err = api.GetStorageJobRequest(keboola.StorageJobKey{ID: keboola.StorageJobID(id)}).Send(ctx)
+	if err != nil {
```
Matovidlo, line 99: By cancel I meant context cancel. This function does not need to finish because context can be canceled.
If we remove the errGroup issue probably we don't need to change this code any further.
In case there is `errors.Is(err, context.Canceled)/ errors.Is(err , context.DeadlineExceeded)` we can do 2 things
1. Clear job to rather unblock the sink
2. Do not clear job and spend time debugging error group/ why connection times out
It would be nice to simulate for example longer downtime of some component to see this more in depth.

I would merge this simple changes and do another PR or discussion about handling timeouting connection imports or canceling contexts by error groups.

Matovidlo (approved):
LGTM, copied comment into parent issue. I expect also to solve context related issues in another PR

PR #2262, opened 2025-02-28
build: Use Go 1.24.0 & go tool

jachym-tousek-keboola:
Jira: https://keboola.atlassian.net/browse/PSGO-989

Did this last week while I was blocked on other tasks.


internal/pkg/utils/testhelper/testtemplateinputs/testtemplateinputs_test.go
```diff
@@ -11,9 +11,9 @@ import (
 func TestNewUserErrorWithCode(t *testing.T) {
 	t.Parallel()
 
-	require.NoError(t, os.Setenv("CUSTOM_ENV", "val1"))      //nolint:forbidigo, tenv
-	require.NoError(t, os.Setenv("KBC_SECRET_VAR2", "val2")) //nolint:forbidigo, tenv
-	require.NoError(t, os.Setenv("KBC_SECRET_VAR3", "val3")) //nolint:forbidigo, tenv
+	require.NoError(t, os.Setenv("CUSTOM_ENV", "val1"))      //nolint:forbidigo, tenv, usetesting
+	require.NoError(t, os.Setenv("KBC_SECRET_VAR2", "val2")) //nolint:forbidigo, tenv, usetesting
+	require.NoError(t, os.Setenv("KBC_SECRET_VAR3", "val3")) //nolint:forbidigo, tenv, usetesting
```
jachym-tousek-keboola, line 16: Linter report:

```
internal/pkg/utils/testhelper/testtemplateinputs/testtemplateinputs_test.go:14:21: os.Setenv() could be replaced by t.Setenv() in TestNewUserErrorWithCode (usetesting)
```

But when I tried that I got this:

```
anic: testing: test using t.Setenv or t.Chdir can not use t.Parallel [recovered]
	panic: testing: test using t.Setenv or t.Chdir can not use t.Parallel
```

Alternatively I can use these new functions and drop `t.Parallel()` here.

Matovidlo: This is correct behaviour when you execute it in parallel. 
We want to keep `t.Parallel()`  and use `os.Setenv()` and `nolint`.

.github/workflows/check-licenses.yml
```diff
@@ -43,13 +43,13 @@ jobs:
         # Ignored:
         #    - github.com/eliukblau/pixterm, license detection failed (Reciprocal - MPL-2.0)
         run: |
-          go tool go-licenses check ./... --disallowed_types forbidden,restricted,unknown \
+          go-licenses check ./... --disallowed_types forbidden,restricted,unknown \
```
jachym-tousek-keboola: go-licenses unfortunately don't work through `go tool` so I had to revert my changes that completely removed `tools.sh` and keep it as it was. I think this is good for now - we might find out that some tools that are only used in release workflows might have similar problem - by keeping tools.sh for now we have an easy way to revert if we need to.

Matovidlo (approved):
LGTM I will not block it but this was unnecessary to do ATM

hosekpeter (approved):

PR #2263, opened 2025-03-03
fix: Ignore sink throttle when file has too many slices

jachym-tousek-keboola:
Jira: https://keboola.atlassian.net/browse/PSGO-1025

**Changes:**
- Similar to the previous case where we ignore sink throttle if file is near expiration, we also ignore it if the file has too many slices.

-----------


internal/pkg/service/stream/storage/node/coordinator/filerotation/operator.go
```diff
@@ -361,12 +361,18 @@ func (o *operator) rotateFile(ctx context.Context, file *fileData) {
...
-			// The file credentials are near expiration but the sink is throttled
+		switch result.result {
+		case expirationThreshold:
+			// The sink is throttled but the file credentials are near expiration
 			// This case indicates that job status tracking is not working properly and throttling is false-positive
-			// In this case we log an error to know about this case but let it continue
+			// In this case we log an error to know about this case but let file rotation continue
 			o.logger.Error(ctx, "file rotation: sink is throttled but file is near expiration")
-		} else {
+		case sliceCountThreshold:
```
Matovidlo, line 370: If I am not mistaken this option can be adjusted via API.
Therefore it is not reliable source of information.
https://github.com/keboola/keboola-as-code/blob/e6d97d774abbedfd6b6ef268ab3b8a457c37e644/internal/pkg/service/stream/storage/level/target/config/config.go#L80

Matovidlo (changes requested):
It won't work properly when chaning default configuration

internal/pkg/service/appsproxy/proxy/apphandler/authproxy/oauthproxy/pagewriter.go
```diff
@@ -95,7 +95,7 @@ func (pw *pageWriter) WriteErrorPage(w http.ResponseWriter, req *http.Request, o
 	// Log warning
 	pw.logger.Warn(req.Context(), strings.Join(messages, "\n")) //nolint:contextcheck // false positive
 
-	pw.pageWriter.WriteErrorPage(w, req, &pw.app, opts.Status, joinedMessage+opts.AppError, exceptionID)
+	pw.pageWriter.WriteErrorPage(w, req, &pw.app, opts.Status, joinedMessage+"\n"+opts.AppError, exceptionID)
```
jachym-tousek-keboola, line 98: Small unrelated fix for appsproxy.

internal/pkg/service/stream/sink/type/tablesink/keboola/bridge/model/repository/job/schema/job_schema.go
jachym-tousek-keboola, line 1: In file repo the schema is in `file_schema.go` + `file_schema_test.go`. I updated the other schemas accordingly.

internal/pkg/service/stream/storage/node/coordinator/filerotation/conditions.go
```diff
@@ -52,12 +53,20 @@ func (c fileRotationConditionResult) String() string {
...
 		return "unknown"
 	}
 }
 
 func shouldImport(cfg targetConfig.ImportConfig, now, openedAt, expiration time.Time, stats statistics.Value) fileRotationConditionResult {
+	if threshold := cfg.MaxSlices; stats.SlicesCount >= threshold {
+		// Max slices take precedence over other settings.
+		// It is necessary in order to prevent issues with etcd txn operation limit.
+		return newConditionResult(maxSlices, fmt.Sprintf("max slices exceeded, slices count: %d, threshold: %d", stats.SlicesCount, threshold))
+	}
```
jachym-tousek-keboola, line 68: New priority condition - if the file has too many slices, we rotate it.

internal/pkg/service/stream/storage/node/coordinator/filerotation/operator.go
```diff
@@ -367,7 +367,7 @@ func (o *operator) rotateFile(ctx context.Context, file *fileData) {
 			// This case indicates that job status tracking is not working properly and throttling is false-positive
 			// In this case we log an error to know about this case but let file rotation continue
 			o.logger.Error(ctx, "file rotation: sink is throttled but file is near expiration")
-		case sliceCountThreshold:
+		case maxSlices:
```
jachym-tousek-keboola, line 370: maxSlices condition result ignores throttle

hosekpeter (approved):

internal/pkg/service/stream/config/config_test.go
```diff
@@ -291,6 +291,8 @@ storage:
                 # Timeout of the file import operation. Validation rules: required,minDuration=30s,maxDuration=60m
                 fileImportTimeout: 15m0s
             import:
+                # Max number of slices in a file before an import is triggered, takes precedence over other settings. Validation rules: required,min=100,max=1000
+                maxSlices: 1000
```
Matovidlo, line 295: It won't work I think. Again we miss the metric which I was expected to see at first time regarding actual TXN amount.  1000 slices for next rotation will result into same issue I think
Based on your script that you prepared, deleting file requires
1. Deleting up to 3 types of statistics
2. Deleting up to 4 types of slices (probably there can be only 2 types at the same time)
3. Deleting one keboola file
4. Deleting 4 types of file ( probably there can be only 1 type of file at the same time)
Overall if I would be pessimistic there are 3+4+1+4 types for single file = 12
If It creates only slices it adds probably 1000 slices but, so it would be =1012, but that's theoretical.

As we have issue in production there is possiblity to take file that is not imported yet and check that keys which would satisfy this comment

jachym-tousek-keboola: I don't think I have kubectl access to the production stack to check this.

Matovidlo: We'll check that on monday

Matovidlo (approved):
LGTM

PR #2268, opened 2025-03-04
feat: stream encryption support for AWS

kauf303:
Jira: [ST-2849](https://keboola.atlassian.net/browse/ST-2849)

**Changes:**
- added service accounts to stream pods
- added environment variables from kubernetes secret
- added SA and empty secret to minikube deployment
-----------

deployment [pipeline](https://dev.azure.com/keboola-prod/Keboola%20Connection%20AWS/_releaseProgress?_a=release-environment-logs&releaseId=48005&environmentId=132697) test.

related Infrastructure [PR](https://github.com/keboola/infrastructure/pull/2290)

[ST-2849]: https://keboola.atlassian.net/browse/ST-2849?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/provisioning/stream/kubernetes/build.sh
+++ b/provisioning/stream/kubernetes/build.sh
@@ -76,8 +76,6 @@ elif [[ "$CLOUD_PROVIDER" == "azure" ]]; then
   envsubst < templates/cloud/azure/service-api.yaml > deploy/cloud/azure/service-api.yaml
    envsubst < templates/cloud/azure/service-http-source.yaml > deploy/cloud/azure/service-http-source.yaml
    envsubst < templates/cloud/azure/sc-stream-api.yaml > deploy/cloud/azure/sc-stream-api.yaml
-  envsubst < templates/cloud/azure/secret-infra.yaml > deploy/cloud/azure/secret-infra.yaml
-  envsubst < templates/cloud/azure/serviceaccount.yaml > deploy/cloud/azure/serviceaccount.yaml
 elif [[ "$CLOUD_PROVIDER" == "local" ]]; then
   mkdir deploy/cloud/local
   envsubst < templates/cloud/local/service-api.yaml > deploy/cloud/local/service-api.yaml
--- a/provisioning/stream/kubernetes/build.sh
+++ b/provisioning/stream/kubernetes/build.sh
@@ -76,8 +76,12 @@ elif [[ "$CLOUD_PROVIDER" == "azure" ]]; then
   envsubst < templates/cloud/azure/service-api.yaml > deploy/cloud/azure/service-api.yaml
    envsubst < templates/cloud/azure/service-http-source.yaml > deploy/cloud/azure/service-http-source.yaml
    envsubst < templates/cloud/azure/sc-stream-api.yaml > deploy/cloud/azure/sc-stream-api.yaml
+  envsubst < templates/cloud/azure/secret-infra.yaml > deploy/cloud/azure/secret-infra.yaml
+  envsubst < templates/cloud/azure/serviceaccount.yaml > deploy/cloud/azure/serviceaccount.yaml
 elif [[ "$CLOUD_PROVIDER" == "local" ]]; then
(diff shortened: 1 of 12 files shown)


Matovidlo (approved):

Matovidlo (changes requested):
Ok sorry my bad I approved it but find issue in CI later on. I see 2 issues
1. We miss in `aws.sh` change the encryption to `aws`. We can fix that probably in next PR
2. The problem is local `minikube` deployment which is done in CI also. https://github.com/keboola/keboola-as-code/actions/runs/13654471799/job/38170215287?pr=2268 . You updated the general templates for streams instead of `aws` related ones. Now we miss `stream-infra-secrets` in whole deployment. It would be ideal to add this `envFrom` into aws instead of top level templates that are for each deployment the same

Just small thing, we updated one unstable test so rebase on main so the CI is green :)

kauf303:
> Ok sorry my bad I approved it but find issue in CI later on. I see 2 issues
> 
>     1. We miss in `aws.sh` change the encryption to `aws`. We can fix that probably in next PR
> 
>     2. The problem is local `minikube` deployment which is done in CI also. https://github.com/keboola/keboola-as-code/actions/runs/13654471799/job/38170215287?pr=2268 . You updated the general templates for streams instead of `aws` related ones. Now we miss `stream-infra-secrets` in whole deployment. It would be ideal to add this `envFrom` into aws instead of top level templates that are for each deployment the same
> 
> 
> Just small thing, we updated one unstable test so rebase on main so the CI is green :)

thanks for review and pointers. I upgraded description according to new changes which seem to be passing tests.

Matovidlo (approved):

provisioning/stream/kubernetes/build.sh
```diff
@@ -80,4 +80,6 @@ elif [[ "$CLOUD_PROVIDER" == "local" ]]; then
   mkdir deploy/cloud/local
   envsubst < templates/cloud/local/service-api.yaml > deploy/cloud/local/service-api.yaml
   envsubst < templates/cloud/local/service-http-source.yaml > deploy/cloud/local/service-http-source.yaml
+  envsubst < templates/cloud/local/secret-infra.yaml > deploy/cloud/local/secret-infra.yaml
+  envsubst < templates/cloud/local/serviceaccount.yaml > deploy/cloud/local/serviceaccount.yaml
```
jachym-tousek-keboola, line 84: Why are you changing things for `"$CLOUD_PROVIDER" == "local"` branch? :eyes:

jachym-tousek-keboola (approved):

kauf303:
I just realized that this would _probably_ break Azure deployments, because it's Infrastructure part is not ready yet but those changes are counting with that state. I'll re-think this change and refactor. stay tuned ;)

provisioning/stream/kubernetes/build.sh
kauf303: "envsubst is weak for kubernetes templates". so I wanted to provide the same resources for all actual deployments/statefulsets by adding "fake empty" secret and serviceaccount.
that fixed local deployment. however there's no implementation of "Infra" parts for Azure so that won't work there now. I'm figuring out migration now and provide more changes.

Matovidlo:
@kauf303 Just one thing maybe to point out. Isn't it better to finish azure too and do it as once PR ? Just want to save you time refactoring in order to deliver it faster and easier for you :slightly_smiling_face:

kauf303:
> @kauf303 Just one thing maybe to point out. Isn't it better to finish azure too and do it as once PR ? Just want to save you time refactoring in order to deliver it faster and easier for you 🙂

you were reading my mind. I worked on Azure Infra part today. it seem to be ready for testing now.

kauf303:
I moved this work into new https://github.com/keboola/keboola-as-code/pull/2282 for better visibility.

PR #2274, opened 2025-03-11
test: Run keboola bridge e2e test with t.Parallel()

jachym-tousek-keboola:
Jira: https://keboola.atlassian.net/browse/PSGO-972

### DIFF
--- a/test/stream/bridge/keboola/keboola_test.go
+++ b/test/stream/bridge/keboola/keboola_test.go
@@ -31,7 +31,9 @@ import (
 )
 
 // To see details run: TEST_VERBOSE=true go test ./test/stream/bridge/... -v.
-func TestKeboolaBridgeWorkflow(t *testing.T) { // nolint: paralleltest
+func TestKeboolaBridgeWorkflow(t *testing.T) {
+	t.Parallel()
+
 	ctx, cancel := context.WithTimeout(t.Context(), 300*time.Second)
 	defer cancel()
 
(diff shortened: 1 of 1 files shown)


test/stream/bridge/keboola/keboola_test.go
```diff
@@ -328,6 +333,9 @@ func TestNetworkIssuesKeboolaBridgeWorkflow(t *testing.T) { // nolint: parallelt
 
 	// Update configuration to make the cluster testable
 	configFn := func(cfg *config.Config) {
+		cfg.API.Listen = "0.0.0.0:8099"
+		cfg.API.PublicURL = &url.URL{Scheme: "http", Host: "localhost:8099"}
```
jachym-tousek-keboola, line 337: @Matovidlo Ultimately the issue of why the tests were not able to run in parallel was simple: they were both trying to use the default port 8000 so of course only one of them succeeded. But I'm unsure how to fix it. Hardcoding a different port into one of the tests doesn't feel like a good solution but I don't think we can use your PortManager class here either since these are separate tests rather than one test with several test cases like in appsproxy.

jachym-tousek-keboola: Nevermind... it looks like we don't actually need to set PublicURL in these tests. This works'
```
cfg.API.Listen = "0.0.0.0:0"
```

test/stream/bridge/keboola/keboola_test.go
```diff
@@ -314,7 +319,9 @@ func TestKeboolaBridgeWorkflow(t *testing.T) { // nolint: paralleltest
 	ts.checkKeboolaTable(t, ctx, 1, 129)
 }
 
-func TestNetworkIssuesKeboolaBridgeWorkflow(t *testing.T) { // nolint: paralleltest
+func TestNetworkIssuesKeboolaBridgeWorkflow(t *testing.T) {
```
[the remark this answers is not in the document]
jachym-tousek-keboola: I tried, it didn't work.

Matovidlo (approved):
LGTM. What about some ` // nolint: paralleltest` issues in our tests using ETCD that you mentioned in slack? Can we fix them also by setting API confguration listen endpoint to port `0`?

hosekpeter (approved):
LGTM

PR #2276, opened 2025-03-11
fix: Possible panic when creating a source or sink

jachym-tousek-keboola:
Jira: https://keboola.atlassian.net/browse/PSGO-1028

**Changes:**
- Return an error if source id or sink id is empty.

-----------

I could easily add an e2e test for this but in my opinion such test is not really worth having.

### DIFF
--- a/internal/pkg/service/stream/api/mapper/sink_request.go
+++ b/internal/pkg/service/stream/api/mapper/sink_request.go
@@ -22,6 +22,9 @@ func (m *Mapper) NewSinkEntity(parent key.SourceKey, payload *api.CreateSinkPayl
 	} else {
 		entity.SinkID = key.SinkID(strhelper.NormalizeName(string(*payload.SinkID)))
 	}
+	if entity.SinkID == "" {
+		return definition.Sink{}, svcerrors.NewBadRequestError(errors.Errorf(`"sinkId" must not be empty`))
+	}
 
 	// Name
 	entity.Name = payload.Name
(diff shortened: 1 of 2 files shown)


hosekpeter (approved):
LGTM

Matovidlo (commented):
That's weird. We say in the API that source,sink ID can be nil, but we return that these values must not be empty
https://github.com/keboola/keboola-as-code/blob/main/api/stream/design.go#L1050
I miss some more specific reason why it cannot be created. either Providerd SinkID is incorrect which would be in else branch or name is incorrect

jachym-tousek-keboola:
> That's weird. We say in the API that source,sink ID can be nil, but we return that these values must not be empty
[main/api/stream/design.go#L1050](https://github.com/keboola/keboola-as-code/blob/main/api/stream/design.go#L1050)

It will always exist in output. On input side it is optional because when undefined we generate it from the name.

> I miss some more specific reason why it cannot be created. either Providerd SinkID is incorrect which would be in else branch or name is incorrect

The issue is specifically if SinkID is not provided and Name is a string that ends up as an empty string after being passed through `strhelper.NormalizeName()`.

jachym-tousek-keboola:
@Matovidlo I really don't think it's possible to implement proper validation with Goa. From my research, all the validation rules are set into this [struct](https://github.com/goadesign/goa/blob/0e921149e574ab58d8118f82bb59e2eef13fb485/expr/attribute.go#L62-L100). Which as you can see has very limited capabilities.

Matovidlo (commented):
Have you used this? https://chatgpt.com/g/g-mLuQDGyro-goa-design-wizard
Also I saw it can review your current DSL model. I want to make sure that I will put Approve only when I am convinced it cannot be done at all. 

It seems to me that we should get rid of goa if it has such limitation because there are plenty HTTP server implementation that can do such validation

jachym-tousek-keboola:
I can't use it:
image

PR #2278, opened 2025-03-12
build: Update dependencies

jachym-tousek-keboola:


go.mod
Matovidlo, line 1: I miss update of these security alerts https://github.com/keboola/keboola-as-code/security/dependabot

hosekpeter (approved):

go.mod
jachym-tousek-keboola: Oh I missed that somehow. How about we merge this as is to fix the test and I'll look into the security alerts separately?

jachym-tousek-keboola: https://github.com/keboola/keboola-as-code/pull/2279

Matovidlo (approved):

PR #2279, opened 2025-03-12
build: Security updates

jachym-tousek-keboola:
Fixing security issues: https://github.com/keboola/keboola-as-code/security/dependabot

This one isn't fixed as there is no new version that would address the problem: https://github.com/keboola/keboola-as-code/security/dependabot/19


jachym-tousek-keboola:
hmm... what should we do about these? I don't know enough about the code where these are reported to decide how to address them. @hosekpeter @Matovidlo 
https://github.com/keboola/keboola-as-code/security/code-scanning

Matovidlo:
I think it does not need to be addressed it is used by cli operation which saves this secrets on local machine

Matovidlo (approved):

PR #2281, opened 2025-03-14
feat: Json number decode. Used in serde

Matovidlo:
Jira: https://keboola.atlassian.net/browse/PSGO-973

**Changes:**
- ...

-----------


internal/pkg/service/templates/api/service/mapper.go
```diff
@@ -86,9 +88,19 @@ func (m Mapper) TaskPayload(model task.Task) (r *Task, err error) {
 			}
 		}
 
-		if v, ok := model.Outputs["configId"].(string); ok {
+		fmt.Println("model.Outputs", model.Outputs)
+		fmt.Println("model.Outputs[configId]", model.Outputs["configId"])
+		fmt.Println("reflect.TypeOf(model.Outputs[configId])", reflect.TypeOf(model.Outputs["configId"]))
+
+		// Handle configId from preserved JSON number
+		if configIDNum, ok := model.Outputs["configId"].(json.Number); ok {
```
Matovidlo, line 96: Here is type casting

pkg/lib/operation/project/local/template/use/operation.go
```diff
@@ -118,7 +119,7 @@ type TemplatePlan struct {
 
 type Result struct {
 	InstanceID string
-	ConfigID   string
+	ConfigID   uint64
```
Matovidlo, line 122: Here is field that is stored within Task structure in ETCD using Serde

internal/pkg/encoding/json/json.go
```diff
@@ -68,6 +68,16 @@ func MustDecodeString(data string, m any) {
 	}
 }
 
+// DecodePreserveNumber decodes JSON data while preserving number types instead of converting all numbers to float64
+func DecodePreserveNumber(data []byte, m any) error {
+	decoder := json.NewDecoder(bytes.NewReader(data))
+	decoder.UseNumber()
```
jachym-tousek-keboola, line 74: I didn't know json had this option, that's indeed useful. However if we go this way imo the decoder should be created just once inside `NewJSON` in `serde.go`, not again every time the decoding function is called.

jachym-tousek-keboola (changes requested):

internal/pkg/encoding/json/json.go
Matovidlo: I just show you the way, I would close this PR and you will create another one where you will do it properly if you see it in different way :slightly_smiling_face:

jachym-tousek-keboola: Alright.

jachym-tousek-keboola: Actually it turns out I was wrong. The decoder takes the data as a reader in the constructor and as such can't be reused. So it can't be improved the way I wanted.

PR #2284, opened 2025-03-18
fix: Integer decode in etcd serde

jachym-tousek-keboola:
Jira: https://keboola.atlassian.net/browse/PSGO-973

**Changes:**
- Fixed json decode in etcd serde to not change integers to floats

-----------


internal/pkg/service/common/etcdclient/config.go
```diff
@@ -22,6 +22,7 @@ type Config struct {
 	KeepAliveTimeout  time.Duration `configKey:"keepAliveTimeout" configUsage:"Etcd keep alive timeout." validate:"required"`
 	KeepAliveInterval time.Duration `configKey:"keepAliveInterval" configUsage:"Etcd keep alive interval." validate:"required"`
 	DebugLog          bool          `configKey:"debugLog" configUsage:"Etcd operations logging as debug messages."`
+	JSONNumbers       bool          `configKey:"-" configUsage:"Use JSON numbers in serde."`
```
jachym-tousek-keboola, line 25: Not quite sure how best to control that the new serde should be used only for templates API. I added new boolean here into the config, intentionally removed the key so that it can't be set by env variable and changed `internal/pkg/service/templates/dependencies/api.go` to set it to true and similarly adjusted mocked dependencies for templates test. Do you have a better idea?

Matovidlo: This would be the best to be done with compilation flag :thinking:  . When I set this to true into production I will run into issues. 
https://pkg.go.dev/cmd/compile#hdr-Function_Directives
Also it is nice practice to know how to set compilation flags, and propagate them into final docker image that is deployed. 
But this is also sufficient to have it under flag

Matovidlo (commented):
@natocTo we are changing the configuration ID from `outputs` response for templates preview endpoint to be integer instead of string. Is there any adjustement on UI side or is it type idempotent? :slightly_smiling_face:  thanks

internal/pkg/service/common/etcdclient/config.go
Matovidlo: And what about compilation flags? I think when you will change the configuration the stream API just fails on `panic`. With compilation flag it is set into stone and cannot be changed :thinking:  . Ofc we do not need to spend time on that now, we can create task

Matovidlo (approved):

PR #2291, opened 2025-03-21
feat: Add adjustable ETCD TXN for provisioning

Matovidlo:
**Changes:**
- Make ETCD TXN OPS adjustable using `kbc-stacks` PRs

-----------

### DIFF
--- a/provisioning/stream/common.sh
+++ b/provisioning/stream/common.sh
@@ -76,6 +76,8 @@ kubectl apply -f ./kubernetes/deploy/namespace.yaml
 
 # Get etcd root password, if it is already present
 export ETCD_ROOT_PASSWORD=$(kubectl get secret --namespace "$NAMESPACE" stream-etcd -o jsonpath="{.data.etcd-root-password}" 2>/dev/null | base64 -d)
+# Override default txn ops
+export STREAM_ETCD_MAX_TXN_OPS="${STREAM_ETCD_MAX_TXN_OPS:=1024}"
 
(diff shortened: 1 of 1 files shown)


provisioning/stream/common.sh
```diff
@@ -92,7 +94,9 @@ helm upgrade \
   --set "resources.limits.memory=$STREAM_ETCD_MEMORY_HARD_LIMIT" \
   --set "resources.requests.cpu=$STREAM_ETCD_CPU_SOFT_LIMIT" \
   --set "extraEnvVars[3].name=GOMEMLIMIT" \
-  --set "extraEnvVars[3].value=${STREAM_ETCD_MEMORY_SOFT_LIMIT}B"
+  --set "extraEnvVars[3].value=${STREAM_ETCD_MEMORY_SOFT_LIMIT}B" \
+  --set "extraEnvVars[4].name=ETCD_MAX_TXN_OPS" \
+  --set-string "extraEnvVars[4].value=${STREAM_ETCD_MAX_TXN_OPS}"
```
jachym-tousek-keboola, line 99: what's the difference between `--set` and `--set-string`? :thinking:

Matovidlo: When `STREAM_ETCD_MAX_TXN_OPS` is `integer` is represents using `--set` as `integer` which fails deployment. When you use `--set-string` you always represent `integer` as `string` which `extraEnvVars` expects

jachym-tousek-keboola (approved):

PR #2292, opened 2025-03-21
feat: Log txn ops when switching to imported state

jachym-tousek-keboola:


Matovidlo (approved):
LGTM, can we add it also to `filerotate` module? It also fails

PR #2296, opened 2025-03-24
refactor: Reduce congnitive complexity of Diskwriter network

Matovidlo:
**Changes:**
- Connection is created/constructed using method
- Client closes connection in separate method

-----------


internal/pkg/service/stream/storage/level/local/diskwriter/network/connection/connection.go
```diff
@@ -153,32 +153,7 @@ func (m *Manager) updateConnections(ctx context.Context) {
 	}
 
 	// Detect new nodes - to open connection
```
jachym-tousek-keboola, line 155: This comment should be moved inside of `getConnections` with possibly a new comment here.

jachym-tousek-keboola (approved):

internal/pkg/service/stream/storage/level/local/diskwriter/network/connection/connection.go
Matovidlo: fixed

hosekpeter (approved):

PR #2297, opened 2025-03-24
feat: Remove decrypted tokens

jachym-tousek-keboola:
Second phase of token migration for streams - removal of the decrypted values.


Matovidlo (commented):
I am unsure what we want to achieve by this PR.
We still have Token and Encrypted token present in the schema. We just encrypt remaining tokens for which we have migration for.
Basically we want to migrate all streams and remove the Token forever. 

Are we just unsure that someone is not migrated?

Matovidlo (approved):
LGTM

PR #2310, opened 2025-03-31
fix: Prevent merge operation issues logger as error

Matovidlo:
Jira: [PSGO-855](https://keboola.atlassian.net/browse/PSGO-855)

**Changes:**
- It is expected that during downtime of ETCD, the `o.files` are all reseted, therefore also local locks are created again and it starts 2 go routines for closing/importing/rotating instead of having 1 go routine chagning state.
- Let other go routine switch state but report it as warning. Should not happen in other case

-----------

### DIFF
--- a/internal/pkg/service/stream/storage/node/coordinator/fileimport/operator.go
+++ b/internal/pkg/service/stream/storage/node/coordinator/fileimport/operator.go
@@ -26,6 +26,7 @@ import (
 	targetConfig "github.com/keboola/keboola-as-code/internal/pkg/service/stream/storage/level/target/config"
 	"github.com/keboola/keboola-as-code/internal/pkg/service/stream/storage/model"
 	storageRepo "github.com/keboola/keboola-as-code/internal/pkg/service/stream/storage/model/repository"
+	fileRepo "github.com/keboola/keboola-as-code/internal/pkg/service/stream/storage/model/repository/file"
 	"github.com/keboola/keboola-as-code/internal/pkg/service/stream/storage/node"
 	"github.com/keboola/keboola-as-code/internal/pkg/service/stream/storage/node/coordinator/clusterlock"
 	"github.com/keboola/keboola-as-code/internal/pkg/service/stream/storage/statistics"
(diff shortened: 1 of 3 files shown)


hosekpeter (approved):

internal/pkg/service/stream/storage/node/coordinator/fileimport/operator.go
```diff
@@ -380,8 +381,16 @@ func (o *operator) doImportFile(ctx context.Context, lock *etcdop.Mutex, file *f
...
+	if errors.Is(err, fileRepo.ErrInvalidStateTransition) {
+		o.logger.Warnf(dbCtx, "skipping file transition to imported state: %s", err)
+		// Mark as processed to avoid retry for known state transition errors
+		file.Processed = true
+		return stats.Staging, nil
+	}
+
 	if err != nil {
-		o.logger.Infof(ctx, `Switch file "%s" to imported state used %d operations`, file.FileKey.String(), result.MaxOps())
+		o.logger.Warnf(ctx, `Not switching file "%s" to imported state used %d operations`, file.FileKey.String(), result.MaxOps())
```
Matovidlo, line 393: This is unrelated change of PR, I have reversed the logging, this is unexpected and result of switching the file is that it is `not switched` as error was returned. Also I set it to higher log level as it we should look sometimes also at warnings we receive from our service.

internal/pkg/service/stream/storage/node/coordinator/fileimport/operator.go
```diff
@@ -380,8 +381,16 @@ func (o *operator) doImportFile(ctx context.Context, lock *etcdop.Mutex, file *f
 	// Switch file to the imported state
 	result := o.storage.File().SwitchToImported(file.FileKey, o.clock.Now()).RequireLock(lock).Do(dbCtx)
 	err = result.Err()
+	// Check specifically for invalid state transition errors
+	if errors.Is(err, fileRepo.ErrInvalidStateTransition) {
+		o.logger.Warnf(dbCtx, "skipping file transition to imported state: %s", err)
+		// Mark as processed to avoid retry for known state transition errors
+		file.Processed = true
+		return stats.Staging, nil
+	}
```
jachym-tousek-keboola, line 390: This doesn't look correct to me. We want to skip the error only if the file already is in the target state. Not if it failed due to being in let's say "Writing" state.

jachym-tousek-keboola (changes requested):

internal/pkg/service/stream/storage/node/coordinator/fileimport/operator.go
jachym-tousek-keboola: Also please add a comment explaining how the edge case that the file is already imported happens.

Matovidlo: > This doesn't look correct to me. We want to skip the error only if the file already is in the target state. Not if it failed due to being in let's say "Writing" state.

It cannot because the `doImport` is executed only when file state is `Importing`. The problem lays only in absence of local lock during ETCD restart which new entity does not have. However the state is always correct. `Importing` -> `Imported` or in other case `Writing` -> `Closing` or `Closing->Importing`

> Also please add a comment explaining how the edge case that the file is already imported happens.

9634e0352a4b8eb2e95bd2503b6afda39e3636a4

jachym-tousek-keboola (approved):

PR #2311, opened 2025-03-31
fix: Remove old session when ETCD was not up

Matovidlo:
Jira: [PSGO-830](https://keboola.atlassian.net/browse/PSGO-830)

**Changes:**
- New session should be ok without lock is locked by another session that was already expired

-----------


internal/pkg/service/common/etcdop/mutex.go
```diff
@@ -224,7 +224,7 @@ func (s *mutexStore) clear(ctx context.Context, mtx *activeMutex) (err error) {
 
 	// Unlock DB if the lock is no more used
 	if unlockDB {
-		if err = mtx.dbMutex.Unlock(ctx); err != nil {
+		if err = mtx.dbMutex.Unlock(ctx); err != nil && mtx.dbSession != nil {
```
Matovidlo, line 227: This caused panic on `mtx.dbSession == nil`

internal/pkg/service/common/etcdop/session.go
```diff
@@ -168,6 +168,11 @@ func (b SessionBuilder) Start(ctx context.Context, wg *sync.WaitGroup, logger lo
 			if err != nil {
 				s.logger.Infof(ctx, "cannot create etcd session: %s", err.Error())
 				if !errors.Is(err, context.Canceled) {
+					if errors.Is(err, context.DeadlineExceeded) {
+						// Do not check error from closeSession, it can be already closed
+						_ = s.closeSession(ctx, "etcd unreachable, remove existing sessions")
```
Matovidlo, line 173: On context deadline exceeded we fire `closeSession` as the ETCD is down and session life is 10 seconds

jachym-tousek-keboola (approved):

hosekpeter (approved):

PR #2313, opened 2025-04-01
fix: Distribution group does not panic

Matovidlo:
Jira: XXXXX

**Changes:**
- The distribution group panic occured when ETCD downtime is longer than 60 seconds, which terminates connection and it results into no nodes present in the mirror structure of distribution and panic of storage coordinator
- From now the termination of nodes or non presence of node results into single false statement and handover of node distribution without panicing

-----------


internal/pkg/service/common/distribution/assigner.go
```diff
@@ -74,16 +74,6 @@ func (a *Assigner) IsOwner(key string) (bool, error) {
 	return node == a.nodeID, nil
 }
 
-// MustCheckIsOwner method returns true, if the node is owner of the key.
-// The method panic if there is no node in the list.
-func (a *Assigner) MustCheckIsOwner(key string) bool {
```
Matovidlo, line 79: This is used in cleanup and operators, so we do not need this.
Also having panic in production code is unfortunate as we loss log information, request information, missing issue from metrics/telemetry.

internal/pkg/service/stream/storage/node/coordinator/fileimport/operator.go
```diff
@@ -156,7 +156,13 @@ func Start(d dependencies, config targetConfig.OperatorConfig) error {
 		).
 			// Check only files owned by the node
 			WithFilter(func(event etcdop.WatchEvent[model.File]) bool {
```
Matovidlo, line 158: The issue lays in WithFilter. It returns also old instances of files where the source is no longer in current `runtime` prefix of distribution group
image
This is current distrubtion group and this is log from WithFilter
```
start of getting node for key: 10012/870124/stream-exjtrgqw
end of getting node for key: 10012/870124/stream-exjtrgqw
{"level":"warn","time":"2025-04-01T10:43:36.154Z","message":"cannot check if the node is owner of the files based on source: no hosts added","nodeId":"stream-storage-coordinator-d88f67769-7thp8","component":"storage.node.operator.file.rotation"}
```
I susspect that `etcd.WithPrevKV()` returns also uncompacted information about file that is was in staging before and it is no longer accessible in runtime causing panic on `MustCheckIsOwner()`

Matovidlo: `etcd.WithPrevKV()` cannot be removed, it ends up on panic, that is misses SourceKey.

Matovidlo: Ok I get it
When I rollout deployment and we have old file e.g 
```
stream/storage/file/level/local/10012/870124/stream-exjtrgqw/test-sink/2025-04-01T10:18:54.679Z
```
Once the ETCD comes up it executes
https://github.com/keboola/keboola-as-code/blob/main/internal/pkg/service/common/distribution/node.go#L211
which removes nodes in which at one time we have from `n` hosts `0` hosts causing issue.
In a while it streams new nodes into assigner and makes it work again.
image
This image shows the new warning messages, then `found new node` comes and from that point the filter using `IsOwner()` function works again.

So this is expected behaviour of ETCD and we should not panic as the updates can come not right away after longer downtime but in batches.

Matovidlo: https://github.com/keboola/keboola-as-code/pull/2228/commits/3ac230183ed14ef166c9a406d442eedbf9186c8d here is old commit chaning the distribution group

jachym-tousek-keboola (approved):

hosekpeter (approved):
LGTM

PR #2320, opened 2025-04-04
Update dependencies and adjust .gitignore

hosekpeter:
Jira: XXXXX

**Changes:**
- Updated dependencies in `go.mod` and `go.sum` to their latest versions.
- Added `.vscode` directory to `.gitignore` for IDE configuration exclusion.

-----------


test/cli/push/variables-add-relation/expected-state.json
```diff
@@ -11,6 +11,7 @@
           "componentId": "keboola.variables",
           "name": "Variables",
           "description": "test fixture",
+          "changeDescription": "Updated from #KeboolaCLI",
```
hosekpeter, line 14: Change by @Matovidlo [PR](https://github.com/keboola/go-client/pull/160)

internal/pkg/utils/testproject/project.go
```diff
@@ -653,7 +653,7 @@ func (p *Project) prepareConfigs(ctx context.Context, grp *errgroup.Group, sendR
 
 			// Send request
 			_, err := p.keboolaProjectAPI.
-				CreateConfigRequest(configWithRows).
+				CreateConfigRequest(configWithRows, true).
```
hosekpeter, line 656: Change by @Matovidlo https://github.com/keboola/go-client/pull/160

Matovidlo: Yes, it is related to KB terraform provider where we need synchronous change of configuration as state and plan of terraform have to be consistent. This change would be breaking for CLI so I added support for sync/async option as CLI does not need to have that

Matovidlo (approved):
LGTM

jachym-tousek-keboola (approved):

PR #2327, opened 2025-04-08
feat: Split metacleanup

jachym-tousek-keboola:
Earlier today we deployed #2326 which seemed to have fixed the problem despite not changing anything significant. From the logs it seems that file cleanup wasn't running on that stack for several days. My hypothesis is that since we split metacleanup to to separate goroutines for jobs and files (because of the separate intervals), somehow the goroutine for files cleanup ended and wasn't running anymore.

While investigating the issues I found several things I wanted to improve:
1. Skipping file listing for sink the node wasn't owner of (this was changed in #2326)
2. Sink mirror map was suboptimal, it was mirroring files instead of just sinks.
3. I didn't like the way how file cleanup and jobcleanup were handled so I did it differently.

The last point might potentially fix the core issue because I effectively reverted the change that added the separate goroutines.


internal/pkg/service/stream/storage/node/coordinator/coordinator.go
```diff
@@ -39,5 +40,9 @@ func Start(ctx context.Context, d dependencies.CoordinatorScope, cfg config.Conf
 		return err
 	}
 
+	if err := jobcleanup.Start(d, cfg.Storage.MetadataCleanup); err != nil {
```
Matovidlo, line 43: Isn't job same metadata as file which is part of metacleanup?

provisioning/stream/common.sh
```diff
@@ -60,9 +60,9 @@ export STREAM_STORAGE_READER_CPU_SOFT_LIMIT="${STREAM_STORAGE_READER_CPU_SOFT_LI
 
 # Storage coordinator watches statistics and based on them, triggers slice upload and file import by modifying the state of the entity in the etcd.
 # Memory/CPU usage is minimal.
-export STREAM_STORAGE_COORDINATOR_MEMORY_SOFT_LIMIT="${STREAM_STORAGE_COORDINATOR_MEMORY_SOFT_LIMIT:="500Mi"}"
-export STREAM_STORAGE_COORDINATOR_MEMORY_HARD_LIMIT="${STREAM_STORAGE_COORDINATOR_MEMORY_HARD_LIMIT:="750Mi"}"
-export STREAM_STORAGE_COORDINATOR_CPU_SOFT_LIMIT="${STREAM_STORAGE_COORDINATOR_CPU_SOFT_LIMIT:="400m"}"
+export STREAM_STORAGE_COORDINATOR_MEMORY_SOFT_LIMIT="${STREAM_STORAGE_COORDINATOR_MEMORY_SOFT_LIMIT:="200Mi"}"
+export STREAM_STORAGE_COORDINATOR_MEMORY_HARD_LIMIT="${STREAM_STORAGE_COORDINATOR_MEMORY_HARD_LIMIT:="250Mi"}"
+export STREAM_STORAGE_COORDINATOR_CPU_SOFT_LIMIT="${STREAM_STORAGE_COORDINATOR_CPU_SOFT_LIMIT:="200m"}"
```
Matovidlo, line 65: Why we changed new limits? There is still bigger usage than 200MiB, we agreed with Halama that this can remain

Matovidlo (commented):
I am confused from naming jobcleanup and metacleanup which is effectivelly file cleanup

internal/pkg/service/stream/storage/node/coordinator/coordinator.go
jachym-tousek-keboola: I'll split the config part as well.

jachym-tousek-keboola:
See also https://github.com/keboola/kbc-stacks/pull/2844

Matovidlo (approved):

hosekpeter (approved):

PR #2334, opened 2025-04-15
feat: Use ectd helm chart defrag configuration

jachym-tousek-keboola:
Jira: https://keboola.atlassian.net/browse/PSGO-1054

**Changes:**
- Changed etcd defragmentation configuration from the helm chart

-----------

Blocked by:
- https://github.com/bitnami/charts/issues/33006

### DIFF
--- a/provisioning/common/etcd/values.yaml
+++ b/provisioning/common/etcd/values.yaml
@@ -115,167 +115,21 @@ livenessProbe:
 readinessProbe:
   initialDelaySeconds: 0 # replaced by startupProbe
 
-# Defragmentation, see extraDeploy.
-defragmentation:
+preUpgradeJob:
+  resources:
+    requests:
+      cpu: "100m"
+      memory: "64Mi"
+    limits:
+      memory: "64Mi"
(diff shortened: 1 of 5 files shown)


jachym-tousek-keboola:
The test for defrag job fails here: https://github.com/bitnami/charts/blob/d526c9a5b91e206e3c00f944c80178c1cf6ccd34/bitnami/etcd/templates/cronjob-defrag.yaml#L76

The line was added in this PR but i'm not quite sure why: 
https://github.com/bitnami/charts/pull/30096

The [commit](https://github.com/bitnami/charts/commit/558e971fd3951ea349cb14739920d3c948f84edb) message contains a vague note:
> We don't have an easy way to generate jwt tokens so for now it only works with
> .Values.auth.token.type = "simple".

But I don't understand why this is a problem only for the defragmentation job but not for preUpgradeJob or snapshotter.

Matovidlo (approved):
LGTM

hosekpeter (approved):
LGTM

PR #2335, opened 2025-04-15
build: Update dependencies

jachym-tousek-keboola:


go.mod
```diff
@@ -36,12 +36,12 @@ require (
...
 	github.com/google/shlex v0.0.0-20191202100458-e7afc7fbc510
 	github.com/hashicorp/yamux v0.1.2
-	github.com/jarcoal/httpmock v1.3.1
+	github.com/jarcoal/httpmock v1.4.0
 	github.com/joho/godotenv v1.5.1
 	github.com/jonboulle/clockwork v0.5.0
 	github.com/jpillora/longestcommon v0.0.0-20161227235612-adb9d91ee629
 	github.com/json-iterator/go v1.1.12
-	github.com/keboola/go-client v1.29.4
+	github.com/keboola/go-client v1.30.0
```
hosekpeter, line 44: can you include go-client version 1.31.0? https://github.com/keboola/go-client/releases/tag/v1.31.0

Matovidlo (changes requested):
How it is possible that we get rid of 347 libraries?
Is it connected to tools?

jachym-tousek-keboola:
> How it is possible that we get rid of 347 libraries?

Those were dependencies of goreleaser. But similar to golangci-lint it was causing issues so I changed it to install goreleaser similarly to how we install linter (not using go tool) which leads to removal of its dependencies from go.mod and go.sum.

hosekpeter (approved):

PR #2344, opened 2025-04-25
build: Disable free-disk-space

jachym-tousek-keboola:
Considering that we didn't run into out of space issues on CI lately, it's possible we no longer need this. And since the free disk space step can take up to a few minutes I think it's worth it to try and disable it to speedup the pipelines.

### DIFF
--- a/.github/workflows/test-k8s-service-stream.yml
+++ b/.github/workflows/test-k8s-service-stream.yml
@@ -40,16 +40,6 @@ jobs:
         with:
           ref: ${{ github.event.pull_request.head.ref }}
 
-      - name: Free Disk Space (Ubuntu)
-        uses: jlumbroso/free-disk-space@main
-        with:
-          android: true
-          dotnet: true
-          haskell: true
-          large-packages: false
-          docker-images: false
-          swap-storage: false
-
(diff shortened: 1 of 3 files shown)


.github/workflows/test-k8s-service-stream.yml
```diff
@@ -40,15 +40,15 @@ jobs:
...
-      - name: Free Disk Space (Ubuntu)
-        uses: jlumbroso/free-disk-space@main
-        with:
-          android: true
-          dotnet: true
-          haskell: true
-          large-packages: false
-          docker-images: false
-          swap-storage: false
+#      - name: Free Disk Space (Ubuntu)
```
hosekpeter, line 43: Do we want it to be commented there ?

jachym-tousek-keboola: I don't mind either way. Do you prefer if I just remove it?

hosekpeter: I would remove it, but you can merge it with this 👍

Matovidlo (approved):

hosekpeter (approved):
LGTM

PR #2359, opened 2025-05-13
build: Update dependencies

jachym-tousek-keboola:
The recent discovery that semver is [outdated](https://github.com/keboola/keboola-as-code/pull/2350) made me realize that there are likely other such packages. I'm adding [`gomajor`](https://github.com/icholy/gomajor) to our tools and updating such packages.

I didn't update `santhosh-tekuri/jsonschema` and `cenkalti/backoff` since they have some incompatible changes that will take a while to figure out.

### DIFF
--- a/Taskfile.yml
+++ b/Taskfile.yml
@@ -205,6 +205,7 @@ tasks:
       - go tool go-mod-upgrade
       - go mod tidy
       - go mod vendor
+      - go tool gomajor list
 
   install-golangci-lint:
     cmds:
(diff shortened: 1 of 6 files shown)


Taskfile.yml
```diff
@@ -205,6 +205,7 @@ tasks:
       - go tool go-mod-upgrade
       - go mod tidy
       - go mod vendor
+      - go tool gomajor list
```
Matovidlo, line 208: Maybe add this also to go skeleton monorepo ?

Matovidlo (approved):
LGTM

Taskfile.yml
hosekpeter: there is not go tool()

hosekpeter (approved):

PR #2365, opened 2025-05-21
build: Use backoff v5

jachym-tousek-keboola:
cenkalti/backoff has some incompatible changes so it took some effort to update to v5

### DIFF
--- a/internal/pkg/service/common/etcdop/watch_restart.go
+++ b/internal/pkg/service/common/etcdop/watch_restart.go
@@ -138,7 +138,6 @@ func wrapStreamWithRestart(ctx context.Context, channelFactory func(ctx context.
 			restart = true
 
 			// Delay is applied only if the restart is caused by an error, not by the manual restart
-			var delay time.Duration
 			if lastErr != nil {
 				// Calculate delay
 				restartDelay = b.NextBackOff()
@@ -150,7 +149,7 @@ func wrapStreamWithRestart(ctx context.Context, channelFactory func(ctx context.
 				select {
 				case <-ctx.Done():
 					return
-				case <-time.After(delay):
+				case <-time.After(restartDelay):
 					// continue
 				}
 			} else if subStream.cancelCause != nil {
--- a/internal/pkg/service/common/etcdop/watch_restart.go
+++ b/internal/pkg/service/common/etcdop/watch_restart.go
@@ -5,7 +5,7 @@ import (
 	"sync"
 	"time"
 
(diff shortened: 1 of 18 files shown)


internal/pkg/service/common/etcdop/watch_restart.go
```diff
@@ -44,7 +44,7 @@ func (s *RestartableWatchStreamRaw) SetupConsumer() WatchConsumerSetup[[]byte] {
 // An exponential backoff is used between attempts.
 // The operation can be cancelled using the context.
 func wrapStreamWithRestart(ctx context.Context, channelFactory func(ctx context.Context) *WatchStreamRaw) *RestartableWatchStreamRaw {
-	b := backoff.WithContext(newWatchBackoff(), ctx)
```
jachym-tousek-keboola, line 47: Not entirely sure but I think the context isn't needed here because it's handled by `case <-ctx.Done():` on line 150?

backoff.WithContext used this but it was removed in v5: https://github.com/cenkalti/backoff/blob/720b78985a65c0452fd37bb155c7cac4157a7c45/context.go#L55-L62

internal/pkg/service/common/etcdop/watch_restart.go
```diff
@@ -150,7 +149,7 @@ func wrapStreamWithRestart(ctx context.Context, channelFactory func(ctx context.
 				select {
 				case <-ctx.Done():
 					return
-				case <-time.After(delay):
+				case <-time.After(restartDelay):
```
jachym-tousek-keboola, line 152: The `delay` variable was always empty so it doesn't make sense. I think the intention was to use `restartDelay`.

internal/pkg/utils/backoff/time.go
```diff
@@ -0,0 +1,49 @@
...
+
+func (b *TimeBackoff) NextBackOff() time.Duration {
+	next := b.BackOff.NextBackOff()
+
+	if b.MaxElapsedTime != 0 {
+		elapsed := b.GetElapsedTime()
+		if elapsed+next >= b.MaxElapsedTime {
+			return backoff.Stop
+		}
+	}
```
Matovidlo, line 37:
```suggestion

	if b.MaxElapsedTime == 0 {
	    return next
	}
	
	elapsed := b.GetElapsedTime()
	if elapsed+next >= b.MaxElapsedTime {
		return backoff.Stop
	}
```

hosekpeter (approved):
LGTM

Matovidlo (approved):
LGTM

PR #2369, opened 2025-05-23
feat: Use jsonschema v6

jachym-tousek-keboola:


internal/pkg/encoding/json/schema/schema.go
```diff
@@ -123,44 +121,30 @@ func validateDocument(schemaStr []byte, document *orderedmap.OrderedMap) error {
 	return schema.Validate(document.ToMap())
 }
 
-func processErrors(errs []*jsonschema.ValidationError, parentIsSchemaErr bool) error {
+func processErrors(errs []jsonschema.OutputUnit) error {
```
jachym-tousek-keboola, line 124: Schema errors are now returned as a different type and don't go through here anymore.

internal/pkg/encoding/json/schema/schema.go
```diff
@@ -123,44 +121,30 @@ func validateDocument(schemaStr []byte, document *orderedmap.OrderedMap) error {
...
-			// Required field in a JSON schema should be an array of required nested fields.
-			// But, for historical reasons, in Keboola components, "required: true" is also used.
-			// In the UI, this causes the drop-down list to not have an empty value, so the error should be ignored.
-			if strings.HasSuffix(e.InstanceLocation, "/required") && e.Message == "expected array, but got boolean" {
-				continue
-			}
-			// JSON schema may contain empty enums, in dynamic selects.
-			if strings.HasSuffix(e.InstanceLocation, "/enum") && e.Message == "minimum 1 items required, but found 0 items" {
-				continue
-			}
```
jachym-tousek-keboola, line 162: Similar case as above, we should probably update the schema before validation but again I need a real example of a schema that would trigger this.

internal/pkg/encoding/json/schema/schema.go
```diff
@@ -123,44 +121,30 @@ func validateDocument(schemaStr []byte, document *orderedmap.OrderedMap) error {
...
+			if err := processErrors(e.Errors); err != nil {
+				formattedErr = err
 			}
-		case isSchemaErr:
-			// Required field in a JSON schema should be an array of required nested fields.
-			// But, for historical reasons, in Keboola components, "required: true" is also used.
-			// In the UI, this causes the drop-down list to not have an empty value, so the error should be ignored.
-			if strings.HasSuffix(e.InstanceLocation, "/required") && e.Message == "expected array, but got boolean" {
-				continue
-			}
```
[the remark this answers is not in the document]
jachym-tousek-keboola: Example component with this issue: https://apps-api.keboola.com/apps/kds-team.ex-activecampaign

internal/pkg/encoding/json/schema/schema.go
jachym-tousek-keboola: Fixed in [`b68a389` (#2369)](https://github.com/keboola/keboola-as-code/pull/2369/commits/b68a389d9f7a02b21588ff7d332d93fde7c4863f)

internal/pkg/encoding/json/schema/generate.go
```diff
@@ -75,24 +72,13 @@ func getDefaultValueFor(schema *jsonschema.Schema, level int) any {
...
-				values.Set(key, getDefaultValueFor(prop, level+1))
-			}
-		}
-		return values
+		return buildOrderedMap(schema, level)
 	case `string`:
-		switch schema.Format {
+		if schema.Format == nil {
+			return ``
+		}
```
Matovidlo, line 79: This seems weird, cannot we extract if condition before switching ? It seems very odd, that we have correct `firstType` but the `schema.Format` is incorrect from start.

internal/pkg/encoding/json/schema/schema.go
```diff
@@ -101,18 +105,54 @@ func ValidateContent(schema []byte, content *orderedmap.OrderedMap) error {
...
+		}
+
+		// JSON schema may contain empty enums, in dynamic selects.
+		if path.Last() == orderedmap.MapStep("enum") {
+			if enumArray, ok := value.([]any); ok && len(enumArray) == 0 {
+				if parentMap, ok := parent.(*orderedmap.OrderedMap); ok {
+					parentMap.Set("enum", []any{"placeholder"})
+				}
+			}
+		}
```
Matovidlo, line 144: cannot we create single condition instead of 3 nested if statments? E.g function that returns `bool` value

internal/pkg/encoding/json/schema/schema.go
Matovidlo: it's not available anymore

internal/pkg/encoding/json/schema/schema_test.go
```diff
@@ -212,9 +213,9 @@ func TestValidateObjects_InvalidSchema_InvalidType(t *testing.T) {
 	require.Error(t, err)
 	expected := `
 invalid JSON schema:
-- allOf failed:
-  - doesn't validate with "https://json-schema.org/draft/2020-12/meta/applicator#":
-    - "properties" is invalid: expected object, but got boolean
+- "file:///schema.json#" is not valid against metaschema: jsonschema validation failed with 'https://json-schema.org/draft/2020-12/schema#'
+  - at '': 'allOf' failed
```
Matovidlo, line 217: this seems odd with `''` empty string

hosekpeter:
Could you add a test with required:true to test/cli/validate?

internal/pkg/encoding/json/schema/schema.go
jachym-tousek-keboola: Ah, sorry... it's because I force pushed due to necessary rebase... It's fixed by the "Normalize schema" commit.

internal/pkg/encoding/json/schema/generate.go
jachym-tousek-keboola: To be honest I'm unsure if / how this can happen but the schema format can be nil in the new version.

Yes we can extract
```
if firstType == `string` && schema.Format == nil {
  return ``
}
```
before the switch but that makes even less sense than this to me.

jachym-tousek-keboola: `schema.Format == nil` doesn't mean that something is incorrect, just that the string value doesn't have any specific format (like date, time, email etc.) - so you know for example something like a name or description. Imo it doesn't make any sense to extract this.

internal/pkg/encoding/json/schema/schema.go
Matovidlo: can we merge at least 2 out of 3 if statments ? :) I understand that casting is necessary

jachym-tousek-keboola: Again yes we technically can, no I don't think it's a good idea:

```
if enumArray, ok := value.([]any); ok && path.Last() == orderedmap.MapStep("enum") && len(enumArray) == 0 {
  if parentMap, ok := parent.(*orderedmap.OrderedMap); ok {
    parentMap.Set("enum", []any{"placeholder"})
  }
}
```

Because this way the `enumArray, ok := value.([]any)` part is executed even when not needed and combining those two conditions together makes it ugly. And a separate function for this doesn't really make sense in my opinion.

Matovidlo: making functions out of 3 conditions makes perfect sense if you are able to say what this condition represents, ofc it can be 2 functions that `isNotEmptyEnum()` for example.

jachym-tousek-keboola: You know what, nevermind. While investigating it more I found out that an enum with no items is not actually considered a schema error so we don't need this code at all. So I removed this and added a test to make sure it won't become a schema error with a library update.

hosekpeter (approved):
LGTM

PR #2375, opened 2025-05-30
build: Update dependencies

jachym-tousek-keboola:
not updating:
- ETCD to 3.6 (separated to #2376)
- OTEL (not working on it yet)

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/keboola/keboola-as-code
 
-go 1.24.2
+go 1.24.3
 
 replace github.com/google/go-jsonnet => github.com/keboola/go-jsonnet v0.20.1-0.20240430105602-71646d8d4fa9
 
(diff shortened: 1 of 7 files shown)


internal/pkg/service/common/etcdop/watch_consumer_test.go
```diff
@@ -10,7 +10,7 @@ import (
 
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
-	"go.etcd.io/etcd/tests/v3/integration"
+	"go.etcd.io/etcd/tests/v3/framework/integration"
```
jachym-tousek-keboola, line 13: In v3.6 etcd moved their testing tools to different package.

go.mod
```diff
@@ -88,11 +88,11 @@ require (
 	go.opentelemetry.io/otel/sdk/metric v1.35.0
 	go.opentelemetry.io/otel/trace v1.35.0
```
jachym-tousek-keboola, line 89: There is otel 1.36.0 now but it seems they changed a lot of metrics and attributes so that update will take more time.

hosekpeter (approved):

PR #2394, opened 2025-07-16
fix: Ignore json schema errors

jachym-tousek-keboola:
Previously I ignored `"required": true`, which is invalid but Michal's comments said that it was used in our components.

However apparently we also have a case of `"required": false` here: https://github.com/keboola/component-custom-python/blob/main/component_config/configSchema.json#L155

In this PR I changed the behavior to throw away both.


Matovidlo (approved):
LGTM Include Vlado or prepare release notes in case we will release this

PR #2396, opened 2025-07-18
fix: Remove empty enums

jachym-tousek-keboola:
[Slack thread](https://keboolaglobal.slack.com/archives/C0441MZD52S/p1752827741803429)
* Remove empty `enum` field due to permisive validation of absent values
  * Values are typically found under `options.async` field which says that the `enum` values are fetched from `sync` action on component.
  * Currently we do not need to support sync action in CLI, so we remove this field
* Worst case scenario:
  * We have component which by mistake has set `enum: []` and we perform wrong validation.

### DIFF
--- a/internal/pkg/encoding/json/schema/schema.go
+++ b/internal/pkg/encoding/json/schema/schema.go
@@ -135,6 +135,15 @@ func NormalizeSchema(schema []byte) ([]byte, error) {
 				}
 			}
 		}
+
+		// Empty enums are removed, we're using those for asynchronously loaded enums.
+		if path.Last() == orderedmap.MapStep("enum") {
+			if arr, ok := value.([]any); ok && len(arr) == 0 {
+				if parentMap, ok := parent.(*orderedmap.OrderedMap); ok {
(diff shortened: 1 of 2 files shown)


internal/pkg/encoding/json/schema/schema.go
```diff
@@ -135,6 +135,15 @@ func NormalizeSchema(schema []byte) ([]byte, error) {
 				}
 			}
 		}
+
+		// Empty enums are removed, we're using those for asynchronously loaded enums.
+		if path.Last() == orderedmap.MapStep("enum") {
```
Matovidlo, line 140: Should we remove this `enum` field only when `options.async` are present? :thinking:

jachym-tousek-keboola: We could but it would be a quite a bit more complicated because the enum can be inside items and not necessarily always on the same level as options. I don't think it's worth it to be honest - we don't have a case where we'd need to keep it.

Matovidlo (approved):
LGTM

PR #2409, opened 2025-08-12
feat: Add GZIP compression middleware support

Matovidlo:
Jira: XXXXX

**Changes:**
- Reason: minimizes sizes of responses from templates API
- [Slack thread](https://keboolaglobal.slack.com/archives/CHPE0FPU5/p1751439098065779)

-----------


jachym-tousek-keboola (approved):
Somewhat odd to see the bash test but you did add normal go test too so fine I suppose.

PR #2457, opened 2025-11-11
feat: Add support for table list command

Matovidlo:
## Release Notes
Add new command `kbc remote table list` listing all tables in storage.

None.

## Impact analysis
minor - new command

## Change type
feature

## Justification
We need to list tables in order to fetch details/downloading it otherwise we need to go through interactive mode always

## Deployment
Merge & automatic deploy.

## Rollback plan
Revert of this PR.

## Post release support plan
None.

### DIFF
--- a/pkg/lib/operation/project/remote/table/list/operation.go
+++ b/pkg/lib/operation/project/remote/table/list/operation.go
@@ -39,10 +39,7 @@ func Run(ctx context.Context, d dependencies) (err error) {
 
 	logger.Info(ctx, "Found tables:")
 	for _, table := range tables {
-		bucketName := ""
-		if table.Bucket != nil {
-			bucketName = table.Bucket.BucketID.String()
-		}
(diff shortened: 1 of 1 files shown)


pkg/lib/operation/project/remote/table/list/operation.go
```diff
@@ -0,0 +1,50 @@
...
+		return err
+	}
+
+	tables := *tablesPtr
+	sort.Slice(tables, func(i, j int) bool { return tables[i].TableID.String() < tables[j].TableID.String() })
+
+	logger.Info(ctx, "Found tables:")
+	for _, table := range tables {
+		bucketName := ""
+		if table.Bucket != nil {
```
hosekpeter: Is it possible to have a table without bucket? I think it's not possible... Or?

Matovidlo:
I think it's not possible, it's always in the bucket. See SDK

PR #2480, opened 2025-12-02
Fixes for errors I discover when I try to sync PM playground project

vojtabiberle:
None.

## Impact analysis
Should only improve saving files with long path in unix/linux operation systems

## Change type
FIX

## Deployment
Merge & automatic deploy.

## Rollback plan
Revert of this PR.

## Post release support plan
None.

### DIFF
--- a/internal/pkg/naming/registry.go
+++ b/internal/pkg/naming/registry.go
@@ -104,7 +104,7 @@ func (r Registry) makeUniquePath(key model.Key, p model.AbsPath) model.AbsPath {
 	// Truncate the filename if it exceeds the maximum length, leaving room for suffix.
 	// Use UTF-8 safe truncation to avoid splitting multibyte characters.
 	if len(file) > maxFilenameLength-suffixReservedLength {
-		file = truncateUTF8(file, maxFilenameLength-suffixReservedLength)
+		file = truncateToBytes(file, maxFilenameLength-suffixReservedLength)
 		// Update the path with the truncated filename
 		p.SetRelativePath(filesystem.Join(dir, file))
 	}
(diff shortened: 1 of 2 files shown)


Matovidlo (approved):

internal/pkg/state/local/save.go
```diff
@@ -115,6 +115,13 @@ func (w *modelWriter) softDelete(path string) error {
 		return nil
 	}
 
+	// Remove existing .old file if it exists from a previous failed operation
+	if w.fs.IsFile(w.ctx, dst) {
+		if err := w.fs.Remove(w.ctx, dst); err != nil {
```
Matovidlo:
```suggestion
	if w.fs.IsFile(w.ctx, dst) && err := w.fs.Remove(w.ctx, dst); err != nil {
```

internal/pkg/naming/registry.go
```diff
@@ -101,7 +118,46 @@ func (r Registry) makeUniquePath(key model.Key, p model.AbsPath) model.AbsPath {
...
+		p.SetRelativePath(filesystem.Join(dir, normalized))
 	}
 	return p
 }
+
+// truncateUTF8 truncates a string to at most maxBytes bytes,
+// ensuring we don't split a multibyte UTF-8 character.
+// Note: The suffix format is limited to -999 (3 digits). If more than 999 paths
+// collide, subsequent paths will continue incrementing but may exceed the limit.
+func truncateUTF8(s string, maxBytes int) string {
```
Matovidlo: Naming is confusing, it sesms to be generally available as truncating is similar to do slicing of string which is by design UTF-8 encoded ?

vojtabiberle: correct, here is better naming: https://github.com/keboola/keboola-as-code/pull/2480/changes/31a8aa5b721e7a143a6c574a48ca78d31d61fb8e

PR #2580, opened 2026-04-13
fix(registry): skip naming.Attach() for rows with unresolved parent path [PSGO-233]

Matovidlo:
## Release Notes
- Fixes `kbc pull --force` crashing with a spurious "Cannot load remote state: naming error" when two configs in the same component have rows with identical names.

## Plans for customer communication
None.

## Impact analysis
- `internal/pkg/state/registry`: `registry.Set()` now skips `naming.Attach()` for objects whose `parentPathSet=false` (parent path not yet resolved). These objects are registered by the path generator later with the correct full path.
- No API or schema changes. No breaking changes.
- Only affects the `kbc pull --force` code path where `manifest.Load(ignoreErrors=true)` can leave rows with unresolved parent paths.

## Change type
Bug fix — `kbc pull --force` naming collision for configs with same-named rows [PSGO-233]

## Justification
SLSP's DevOps pipeline runs `kbc pull --force` against a project where two configs under `keboola.wr-azure-event-hub` both have a row named `besc-json-export`. The `--force` flag causes `manifest.Load()` with `ignoreErrors=true`, silently swallowing a `SetRecords()` early-return error. This leaves config rows with `parentPathSet=false`. `registry.Set()` then called `naming.Attach()` with the partial path `"rows/besc-json-export"` instead of the full resolved path, producing a false collision that aborted the pull.

## Deployment
Merge & automatic deploy.

## Rollback plan
Revert of this PR.

## Post release support plan
None.

[PSGO-233]: https://keboola.atlassian.net/browse/PSGO-233?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #2591, opened 2026-04-28
fix(rename): skip stateless objects in PathsGenerator to prevent panic [PSGO-233]

Matovidlo:
## Summary

- **Root cause:** `PathsGenerator.doUpdate` calls `objectState.LocalOrRemoteState()` unconditionally, which panics when a manifest entry exists for a config that has neither local nor remote state. This happens when an orphaned scheduler config remains in the manifest after its target orchestrator was deleted — the state entry is created from the manifest but never populated with actual data from either local files or the remote API.
- **Crash location:** `internal/pkg/model/objectstate.go:265` → `LocalOrRemoteState()` → called from `internal/pkg/state/local/paths.go:79` → `PathsGenerator.doUpdate` → `internal/pkg/plan/rename/rename.go` → `rename.NewPlan` → `sync/pull`
- **Why CI passes but local pull crashes:** CI runs always start from an empty git repository (no pre-existing manifest), so there are no stale entries. Client machines have an existing manifest that accumulated the orphaned scheduler entry.

## Changes

- `paths.go`: Added two guards to `doUpdate`:
  1. Return early when the object itself has neither local nor remote state (prevents `LocalOrRemoteState()` panic).
  2. Return early after a recursive parent update when the parent is also stateless (prevents path resolution on an unresolvable parent).
- `plan_test.go`: New test `TestRenameAllPlan_StatelessObjectSkipped` injects a stateless `ConfigState` into a loaded state and asserts `NewPlan` returns without error and excludes the orphan from rename actions.

## Relation to previous fixes

PR #2580 fixed the naming registry collision. PR #2585 fixed the manifest `SetRecords` fail-fast and orphaned-scheduler handling in `push`/`diff`. This PR fixes the remaining crash path that triggers during `kbc pull` when the rename plan builder encounters the stateless state entry that those earlier fixes may not have fully cleaned up.

## Test plan

- [x] `go test -race ./internal/pkg/plan/rename/... -run TestRenameAllPlan`
- [x] `go test -race ./internal/pkg/state/local/...`
- [ ] Manual: `kbc pull` on a project with an orphaned scheduler config in the local manifest — should complete without crashing

Fixes PSGO-233 / SUPPORT-15941 (follow-up crash after #2585)


internal/pkg/plan/rename/plan_test.go
```diff
@@ -83,6 +83,61 @@ func TestRenameAllPlan(t *testing.T) {
...
+	d.MockedHTTPTransport().RegisterResponder("GET", `=~/storage/components/ex-generic-v2`, getGenericExResponder.Once())
+	d.MockedHTTPTransport().RegisterResponder("GET", `=~/storage/components/keboola.ex-db-mysql`, getMySQLExResponder.Once())
+
+	projectState, err := d.MockedProject(fs).LoadState(t.Context(), loadState.Options{LoadLocalState: true}, d)
+	require.NoError(t, err)
+
+	// Inject an orphaned ConfigState: manifest entry exists but both Local and
+	// Remote are nil. This simulates a scheduler config referencing a deleted
+	// orchestrator, leaving a state entry that has no actual data.
+	orphanKey := model.ConfigKey{BranchID: 123, ComponentID: "keboola.scheduler", ID: "orphaned-01"}
```
hosekpeter: Test only covers the first check. The second check (after the recursive parent call) isn't triggered because BranchID: 123 isn't in state. Could you add a case where the child has state but its parent doesn't?

internal/pkg/plan/rename/plan_test.go
```diff
@@ -83,6 +83,61 @@ func TestRenameAllPlan(t *testing.T) {
...
+	}
+	require.NoError(t, projectState.State().Set(orphanState))
+
+	// NewPlan must not panic and the orphaned entry must be silently excluded.
+	plan, err := NewPlan(projectState.State())
+	require.NoError(t, err)
+
+	// The orphaned config must not appear in the rename actions.
+	for _, action := range plan.actions {
+		assert.NotContains(t, action.OldPath, "orphaned-01")
```
hosekpeter: Better to assert no action has Manifest.Key() equal to orphanKey instead of substring matching.

hosekpeter (commented):
Nit (out of scope):TestRenameAllPlan lines 51-59  action.Manifest = nil is set twice here. Could be cleaned up while we're here.

internal/pkg/state/local/paths.go
```diff
@@ -67,6 +67,12 @@ func (g *PathsGenerator) doUpdate(objectState model.ObjectState, origin model.Ke
 		return nil
 	}
 
+	// Orphaned manifest entries have no local or remote state; skip them so
+	// LocalOrRemoteState() does not panic and they are excluded from rename.
+	if !objectState.HasLocalState() && !objectState.HasRemoteState() {
+		return nil
+	}
+
```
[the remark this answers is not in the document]
Matovidlo: **Fixed** ✅ in commit 43292124d5a37e7b336fb89e6ea1eb465edbc9bf

Added `g.processed[objectState.Key().String()] = true` before the `return nil` in the first guard (stateless object check at entry). Without this mark, a stateless object encountered via multiple children would be traversed redundantly on each visit rather than short-circuiting after the first.

internal/pkg/state/local/paths.go
```diff
@@ -93,6 +99,11 @@ func (g *PathsGenerator) doUpdate(objectState model.ObjectState, origin model.Ke
 			return err
 		}
 
+		// Skip if the parent is stateless — its path cannot be resolved.
+		if !parent.HasLocalState() && !parent.HasRemoteState() {
```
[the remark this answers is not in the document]
Matovidlo: **Fixed** ✅ in commit 43292124d5a37e7b336fb89e6ea1eb465edbc9bf

Added `g.processed[objectState.Key().String()] = true` before the `return nil` in the second guard (stateless parent check). Note: we mark the *current* object (not the parent), since the parent was already marked by the first guard when its own `doUpdate` returned. This prevents re-processing the current child if it is referenced from multiple places.

internal/pkg/plan/rename/plan_test.go
```diff
@@ -83,6 +83,61 @@ func TestRenameAllPlan(t *testing.T) {
...
+		assert.NotContains(t, action.NewPath, "orphaned-01")
+	}
```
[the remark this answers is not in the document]
Matovidlo: **Fixed** ✅ in commit 43292124d5a37e7b336fb89e6ea1eb465edbc9bf

Replaced `TestRenameAllPlan_StatelessObjectSkipped` with `TestRenameAllPlan_StatelessParentSkipped`. The old test was vacuous: it injected a stateless orphan directly into state, but `Registry.All()` already filters out stateless objects before they reach `PathsGenerator`, so neither guard was ever exercised.

The new test reproduces the real PSGO-233 production scenario: a scheduler config with Remote state whose `SchedulerForRelation` points to a stateless orchestrator (orphaned manifest entry). `State.All()` includes the scheduler (has Remote state), `doUpdate` recurses into the stateless orchestrator via `MustGet`, the first guard fires and returns, then the second guard fires on the scheduler itself. Both guards are now exercised. Assertions use `action.Manifest.Key()` equality instead of substring matching.

internal/pkg/plan/rename/plan_test.go
Matovidlo: **Fixed** ✅ in commit 43292124d5a37e7b336fb89e6ea1eb465edbc9bf

The new `TestRenameAllPlan_StatelessParentSkipped` test directly covers this scenario: the scheduler child has Remote state (returned by `State.All()`), while its orchestrator parent is stateless (nil Local, nil Remote). The second guard in `doUpdate` fires when processing the scheduler after its parent traversal returns early. The test verifies neither the scheduler nor the orchestrator appear in the rename plan actions.

internal/pkg/plan/rename/plan_test.go
Matovidlo: **Fixed** ✅ in commit 43292124d5a37e7b336fb89e6ea1eb465edbc9bf

Changed assertions from `assert.NotContains(path, "orphaned-01")` substring checks to `assert.NotEqual(t, action.Manifest.Key(), orchKey)` and `assert.NotEqual(t, action.Manifest.Key(), schedulerKey)`. This ties the assertion directly to the struct identity of the injected objects rather than relying on path string matching, and works correctly even if path generation changes.

internal/pkg/plan/rename/plan_test.go
```diff
@@ -83,6 +83,85 @@ func TestRenameAllPlan(t *testing.T) {
...
+
+	require.NoError(t, projectState.State().Set(orchState))
+	require.NoError(t, projectState.State().Set(schedulerState))
+
+	// NewPlan must not panic; neither orphaned object should appear in rename actions.
+	plan, err := NewPlan(projectState.State())
+	require.NoError(t, err)
+
+	for _, action := range plan.actions {
```
romanbracinik, line 164: Ak by NewPlan vrátilo prazdny plan (regresna chyba), cyklus sa nikdy nevykona a test trivialne prebehne. Treba pridat overenie, 
  ze legitímna ex-generic-v2 rename akcia sa v plane stale nachadza — inak test nedokaze rozlisit "spravne skipol osirely objekt"
  od "skipol vsetko".

romanbracinik (changes requested):
ok fixol by som len ten test este

internal/pkg/plan/rename/plan_test.go
Matovidlo: **Fixed** ✅ in commit c6695dba434d0b3f1060edd403058f72f375d609

Added `require.Len(t, plan.actions, 3)` before the loop. This ensures the plan contains the fixture's 3 legitimate rename actions (branch, mysql config, mysql row) — if `NewPlan` regressed to returning an empty plan, the loop body would never execute and the `NotEqual` assertions would fire vacuously. The count of 3 matches `TestRenameAllPlan` for the same fixture.

romanbracinik (approved):
👌

PR #2596, opened 2026-05-11
feat(stream): expose OTLP base URL on source API response (Phase 3)

jordanrburger:
> **Stacked on top of #2595.** Targets `feat/stream-otlp-logs-phase1`; rebase to `main` after that merges.

## Summary

Surfaces the OTLP/HTTP endpoint URL alongside the existing stream URL on every source API response so the UI can display it without reconstructing it client-side. Closes the "Phase 3" backend scope from the design doc — the remaining Phase 3 items (one-click UI button, column-mapping JSON templates, quickstart docs) live in `keboola-ui` and the docs repos, not here.

## Changes

- **`Source.FormatOTLPSourceURL(publicURL)`** — mirrors `FormatHTTPSourceURL` but with the `/otlp/` path prefix. Returns the base URL only; the OTel SDK appends `/v1/logs`, `/v1/metrics`, `/v1/traces` itself when given `OTEL_EXPORTER_OTLP_ENDPOINT`.
- **`api.HTTPSource.OtlpURL` field** in the Goa design — regenerated server stubs, OpenAPI v3 / v2 JSON / YAML.
- **`mapper.NewSourceResponse` and `mapper.NewAggregationSourceResponse`** populate the new field.
- **15 E2E response fixtures** updated to include `otlpUrl` alongside `url` inside their `http` blocks.

## Example

```bash
$ curl https://stream.keboola.com/v1/branches/123/sources/my-source
{
  ...
  \"http\": {
    \"url\":     \"https://stream-in.keboola.com/stream/12345/my-source/abc...\",
    \"otlpUrl\": \"https://stream-in.keboola.com/otlp/12345/my-source/abc...\"
  },
  ...
}
```

User then sets one env var and any OTel SDK works:
```bash
export OTEL_EXPORTER_OTLP_ENDPOINT=https://stream-in.keboola.com/otlp/12345/my-source/abc...
```

## Test plan

- [x] `go build ./internal/pkg/service/stream/...` passes
- [x] `go vet` clean
- [x] Unit test for `FormatOTLPSourceURL` (canonical case, trailing-slash normalization, invalid-URL error path)
- [ ] E2E tests against API in CI (existing test suite — fixtures updated, mapper change wired through)
- [ ] Manual smoke: create a source via API, confirm both fields appear in the response

## Out of scope

- One-click "Create OTel Source" wizard in the Keboola UI (separate repo)
[...]

PR #2595, opened 2026-05-11
feat(stream): add native OTLP/HTTP endpoints (logs, metrics, traces)

jordanrburger:
## Summary

Adds a native **OTLP/HTTP** receiver to the Stream HTTP source so any OpenTelemetry SDK can export **logs, metrics, and traces** directly to Keboola — no Collector sidecar, no Docker, just an env var:

```bash
export OTEL_EXPORTER_OTLP_ENDPOINT=https://stream.keboola.com/otlp/<projectID>/<sourceID>/<secret>
```

The handler rides on the existing fasthttp server and dispatcher. It decodes the relevant OTLP request (protobuf or JSON, optional gzip), flattens the nested `resource → scope → record` tree into one `*orderedmap.OrderedMap` per record/data-point/span, and dispatches each through the standard pipeline via a new `recordctx.Context` implementation. The sink router, column renderer, and sinks are unchanged.

### How it fits together

```
POST /otlp/<projectID>/<sourceID>/<secret>/v1/{logs|metrics|traces}
        │
        ├─ DetectEncoding (Content-Type)
        ├─ DecompressIfGzip (Content-Encoding)
        ├─ Decode{Logs|Metrics|Traces} → pdata
        ├─ Flatten{Logs|Metrics|Traces} → []FlatRecord    (1 request → N records)
        └─ DispatchRecords
              │ for each record:
              │   recordctx.FromOTLP(...)
              │   dispatcher.Dispatch(...)                  (existing pipeline)
              └─ Build{Logs|Metrics|Traces}Response          (OTLP partial_success / 5xx)
```

### Signal coverage

| Signal | Route | One record per |
|---|---|---|
| Logs | `POST /v1/logs` | LogRecord |
| Metrics | `POST /v1/metrics` | data point (any of gauge / sum / histogram / exponential_histogram / summary) |
| Traces | `POST /v1/traces` | Span (events and links nested under the span) |

Metric type-specific fields (`value`, `count`, `bucket_counts`, `quantile_values`, `is_monotonic`, `aggregation_temporality`, …) are emitted only when they apply — Path columns fall back to `defaultValue` for fields that don't apply to a given metric type.

### Response semantics (OTLP-conformant)

| Outcome | Status | Body |
|---|---|---|
| All records dispatched OK | `200` | empty service response |
| Some records rejected | `200` | `partial_success` with rejected count (client should NOT retry) |
| All records rejected, retryable | `5xx` / `429` | top-level error (client SHOULD retry the batch) |
[...]

PR #2597, opened 2026-05-11
feat(stream): OTLP optimizations — zstd, header auth, parallel dispatch (Phase 4)

jordanrburger:
> **Stacked on top of #2595.** Targets \`feat/stream-otlp-logs-phase1\`; rebase to \`main\` after that merges. Independent of #2596 (Phase 3).

## Summary

Three self-contained optimizations on top of the OTLP receiver. Each is independently useful and reviewable; batched here to keep the PR count down.

### 1. zstd compression

\`DecompressIfGzip\` → \`DecompressBody\`. Now handles \`Content-Encoding: gzip\` and \`Content-Encoding: zstd\`. zstd is OTLP-optional and supported by every major OTel SDK (Java, Go, Python, JS, …). \`klauspost/compress\` was already a direct dependency, so no new import surface.

### 2. \`Authorization: Bearer\` header auth

New route shape **without** the secret in the URL:

```
POST /otlp/<projectID>/<sourceID>/v1/{logs,metrics,traces}
Authorization: Bearer <secret>
```

Existing URL-embedded form is unchanged for backward compat. Both are wired to the same handlers; \`parseAuthParams\` falls back to the header when the URL has no \`{secret}\` param.

**Why it matters:** the existing \`/stream/.../{secret}\` and Phase-1 \`/otlp/.../{secret}\` URL pattern leaks the secret into every layer that logs URLs — fasthttp access logs, ALB/CloudFront logs, Datadog APM \`http.url\` attributes, browser history if anyone ever pastes it. Bearer-header auth keeps the secret out of all of those.

**OTel SDK setup:**
```bash
export OTEL_EXPORTER_OTLP_ENDPOINT=https://stream-in.keboola.com/otlp/12345/my-source
export OTEL_EXPORTER_OTLP_HEADERS=\"Authorization=Bearer your-secret\"
```

### 3. Parallel dispatch with concurrency limit

\`DispatchRecords\` now overlaps up to **\`DefaultDispatchConcurrency = 8\`** records concurrently. Below the threshold (or for 1-record batches) the loop stays serial — no goroutine cost. \`DispatchRecordsWithConcurrency\` is exposed for tests and future config plumbing.

Per-record outcomes aggregate under a mutex. \`FirstError\` is the **lowest-index** rejection (deterministic across runs), not the first goroutine to fail, which would be racy.

**Why 8:** \`dispatcher.Dispatch\` is largely an etcd-mirror lookup plus per-sink pipeline writes. The bottleneck is downstream pipeline I/O, so a small concurrency cap is enough to overlap I/O without flooding sinks. Tunable later if profiling says otherwise.

## What changed

| File | Change |
|---|---|
| \`source/type/otlpsource/decode.go\` | \`DecompressIfGzip\` → \`DecompressBody\`; adds zstd branch |
| \`source/type/otlpsource/handler.go\` | \`parseAuthParams\` falls back to \`Authorization: Bearer <token>\` when no URL \`{secret}\` param; \`extractBearerToken\` helper |
| \`source/type/otlpsource/dispatch.go\` | Bounded-concurrency pool replacing serial loop; \`DispatchRecordsWithConcurrency\` for tests |
| \`source/type/httpsource/httpsource.go\` | Register the 6 new header-auth routes alongside the existing 6 URL-embedded ones |

## Tests

| Test | Covers |
|---|---|
| \`TestDecompressBody_NoCompression\` / \`_Identity\` | Pass-through for empty / unknown / \"identity\" encodings |
| \`TestDecompressBody_Gzip\` / \`_Zstd\` | Round-trip both compression formats |
| \`TestDecompressBody_InvalidGzip\` / \`_InvalidZstd\` | Malformed compressed input → error |
| \`TestExtractBearerToken\` | 11 cases: canonical, case-insensitive scheme, whitespace handling, malformed/empty/wrong-scheme |
| \`TestStatusCodeFromError_*\` | WithStatusCode interface unwrap; fallback to 500 |
| \`TestRecordOutcome_*\` | Nil is no-op; multi-error aggregation preserves FirstError + tracks worst status |

All green under \`-race\`. Full integration test of the parallel dispatch path requires a real \`dispatcher.Dispatcher\` (etcd-backed) — called out as follow-up.

## Test plan

- [x] \`go build ./internal/pkg/service/stream/...\` passes
- [x] \`go vet\` clean on \`otlpsource\`, \`httpsource\`
- [x] \`go test -race\` green for \`otlpsource\`
- [ ] \`task lint\` passes in CI
- [ ] \`task tests\` in CI (existing \`TestHTTPSource\` requires \`UNIT_ETCD_ENDPOINT\`)
- [ ] Manual smoke: zstd-compressed batch via Python OTel SDK; bearer-token batch via curl; high-volume batch (1000 records) to exercise the concurrency cap

## Out of scope / follow-ups

- **Batch dispatch** (one \`Dispatch\` call for N records) — would require changing the \`dispatcher.Dispatcher\` API and affects the \`/stream/\` path too. Worth its own design pass.
- **Configurable concurrency** — \`DispatchRecordsWithConcurrency\` is plumbing-ready, but no config wiring yet. Pending product input on whether to expose it.
- **\`Authorization: Basic\`** as an alternative to Bearer — not requested; OTel docs and most SDKs default to Bearer-style headers.

### DIFF
--- a/internal/pkg/service/stream/source/type/httpsource/httpsource.go
+++ b/internal/pkg/service/stream/source/type/httpsource/httpsource.go
@@ -153,6 +153,13 @@ func Start(ctx context.Context, d dependencies, cfg Config) error {
 	// HTTP source (same server, same dispatcher) — the only new pieces are
 	// route registration, OTLP decoding, record flattening, and OTLP-conformant
 	// response construction.
+	//
+	// Two route shapes are supported:
+	//   /otlp/<projectID>/<sourceID>/<secret>/v1/{logs,metrics,traces}
+	//     — secret embedded in the URL (convenient, matches /stream/...)
+	//   /otlp/<projectID>/<sourceID>/v1/{logs,metrics,traces}
+	//     — secret in "Authorization: Bearer <secret>" header (keeps the
+	//       secret out of HTTP access logs and APM URL attributes)
 	otlpHandler := otlpsource.New(ctx, logger, d.Clock(), dp, errorHandler)
 	router.Options("/otlp/<projectID>/<sourceID>/<secret>/v1/logs", otlpHandler.HandleOptions)
 	router.Post("/otlp/<projectID>/<sourceID>/<secret>/v1/logs", otlpHandler.HandleLogs)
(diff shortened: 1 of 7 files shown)


Matovidlo:
closing in favor of https://github.com/keboola/keboola-as-code/pull/2598

PR #2604, opened 2026-05-21
feat(apps-proxy): AJDA-2746 track WS activity per data frame instead of presence

sykora-ji:
[link to issue](https://linear.app/keboola/issue/AJDA-2746/track-app-activity-from-websocket-data-frames)

## Description

Replaces presence-based WebSocket activity tracking with per-frame activity tracking in `apps-proxy`.

**Removed.** The 30 s ticker goroutine in `AppUpstream.NewUpstream` that called `notify()` whenever `activeWsCount > 0`, the `activeWsCount atomic.Int64` field, and its `Add(1)` / `Add(-1)` bookkeeping in `newWebsocketProxy`.

**Added.** A new internal package `internal/pkg/service/appsproxy/proxy/apphandler/upstream/wsactivity/`:

- `frame.go` — a stateful, allocation-free RFC 6455 frame parser. State machine: `header0 → header1 → extLen → maskKey → payload → header0`. Discriminates control vs data via `opcode & 0x8` (data: 0x0–0x7, control: 0x8–0xF). Handles 7-, 16-, and 64-bit payload lengths and masked frames. Payload bytes are skipped via a `uint64` counter — no buffer is sized from the wire (payload length can be up to 2^63-1 per RFC).
- `conn.go` — `Wrap(io.ReadWriteCloser, FrameCallback) io.ReadWriteCloser` instantiates one parser per direction. Read feeds the server→client parser, Write feeds the client→server parser; bytes pass through unchanged.

**Integration in `upstream.go`.** `newWebsocketProxy` installs a `proxy.ModifyResponse` hook that, on 101 Switching Protocols, wraps `res.Body` — which `httputil.ReverseProxy.handleUpgradeResponse` type-asserts to `io.ReadWriteCloser` and uses as the upstream end of the bidirectional copy with the hijacked client conn. A single wrap captures both directions of the WebSocket stream. The wrapper invokes `u.notify(reqCtx)` once per non-control frame; the existing 30 s-per-app throttle in `notify.Manager` absorbs the burst (at most one Sandboxes Service PATCH per app per 30 s, regardless of frame rate). The existing `trace().GotConn` notify on the upstream handshake is preserved.

**Tests.**

- `wsactivity/{frame,conn}_test.go` — unit tests across all frame-header shapes (7/16/64-bit length, masked/unmasked), control-frame skipping, fragmented messages (one callback per frame), multi-frame buffers, byte-by-byte streaming, reserved opcode classification, fuzz test for parser panic safety, and `Wrap` pass-through / short-write / Close semantics.
- `proxy_test.go` — new `TestWebsocketActivityTracking` using a `clockwork.FakeClock`: opens an idle WS to a new `/ws-idle` endpoint, advances the fake clock past the 30 s throttle window, and asserts (a) notify count stays at 1 (regression test for the removed ticker) and (b) sending a real text frame past the window increments notify to 2 (proves the per-frame path is wired end-to-end).
- `testutil/api.go` — `sync.Mutex` around `DataAppsAPI.Notifications` plus `NotificationsCount` / `NotificationsSnapshot` getters, needed for race-free reads while the proxy is still running.
- `testutil/app.go` — `/ws-idle` endpoint that accepts a WebSocket and reads forever without writing any data frames.

## Release Notes

- **Justification**
    - A customer reported that the `autoSuspendAfterSeconds` setting was effectively disabled for Streamlit data apps with regular viewers: a forgotten browser tab keeps a WebSocket open, the 30 s ticker resets the activity timestamp on every iteration, and the app never suspends. Compute is billed for sessions nobody is using.
    - From now on, the proxy looks at actual WebSocket traffic. Idle tabs that only exchange protocol-level keep-alive pings stop counting as "active", so `autoSuspendAfterSeconds` becomes a usable cost control again.
    - User-visible behavior shift: an open WebSocket that only exchanges ping/pong frames will no longer prevent an app from auto-suspending. Any app that relied on the 30 s heartbeat to stay alive without emitting data frames must now emit real frames as a liveness signal. This should be flagged in the apps-proxy operator notes / changelog when the PR lands. The reporting customer is awaiting this fix and should be notified.
- **Impact Analysis**
    - Strict subset of the previous notify firing set (real activity ⊆ "WS is open"), so the only behavioral regression is for apps whose framework emits zero data frames between renders — undocumented and not a supported pattern.
    - Not behind a feature flag. The change is the explicit goal of the ticket and Pepa's spec; gating it would dilute the fix.
- **Deployment Plan**
    - Continuous deployment via the standard `apps-proxy` rollout — test on `dev-*` tag first, have PM verify on testing stack, then promote stack by stack.
- **Rollback Plan**
    - Trivial revert: the change is isolated to `apps-proxy` and contains no schema or data migration. Re-deploying the prior image restores the old presence-based tracking.
    - Watch for new tickets reporting "my app suspended unexpectedly" — those will identify frameworks that silently relied on the old behavior. Otherwise no additional monitoring beyond the existing apps-proxy SLOs.

### DIFF
--- a/internal/pkg/service/appsproxy/proxy/apphandler/upstream/upstream.go
+++ b/internal/pkg/service/appsproxy/proxy/apphandler/upstream/upstream.go
@@ -3,13 +3,13 @@ package upstream
 
 import (
 	"context"
+	"io"
 	"net/http"
 	"net/http/httptrace"
 	"net/http/httputil"
 	"net/url"
 	"strings"
 	"sync"
-	"sync/atomic"
 	"time"
 
 	"go.opentelemetry.io/otel/attribute"
@@ -22,6 +22,7 @@ import (
 	"github.com/keboola/keboola-as-code/internal/pkg/service/appsproxy/dataapps/notify"
 	"github.com/keboola/keboola-as-code/internal/pkg/service/appsproxy/dataapps/wakeup"
 	"github.com/keboola/keboola-as-code/internal/pkg/service/appsproxy/proxy/apphandler/chain"
+	"github.com/keboola/keboola-as-code/internal/pkg/service/appsproxy/proxy/apphandler/upstream/wsactivity"
 	"github.com/keboola/keboola-as-code/internal/pkg/service/appsproxy/proxy/pagewriter"
 	"github.com/keboola/keboola-as-code/internal/pkg/service/common/ctxattr"
 	"github.com/keboola/keboola-as-code/internal/pkg/service/common/servicectx"
@@ -50,13 +51,12 @@ type Manager struct {
 }
 
 type AppUpstream struct {
-	manager       *Manager
-	app           api.AppConfig
-	target        *url.URL // parsed from appsProxy.upstreamUrl at creation; nil when absent
-	handler       *chain.Chain
-	wsHandler     *chain.Chain
-	cancelWs      context.CancelCauseFunc
-	activeWsCount atomic.Int64
+	manager   *Manager
+	app       api.AppConfig
+	target    *url.URL // parsed from appsProxy.upstreamUrl at creation; nil when absent
+	handler   *chain.Chain
+	wsHandler *chain.Chain
+	cancelWs  context.CancelCauseFunc
 }
 
 type dependencies interface {
@@ -123,20 +123,9 @@ func (m *Manager) NewUpstream(ctx context.Context, app api.AppConfig) (upstream 
 	upstream.handler = upstream.newProxy(m.config.Upstream.HTTPTimeout)
 	upstream.wsHandler = upstream.newWebsocketProxy(m.config.Upstream.WsTimeout)
 
-	// Call notify while there is an active websocket connection
-	go func(ctx context.Context) {
-		for {
-			select {
-			case <-ctx.Done():
-				return
-			default:
-				if upstream.activeWsCount.Load() > 0 {
-					upstream.notify(ctx)
-				}
-				time.Sleep(30 * time.Second)
-			}
-		}
-	}(ctx)
+	// WebSocket activity tracking is per-frame (see newWebsocketProxy).
+	// No periodic ticker is needed — idle connections (only ping/pong) must
+	// not keep the app alive.
 
 	return upstream, nil
 }
@@ -255,6 +244,40 @@ func (u *AppUpstream) newWebsocketProxy(timeout time.Duration) *chain.Chain {
 		}
 	}
 
+	// Per-frame WebSocket activity tracking.
+	//
+	// httputil.ReverseProxy.handleUpgradeResponse type-asserts res.Body to
+	// io.ReadWriteCloser and uses it as the upstream end of a bidirectional
+	// copy with the hijacked client conn. Wrapping res.Body once therefore
+	// observes both directions of the WebSocket stream:
+	//
+	//   - Read on the wrapped RWC = server→client bytes (unmasked frames),
+	//   - Write on the wrapped RWC = client→server bytes (masked frames).
+	//
+	// wsactivity.Wrap parses frame headers in both directions and invokes
+	// the callback once per non-control frame. notify.Manager already
+	// throttles per app to one outbound call per 30s, so we can fire the
+	// callback per frame without flooding the Sandboxes Service.
+	proxy.ModifyResponse = func(res *http.Response) error {
+		if res.StatusCode != http.StatusSwitchingProtocols {
+			return nil
+		}
+		rwc, ok := res.Body.(io.ReadWriteCloser)
+		if !ok {
+			// httputil.ReverseProxy itself requires this assertion to succeed
+			// for a 101 response; if it doesn't, the proxy errors out anyway.
+			// We just skip wrapping and let the proxy report the error.
+			return nil
+		}
+		// Bind the callback to the request context so notify() can decorate
+		// its span/log with the right request attributes. notify() itself
+		// uses context.WithoutCancel, so the in-flight call survives the WS
+		// timeout and any per-request cancellation.
+		reqCtx := res.Request.Context()
+		res.Body = wsactivity.Wrap(rwc, func() { u.notify(reqCtx) })
+		return nil
+	}
+
 	return chain.
 		New(chain.HandlerFunc(func(w http.ResponseWriter, req *http.Request) error {
 			ctx := ctxattr.ContextWith(req.Context(), attribute.Bool(attrWebsocket, true))
@@ -263,9 +286,6 @@ func (u *AppUpstream) newWebsocketProxy(timeout time.Duration) *chain.Chain {
 
 			ctx, c := context.WithCancelCause(ctx)
 			u.cancelWs = c
-
-			u.activeWsCount.Add(1)
-			defer u.activeWsCount.Add(-1)
 
 			proxy.ServeHTTP(w, req.WithContext(ctx))
 			return nil
--- a/internal/pkg/service/appsproxy/proxy/apphandler/upstream/upstream.go
+++ b/internal/pkg/service/appsproxy/proxy/apphandler/upstream/upstream.go
@@ -144,13 +144,14 @@ func (u *AppUpstream) ServeHTTPOrError(rw http.ResponseWriter, req *http.Request
 		case isFrameworkBackgroundPoll(req.URL.Path):
 			// Auto-suspended app + framework background poll (e.g. Streamlit's
 			// /_stcore/health emitted by the frontend on its WS reconnect
-			// cycle while the tab stays open). Returning a wakeup here would
-			// defeat auto-suspend on every forgotten tab. Reply 503 with
-			// Retry-After so the client backs off — the user has to perform a
-			// meaningful action (refresh, click) to wake the app, which lands
-			// on a non-poll path and falls into the default branch below.
-			rw.Header().Set("Retry-After", "60")
-			rw.WriteHeader(http.StatusServiceUnavailable)
+			// cycle while the tab stays open). Triggering a wakeup here would
+			// defeat auto-suspend on every forgotten tab, so instead we serve a
+			// 503 with a plain-text message that the frontend shows in its
+			// connection modal ("paused due to inactivity, refresh to start").
+			// The user has to perform a meaningful action (reload) to wake the
+			// app, which lands on a non-poll path (GET /) and falls into the
+			// default branch below.
+			u.manager.pageWriter.WriteSuspendedPage(rw)
 		default:
 			u.wakeup(ctx, errors.Errorf("app state is %s", appInfo.ActualState))
 			u.manager.pageWriter.WriteSpinnerPage(rw, req, u.app)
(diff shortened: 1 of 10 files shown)


internal/pkg/service/appsproxy/proxy/apphandler/upstream/wsactivity/frame_test.go
```diff
@@ -0,0 +1,282 @@
...
+		buf = append(buf, 0xDE, 0xAD, 0xBE, 0xEF)
+	}
+
+	// Payload bytes (parser only walks past them via the counter).
+	if payloadLen > 0 {
+		payload := make([]byte, payloadLen)
+		for i := range payload {
+			payload[i] = byte(i)
+		}
+		buf = append(buf, payload...)
```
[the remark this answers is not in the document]
sykora-ji: This is a misread of the Go spec. `make([]T, n)` accepts any value of integer type for `n`, not specifically `int` — see the language reference: "The size arguments n and m must be of integer type, have a type set containing only integer types, or be an untyped constant." `uint64` is an integer type, so `make([]byte, payloadLen)` with `payloadLen uint64` compiles fine. The test suite for this file runs locally and on CI (`Unit Tests / linux-core` passed in this PR).

The defensive-allocation concern is also moot here: every call site of `buildFrame` passes a small literal (`5`, `10`, `200`, `65535`, `65536`). The only "huge" length appears in `TestFrameParser_ExtLen64_NoAllocation_HugeLen`, which constructs the header by hand and never goes through `buildFrame`.

internal/pkg/service/appsproxy/proxy/apphandler/upstream/wsactivity/conn_test.go
```diff
@@ -0,0 +1,215 @@
...
+	c := Wrap(rwc, func() { count++ })
+
+	n, err := c.Write(frame)
+	require.ErrorIs(t, err, io.ErrShortWrite)
+	assert.Equal(t, len(frame)/2, n)
+
+	// The first half includes only the partial header; the parser shouldn't
+	// have fired a callback yet (it needs the full header).
+	// Now simulate the caller retrying with the unwritten tail (no shortWrite
+	// this time so it succeeds completely).
```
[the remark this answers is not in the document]
sykora-ji: Good catch — the comment was inaccurate. For the 16-byte masked text frame used in the test, the 8-byte half-point lands one byte past the end of the header, so the callback does fire in the first chunk. The test still proves the no-double-feed invariant (a buggy `feed-full-p` implementation would re-walk mask + payload bytes on retry and fire spurious callbacks), but the original wording suggested otherwise. Reworded the comment in commit dd59ebe to describe what the test actually catches.

internal/pkg/service/appsproxy/proxy/apphandler/upstream/wsactivity/frame.go
```diff
@@ -0,0 +1,183 @@
...
+}
+
+// newFrameParser returns a parser that calls onDataFrame once per non-control frame.
+func newFrameParser(onDataFrame FrameCallback) *frameParser {
+	return &frameParser{onDataFrame: onDataFrame}
+}
+
+// Feed advances the parser through the bytes in p. It never reads or modifies
+// p beyond walking its bytes and never allocates based on payload length.
+func (fp *frameParser) Feed(p []byte) {
```
Matovidlo, line 78: This seems like C function, I cannot say it's not correct it does not feel like go syntax, too low things are done here checking code, masks etc. I feel like it should be somehow better done

sykora-ji: Fair point — looking at it again with fresh eyes, you're right. The byte-by-byte switch with 7 mutable bookkeeping fields and three near-duplicate "consume up to N bytes" loops did read like a C state machine.

Refactored in commit 7fd73e0 to a header-buffer + `binary.BigEndian` decode approach:

- The header is at most 14 bytes (2 base + 8 ext len + 4 mask key), so the parser now buffers the whole header into a fixed-size `[14]byte` array, then decodes it all at once via `binary.BigEndian.Uint16` / `binary.BigEndian.Uint64`.
- Implicit state lives in three counters (`headerLen`, `headerNeed`, `payloadLeft`) with two obvious phases — drain payload, accumulate header — instead of 5 explicit `parserState` values.
- Single `copy()` call replaces the three near-duplicate "consume up to N bytes" loops.
- No more `//nolint:gosec` annotations (header offsets are statically bounded, payload is skipped via `uint64` counter, no narrowing conversions).
- 4 struct fields instead of 7.

Semantics unchanged — all existing unit tests pass without modification except for one assertion that referenced the now-removed internal `state` field (updated to check the equivalent post-condition on the counters).

Matovidlo (approved):
LGTM

Matovidlo:
@sykora-ji ideally put it on canary/testing and retest the functionality of new frame activity parser

internal/pkg/service/appsproxy/proxy/apphandler/upstream/upstream.go
```diff
@@ -258,6 +258,40 @@ func (u *AppUpstream) newWebsocketProxy(timeout time.Duration) *chain.Chain {
...
+			// for a 101 response; if it doesn't, the proxy errors out anyway.
+			// We just skip wrapping and let the proxy report the error.
+			return nil
+		}
+		// Bind the callback to the request context so notify() can decorate
+		// its span/log with the right request attributes. notify() itself
+		// uses context.WithoutCancel, so the in-flight call survives the WS
+		// timeout and any per-request cancellation.
+		reqCtx := res.Request.Context()
+		res.Body = wsactivity.Wrap(rwc, func() { u.notify(reqCtx) })
```
MiroCillik, line 288: **Medium — per-frame goroutine + span churn; the 30s throttle does not absorb it.**

This callback fires `u.notify(reqCtx)` once per data frame, and `u.notify` *unconditionally* (a) spawns a goroutine via `u.manager.wg.Go`, (b) starts an OTel span `keboola.go.apps-proxy.upstream.notify`, and only then (c) calls `notify.Manager.Notify`, where the 30s throttle check actually lives (`notify.go:60`).

So the throttle caps the outbound *PATCH* but not the goroutine spawn or the exported span. `handleUpgradeResponse` copies with a 32 KB buffer and the parser fires once per frame within each chunk — a data app streaming small frames (live dashboard, chat, log tail) can produce thousands of frames per Read → thousands of goroutines + thousands of discarded spans per second. The old ticker was one call / 30s / app.

Suggestion: gate cheaply *before* `u.notify` — e.g. a per-upstream `atomic.Int64` next-allowed unix-nano compared against `clock.Now()`, only calling `u.notify` when the window has elapsed. The authoritative throttle in `notify.Manager` still guarantees correctness; this just keeps a cheap fast-path for the ~99.9% of frames that get throttled out. At minimum, benchmark on a high-frame-rate app before production.

MiroCillik (commented):
Reviewed the per-frame WS activity tracking. Architecture is clean and very well-tested (fuzz, split-feeds, masked/unmasked, short-write-no-double-feed). Build + `upstream`/`wsactivity` test packages pass locally. One performance item worth addressing (or at least benchmarking) before rollout, plus two minor cleanups — inline.

internal/pkg/service/appsproxy/proxy/apphandler/upstream/upstream.go
```diff
@@ -332,3 +373,24 @@ func (u *AppUpstream) Cancel(err error) {
...
+//
+// Currently covers Streamlit's /_stcore/health and /_stcore/host-config. These
+// are emitted on every WebSocket (re)connect — including the periodic ~20 min
+// reconnect cycle imposed by an external idle timeout — and would otherwise
+// either bump lastRequestTimestamp on a Running app (defeating auto-suspend)
+// or wake a Suspended one (defeating it again). Apps-proxy treats them as
+// non-activity: notify is skipped on a Running app and the request is rejected
+// with 503 Retry-After on a Suspended one, requiring the user to perform a
+// meaningful action (refresh, click into the UI) to wake the app.
+func isFrameworkBackgroundPoll(path string) bool {
```
MiroCillik, line 386: **Minor — hardcodes Streamlit-internal paths.** Exact-match is correctly tested (trailing-slash / query-string safe), but this couples apps-proxy to Streamlit's private `/_stcore/*` endpoints and will silently grow per framework. Consider framing it explicitly as a temporary/framework-specific workaround like the adjacent `Origin`-rewrite block already does, and note the long-term follow-up (server-side config / app-level beacon) so it doesn't quietly calcify.

internal/pkg/service/appsproxy/proxy/testutil/api.go
```diff
@@ -21,8 +22,36 @@ import (
...
+// while the proxy is still running).
+func (v *DataAppsAPI) NotificationsCount(appID string) int {
+	v.notifyLock.Lock()
+	defer v.notifyLock.Unlock()
+	return v.Notifications[appID]
+}
+
+// NotificationsSnapshot returns a copy of the current Notifications map under
+// a lock.
+func (v *DataAppsAPI) NotificationsSnapshot() map[string]int {
```
MiroCillik, line 47: **Minor — dead code.** `NotificationsSnapshot` is never called (only `NotificationsCount` is used by the new test). Exported test-helper methods won't trip the unused linter. Either drop it or use it — project guidance is to remove unused functions rather than keep them.

MiroCillik:
## Validation note: will this fully solve SUPPORT-16324 (GRPN forgotten-tab autosuspend)?

This PR correctly fixes the **stated root cause** — the presence ticker is gone, Tornado's `PING`/`PONG` are control frames (`opcode & 0x8 != 0`) and are ignored, and the `/_stcore/health` / `/_stcore/host-config` HTTP polls are filtered. So a **truly idle tab** (only ping/pong + health polls, user away) now produces zero activity notifies and autosuspend works. 👍

One caveat worth confirming **before we close the support ticket**, since the report specifically says *"logs show periodic Streamlit **reruns** from a tab held open"*:

- A Streamlit rerun pushes WebSocket **data frames** (`ForwardMsg`, binary opcode `0x2`) server→client.
- This PR counts data frames **in both directions** (per the spec), so **any** rerun — even one with no user behind it — counts as activity and resets the timer.
- Periodic reruns happen without user interaction in common cases: `st_autorefresh` / `streamlit-autorefresh`, `@st.fragment(run_every=...)`, a polling/`time.sleep` script loop, or the **~20 min WS reconnect cycle** (each reconnect re-runs the script → full re-render → data frames; see the `isFrameworkBackgroundPoll` comment that references this reconnect cycle).

**If the GRPN app's "periodic reruns" are app-driven (e.g. autorefresh) rather than user-driven, this change will not suspend it** — the server-pushed data frames keep it alive via the new path, just as the old ticker did.

Suggested follow-ups before declaring SUPPORT-16324 resolved:
1. Confirm what is producing the periodic reruns on that specific app (app-side autorefresh vs. genuine user). If app-driven, set expectations: this PR alone won't suspend it — that's the #3 (user-active beacon) / #4 (configurable idle mode) follow-up.
2. Verify the customer's `autoSuspendAfterSeconds` is shorter than the WS reconnect interval, otherwise reconnect-triggered reruns can still defeat it.

Possible refinement (design conversation, not a blocker — diverges from the bidirectional spec): counting **only client→server** data frames as activity would more closely match "a user actually did something" (genuine interaction → client `BackMsg`; app-driven refresh → only server `ForwardMsg`), at the cost of some edge cases (automatic client frames, initial server-side render).

internal/pkg/service/appsproxy/proxy/apphandler/upstream/upstream.go
sykora-ji: Fixed

internal/pkg/service/appsproxy/proxy/testutil/api.go
sykora-ji: Fixed

internal/pkg/service/appsproxy/proxy/apphandler/upstream/upstream.go
sykora-ji: Fixed

PR #2607, opened 2026-05-27
feat(stream): rotate OTLP/HTTP source secret without recreating the source [AI-3199]

Matovidlo:
## Release Notes
- Adds `POST /v1/branches/{branchId}/sources/{sourceId}/rotate-secret` to the Stream API
- Regenerates the source's 48-character secret in place and returns the updated `Source` synchronously (200) with the refreshed `otlp.url`/`secret` (or `http.url`)
- `sourceId` and the OTLP `baseUrl` stay stable, so producers only need the new secret/URL — no delete & recreate
- The previous secret stops working immediately (the ingest dispatcher mirrors source secrets from etcd)

None. (Frontend "Rotate secret" button is a separate sub-issue.)

## Impact analysis
- `api/stream/design.go` + regenerated `internal/pkg/service/stream/api/gen/...` and `openapi.*`: new `RotateSourceSecret` method (additive, no changes to existing endpoints)
- `internal/pkg/service/stream/api/service/source.go`: new synchronous handler; admin-only like the other source mutations; rotation is a single atomic `Source().Update` (version bump, no provisioning)
- No schema or dispatcher changes — the dispatcher already matches secrets from its etcd watch
- Fully backwards-compatible; no performance impact

## Change type
Feature — Secret rotation for OTLP/HTTP stream sources (AI-3199)

## Justification
OTLP/HTTP ingest URLs embed a 48-char secret in the path. If it leaks there is no way to rotate it today — the only recourse is delete + recreate, which loses the `sourceId`, drops destination tables, and forces reconfiguring every producer. This endpoint rotates the secret in place so leak recovery only requires distributing the new URL.

Linear: https://linear.app/keboola/issue/AI-3199

## Deployment
Merge & automatic deploy.

## Rollback plan
Revert of this PR.

## Post release support plan
None.


Matovidlo:
@claude review

jordanrburger (approved):

PR #2613, opened 2026-06-09
fix(cli): tolerate empty-array column definition in table download

Matovidlo:
## Release Notes
- Fixes `kbc remote table download` (and other commands that fetch table metadata) failing with `cannot decode JSON result: ... readObjectStart: expect { or n, but found [` after the table was already unloaded.
- Root cause is server-side: the Storage API now returns `"definition":[]` (an empty array) for columns of **untyped** tables instead of omitting the field. The fix is entirely client-side — no PHP/server change required.
- Bumps `github.com/keboola/keboola-sdk-go/v2` to the commit adding `Column.UnmarshalJSON`, which treats any non-object `definition` shape (null, empty, array) as "no definition" and decodes objects as before.

None.

## Impact analysis
- `go.mod` / `go.sum`: `keboola-sdk-go/v2` `v2.19.0` → `v2.20.1-0.20260609143011-f51d570a7be4` (commit pin; pseudo-version pending the SDK release).
- `go` directive raised `1.26.2` → `1.26.3` (required by the SDK module).
- Affects any CLI command that decodes table metadata (`remote table download`, table detail flows). Fully backwards-compatible: typed-column definitions decode exactly as before; only the previously-fatal empty-array shape is now tolerated.
- No API, schema, or runtime-config changes. Vendor dir is gitignored, so only `go.mod`/`go.sum` are tracked.

## Change type
Bug fix — robust parsing of untyped-table column definitions from the Storage API.

## Justification
A Storage API change started returning `"definition":[]` for untyped-table columns. The SDK type is `*ColumnDefinition` and the jsoniter decoder expected an object or null, so it aborted after the table had already been unloaded — leaving the command broken through no fault of the user. Per request, the fix belongs in the client (SDK) for robust parsing rather than reverting the server. SDK fix: https://github.com/keboola/keboola-sdk-go/pull/91

## Deployment
Merge & automatic deploy.

## Rollback plan
Revert of this PR.

## Post release support plan
None. Follow-up: once the SDK PR (https://github.com/keboola/keboola-sdk-go/pull/91) merges and is tagged, replace the commit pin with the tagged release to satisfy the no-pseudo-version policy.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/keboola/keboola-as-code
 
-go 1.26.2
+go 1.26.3
 
 replace github.com/google/go-jsonnet => github.com/keboola/go-jsonnet v0.22.1-0.20260427095904-b7e5c35127af
 
@@ -44,7 +44,7 @@ require (
 	github.com/json-iterator/go v1.1.12
 	github.com/keboola/go-cloud-encrypt v0.0.0-20250422071622-41a5d5547c43
 	github.com/keboola/go-utils v1.4.0
-	github.com/keboola/keboola-sdk-go/v2 v2.19.0
+	github.com/keboola/keboola-sdk-go/v2 v2.20.1
 	github.com/keboola/keboola-sdk-go/v2/transfer v1.1.0
 	github.com/klauspost/compress v1.18.5
 	github.com/klauspost/pgzip v1.2.6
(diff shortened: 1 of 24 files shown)


Matovidlo:
@claude review

zajca (approved):

PR #2621, opened 2026-06-22
feat(stream,templates): accept programmatic tokens via storage token exchange [PAT-1838]

Matovidlo:
## Release Notes
- The Templates API and Stream API now accept Connection programmatic tokens (`kbc_at_*` / `kbc_pat_*`) on endpoints that previously took only legacy Storage API tokens.
- A programmatic token is detected by prefix and exchanged for the target project's legacy Storage token via Connection's auth-bridge resolver (keboola-sdk-go `StorageTokenExchanger`), with the service authenticating itself using its projected Kubernetes ServiceAccount JWT.
- The target project is named by the new `X-KBC-ProjectId` request header, since programmatic tokens are not project-scoped on their own.
- Gated per service by the new `api.kubernetesTokenPath` config (path to the projected SA token); empty disables the feature, so existing deployments are unchanged.
- Bumps `keboola-sdk-go/v2` to `v2.21.0`.

None.

## Impact analysis
- `internal/pkg/service/common/dependencies`: new `IsProgrammaticToken` + `ExchangeProgrammaticToken` helpers (resolve via auth-bridge, build a `ProjectScope` directly from the resolved token detail — no redundant tokens/verify round-trip).
- `internal/pkg/service/common/httpserver/middleware`: new `ProjectIDHeader` (`X-KBC-ProjectId`) const + `ProjectIDFromHeader` accessor.
- `templates` & `stream`: `NewProjectRequestScope` branches on token kind; legacy Storage tokens keep the existing verify path unchanged. Master-token requirement is preserved on both paths.
- New config field `api.kubernetesTokenPath` on both services; defaults empty (feature disabled) — fully backwards-compatible.
- **appsproxy**: not changed — its `NewProjectRequestScope` is a no-op stub that never verifies a Storage token, so there is no project scope to exchange against.
- No DB migration, no performance impact on the legacy-token path (prefix check only).

## Change type
Feature — Accept Connection programmatic tokens via Storage token exchange (PAT-1838).

## Justification
Connection programmatic tokens (`kbc_at_*` / `kbc_pat_*`) are bearer tokens that cannot call legacy Storage-token APIs directly. This change lets the Templates and Stream APIs transparently exchange them for the project's Storage token through Connection's auth-bridge, mirroring the implementation already landed for query & metastore in the go-monorepo (PAT-1838). The feature is opt-in per service via the SA-token path config, so it is inert until the deployment provides the projected ServiceAccount token.

## Deployment
Merge & automatic deploy. Feature stays disabled until `api.kubernetesTokenPath` is configured with a projected ServiceAccount token whose mapping carries the `internal:auth-bridge:resolve-storage-token` scope.

## Rollback plan
Revert of this PR.

## Post release support plan
None. Once enabled, watch for auth-bridge resolver failures (exchange errors are logged; the SDK maps resolver status to the client HTTP status without exposing token material).

### DIFF
--- a/api/stream/design.go
+++ b/api/stream/design.go
@@ -114,7 +114,7 @@ var _ = Service("stream", func() {
 	Description("A service for continuously importing data to the Keboola platform.")
 	// CORS
 	cors.Origin("*", func() {
-		cors.Headers("Content-Type", "X-StorageApi-Token")
+		cors.Headers("Content-Type", "X-StorageApi-Token", "X-KBC-ProjectId")
 		cors.Methods("GET", "POST", "PATCH", "DELETE")
 	})
 
--- a/api/templates/design.go
+++ b/api/templates/design.go
@@ -88,7 +88,7 @@ var _ = Service("templates", func() {
 	Description("Service for applying templates to Keboola projects.")
 	// CORS
 	cors.Origin("*", func() {
-		cors.Headers("Content-Type", "X-StorageApi-Token")
+		cors.Headers("Content-Type", "X-StorageApi-Token", "X-KBC-ProjectId")
 		cors.Methods("GET", "POST", "PUT", "DELETE")
 	})
 
--- a/go.mod
+++ b/go.mod
@@ -44,7 +44,7 @@ require (
 	github.com/json-iterator/go v1.1.12
 	github.com/keboola/go-cloud-encrypt v0.0.0-20250422071622-41a5d5547c43
 	github.com/keboola/go-utils v1.4.0
-	github.com/keboola/keboola-sdk-go/v2 v2.20.1
+	github.com/keboola/keboola-sdk-go/v2 v2.21.0
 	github.com/keboola/keboola-sdk-go/v2/transfer v1.1.0
 	github.com/klauspost/compress v1.18.5
 	github.com/klauspost/pgzip v1.2.6
--- a/go.sum
+++ b/go.sum
@@ -601,8 +601,8 @@ github.com/keboola/go-oauth2-proxy/v7 v7.15.3-0.20260427095909-2cab86bcf028 h1:s
 github.com/keboola/go-oauth2-proxy/v7 v7.15.3-0.20260427095909-2cab86bcf028/go.mod h1:+YPrVbsB1rdZx5K45/ealm94rx9PwzSGK/uT5YwTKmU=
 github.com/keboola/go-utils v1.4.0 h1:WTyj95yrr8O8HxtC8TSTyUcElZiRGDeEdVvDpFo6HUo=
 github.com/keboola/go-utils v1.4.0/go.mod h1:IopwJzFz2gh0Yj3fUbIe2eamRoDKzbXvjqFjQyw3ZdQ=
-github.com/keboola/keboola-sdk-go/v2 v2.20.1 h1:R1qoJHZBXjuHyzlOorykJPtwJQ1yRbkxuL30hi7TOeY=
-github.com/keboola/keboola-sdk-go/v2 v2.20.1/go.mod h1:zpTnNSewMw0zlo7lHEdUKoFBocPOf05eq1sp0wGookU=
+github.com/keboola/keboola-sdk-go/v2 v2.21.0 h1:E3BARR14abNwc8rMEtNRLQNv9AloRXzMGCTSyEKUjxg=
+github.com/keboola/keboola-sdk-go/v2 v2.21.0/go.mod h1:zpTnNSewMw0zlo7lHEdUKoFBocPOf05eq1sp0wGookU=
 github.com/keboola/keboola-sdk-go/v2/transfer v1.1.0 h1:q+Ux+EOOqDIeDol3sce90pzC5jh3RdKMwfsWAjFgKWI=
 github.com/keboola/keboola-sdk-go/v2/transfer v1.1.0/go.mod h1:FHJQSxvveci565IVjBqQHMFMiYMVX3sA98z+xWu5fCY=
 github.com/keybase/go-keychain v0.0.1 h1:way+bWYa6lDppZoZcgMbYsvC7GxljxrskdNInRtuthU=
--- a/internal/pkg/service/common/dependencies/exchange.go
+++ b/internal/pkg/service/common/dependencies/exchange.go
@@ -1,0 +1,88 @@
+package dependencies
+
+import (
+	"context"
+	"net/http"
+	"strings"
+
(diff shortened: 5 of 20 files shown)


Matovidlo:
@claude review

zajca:
## Code review

Clean, well-documented feature. Architecture is sound: shared helper in `common/dependencies`, per-service feature gate, the legacy token path is untouched, and the master-token requirement is preserved on both paths (`ExchangeProgrammaticToken` calls `newProjectScope`, which enforces it — `project.go:102`). One blocking issue around HTTP status mapping, plus a few smaller items.

### 🔴 Blocking — exchange errors map to HTTP 500 instead of the proper status code

`StorageTokenExchangeError` (keboola-sdk-go) does **not** implement `StatusCode() int` — it only has a `StatusCode` *field* and a `ClientStatusCode() int` *method*. Our error writer (`internal/pkg/service/common/errors/statuscode.go:21` → `HTTPCodeFrom`) matches the `WithStatusCode { StatusCode() int }` interface (`errors/error.go:4`) via `errors.As`. Since `*StorageTokenExchangeError` doesn't satisfy it, **every exchange failure falls through to the default `http.StatusInternalServerError`**.

Effect (only the new programmatic-token path — the legacy `*StorageError` path correctly returns 401, unchanged):

| Case | Now | Expected |
|---|---|---|
| invalid programmatic token | 500 | 401 |
| token cannot access project | 500 | 403 |
| missing/invalid `X-KBC-ProjectId` | 500 | 400 |
| programmatic token, feature disabled | 500 ("not configured") | 400/401 |

Beyond the wrong status, 5xx in `WriteOrErr` generates an exception ID, logs at **Error** level, and returns the generic "contact support" message (`error.go:82,93`) — so ordinary client auth failures pollute error logs and alerting. The `exchange.go` doc comment even says callers should `errors.As` + `ClientStatusCode()`, but `resolveProjectScope` / `ExchangeProgrammaticToken` just return the error verbatim.

Suggested fix in `ExchangeProgrammaticToken` (benefits both services); `WrapWithStatusCode` already exists at `errors/statuscode.go:49`:

```go
res, err := exchanger.Exchange(ctx, subjectToken, projectID)
if err != nil {
    var exErr *keboola.StorageTokenExchangeError
    if errors.As(err, &exErr) {
        return nil, svcerrors.WrapWithStatusCode(err, exErr.ClientStatusCode())
    }
    return nil, err
}
```

The two other new error paths have the same problem and should also carry a status code:
- `resolveProjectScope`: `errors.Errorf("...missing valid %s header")` for a bad/missing `X-KBC-ProjectId` → should be **400**.
- `ExchangeProgrammaticToken`: `errors.New("...not configured")` when the feature is off → should be **400/401**, and it currently also logs at Error level.

### 🟡 Missing `X-KBC-ProjectId` in CORS allowed headers

`api/templates/design.go:91` and `api/stream/design.go:117` declare `cors.Headers("Content-Type", "X-StorageApi-Token")`. The new request header `X-KBC-ProjectId` is not listed, so a browser cross-origin request carrying it will fail the CORS preflight. Templates API is called from the UI, so this matters for browser clients. Add `X-KBC-ProjectId` to `cors.Headers` in both designs and regenerate.

### 🟡 No service-level test for the new programmatic branch

`exchange_test.go` covers `IsProgrammaticToken`, `connectionURL`, and the no-config gate — good. But there's no test for the programmatic branch in `resolveProjectScope` (programmatic token + header → exchange), nor for status-code mapping (a test there would have caught the blocking issue). Worth at least one test asserting an exchange failure surfaces the expected HTTP code.

### 🟢 Minor

- **Lost telemetry on the programmatic path.** `NewProjectDeps` sets span attributes (`keboola.project.id/name`, `token.id`, `token.is_master`) and a debug log (`project.go:84-93`). The exchange path goes straight to `newProjectScope` and skips them, so programmatic-token requests get a thinner trace. Consider setting at least project/token id from `res.Token`.
- **Config read inconsistency.** Templates reads `pubScp.APIConfig().API.KubernetesTokenPath`; stream adds a `KubernetesTokenPath()` accessor on `APIScope`. Works, explained in the PR — purely cosmetic.
- **`projectID` validation.** `strconv.Atoi` accepts `0` and negatives; the resolver rejects them, but an explicit `projectID <= 0` check would give a clean 400 locally (after the status-code fix).

### Looks good
- Master-token requirement preserved on both paths ✅
- Feature gate (empty path = disabled) → fully backwards compatible ✅
- SDK is careful with token material (`StorageToken` never logged, `Error()` carries no token) ✅
- `connectionURL` adds the scheme correctly; SA token read per request (rotation without restart) ✅
- appsproxy correctly untouched (no-op scope) ✅

**Verdict: NEEDS-WORK** — the status-code mapping isn't cosmetic; it changes API behavior (auth errors → 500 + error-level alerts) once the feature is enabled. After wiring `ClientStatusCode()` through (and ideally CORS + one test) this is good to merge.

Matovidlo:
Thanks @zajca, all addressed in the latest push.

**🔴 Blocking — exchange errors → HTTP 500**
Fixed in `ExchangeProgrammaticToken` (8117f1c0): on `*keboola.StorageTokenExchangeError` we now `errors.As` + `WrapWithStatusCode(err, exErr.ClientStatusCode())`, so invalid token → 401, no project access → 403, etc. map correctly instead of falling through to 500.
- **"not configured" path**: the feature-disabled branch now returns `WrapWithStatusCode(..., 400)` instead of a bare error — no more 500 + Error-level log / alert for an unsupported-auth request.
- **Missing/invalid `X-KBC-ProjectId`**: `resolveProjectScope` (both services) now wraps the error as 400, and additionally rejects `projectID <= 0` (covers your minor point — `strconv.Atoi` accepting `0`/negatives).

**🟡 CORS** — added `X-KBC-ProjectId` to `cors.Headers` in `api/templates/design.go` and `api/stream/design.go` and regenerated (server.go updated in a635681 / 12df7c04).

**🟡 Test** — `TestExchangeProgrammaticToken_NoConfig` now asserts `HTTPCodeFrom(err) == 400`, which directly guards the status-mapping regression (the disabled-feature row from your table). The full resolver-error path (401/403) needs an HTTP-mocked auth-bridge; left as a follow-up rather than bolting a mock server onto this PR.

**🟢 Lost telemetry** — the programmatic path now sets the same project/token span + ctx attributes as the legacy path. Extracted `setProjectScopeTelemetry(ctx, token)` in `project.go` and call it from both `NewProjectDeps` and `ExchangeProgrammaticToken`, so they share identical telemetry by construction (no copy-paste drift).

**🟢 Config read inconsistency** — left as-is per your note (cosmetic, explained in the PR).

zajca (approved):

PR #2626, opened 2026-07-03
PAT-1904 Fix flaky app auth config propagation

pepamartinec:
## Release Notes
https://linear.app/keboola/issue/PAT-1904/apps-flaky-auth-config

App auth config changes were taking a long time to propagate and were flaky — the same app would intermittently serve the old and the new auth config (e.g. password login vs. OIDC) depending on which apps-proxy replica handled the request, sometimes for hours. Two independent apps-proxy bugs kept a deploy's new config from converging across replicas.

First, the per-app config refresh ran on the incoming HTTP request's context, so a client disconnect — common during the OAuth redirect flow — aborted the refresh with `context canceled`, returned a `503`, and left that replica's cache un-updated. The refresh now runs on a context detached from the request (`context.WithoutCancel` + a bounded `refreshTimeout`), so a disconnect can no longer abort the shared cache refresh. Second, the auth handler was rebuilt only on a one-shot `modified` flag returned by the config loader for the single request that fetched a fresh `200`; if that request didn't rebuild the handler, the replica kept serving the stale auth handler indefinitely. Handler recreation is now keyed on config identity (the config `ETag`), matching how the upstream target is already tracked, and the now-dead `modified` plumbing is removed.

Reducing cross-replica divergence during the config TTL window itself is intentionally left to the planned CRD-based config-distribution refactor already in the backlog.

None.

## Impact analysis
No end-user impact beyond the fix itself: app auth config deployed to a running app now converges reliably within the ~60s config TTL on every apps-proxy replica instead of intermittently serving stale config. No API, config, or behavior changes elsewhere.

## Change type
Fix

## Justification
Fix flaky/slow propagation of app auth config changes (PAT-1904).

## Deployment
Merge & automatic deploy.

## Rollback plan
Revert of this PR.

## Post release support plan
None.

### DIFF
--- a/internal/pkg/service/appsproxy/dataapps/appconfig/appconfig.go
+++ b/internal/pkg/service/appsproxy/dataapps/appconfig/appconfig.go
@@ -21,8 +21,13 @@ import (
 // staleCacheFallbackDuration is the maximum duration for which the old configuration of an application is used if loading new configuration is not possible.
 const staleCacheFallbackDuration = time.Hour
 
+// refreshTimeout bounds how long a single config refresh may run.
+// The refresh holds the per-app lock, so this also bounds how long concurrent
+// requests for the same app wait for an in-flight refresh.
+const refreshTimeout = 30 * time.Second
+
 type Loader interface {
-	GetConfig(ctx context.Context, appID api.AppID) (config api.AppConfig, modified bool, err error)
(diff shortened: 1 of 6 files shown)


romanbracinik (approved):
pride mi to ok

PR #2642, opened 2026-08-04
feat(apps-proxy): UT-4805 redesign the data app pages as a four-state system

evviku:
## Release Notes

Redesigns the six pages an end user can hit when they open a data app.

Today those six pages share **two illustrations between them**: a flying rocket
means both "your app is starting" and "you are being signed in", and a crashed
rocket means both "the request failed" and "the app is stopped" — which is not a
failure at all. Customers have reported that the loading screen reads like a 404
error page.

This replaces them with one visual language and four states, each with its own
silhouette, its own motion rule, and its own page-level chrome:

| State | Pages | Motion | Top strip |
| --- | --- | --- | --- |
| Working | `spinner`, `redirect` | assembles, then perpetual — still happening | sweeping segment |
| Action | `login`, `selector` | settles in, then waits for you | none |
| At rest | `restart_disabled` | collapses once, then idles in place | none |
| Failed | `error` | breaks once, then freezes — terminal | static red edge |

The strip is a 3px band at the top edge, and how it behaves *is* the state, which
puts the motion rule where a user actually sees it rather than in an illustration
they would need a second state to decode. It replaces a per-state radial wash
that reviewers found cheap-looking — see "Justification" for why it was. Action
and at-rest get nothing, because on neither is anything in progress.

The sweep is deliberately **indeterminate**. The proxy learns one bit about a cold
start — whether the upstream still answers 503 — so a percentage or a filling bar
would be invented, and invented ones stall at 90% exactly when the user is already
anxious.

Also deletes `rocket.gif` (576 KB, `go:embed`-ed into the binary). It was being
displayed wrong anyway: a 500×750 portrait asset in a 200×145 box with
`object-fit: cover`, cropping away roughly half the illustration, which is why
its 120-frame animation was barely perceptible. Replacements are inline SVG,
a few KB in total.

Two further changes come from review feedback on this PR:

**The error page now explains the failure in words.** It used to lead with
`502 Bad Gateway` and offer only "quote the exception ID" — HTTP jargon, no
cause, no next step. Each status now maps to a plain-language title and a piece
of guidance, and a **Try again** button appears only where a retry could
actually succeed:

| Status | Title | Retry offered |
| --- | --- | --- |
| 401, 403 | You don't have access to this app | no |
| 404 | This app doesn't exist | no |
| 408, 429 | The app is busy right now | yes |
| 502, 503, 504 | This app isn't responding | yes |
| other 5xx | Something went wrong on our side | yes |
| other 4xx | This request couldn't be completed | no |

The code and its status text move down into the reference block beside the
exception ID, where support looks for them. `Technical details` still carries
the raw `Details`. The Streamlit plain-text fallback uses the same wording, so
the modal and the full page no longer explain one failure two ways.

**The starting page escalates as a cold start drags on.** It used to say the same
thing at second 90 as at second 3, so a long wait still read as stuck — the
original complaint in its slower form. Three stages on the poll that already
ticks once a second: the opening line, then "still starting, building from cold"
at 15s, then "taking longer than usual" with the support path at 45s. Wall clock
rather than a tick count, so a backgrounded tab still escalates on real time.

**The login page has a password reveal toggle.** An eye button inside the field,
swapping to a crossed-out eye while the password is showing.

**The exception ID is copyable.** The page asks the user to send it to support and
then left them transcribing 30-odd characters of hex by hand.

**Both cards lost their duplicated heading.** "This app is password protected" was
followed by a card titled `Basic Authentication` — the auth mechanism's config
name, not anything an end user recognises — and "Choose how to sign in" by
"Select a Provider". Login is now label, field, button.

**No new dependencies.** Two scripts are added: ~15 lines on the login page for
the reveal toggle and ~20 on the error page for the copy button. Both are
progressive enhancement — each button ships `hidden` and is only revealed once
the script confirms it can do the job (the copy button additionally requires
`navigator.clipboard`), so a client without JavaScript sees a plain password field
and plain selectable text rather than controls that do nothing. The spinner's
escalation lives inside the existing 503 poll, whose contract is unchanged; no
other page gained a script.

**Visual review surface** — all nine variants rendered from the templates on this
branch with the real stylesheet, live rather than as screenshots, so the entry
animations replay on reload and the password toggle is clickable in place:
https://claude.ai/code/artifact/c6548957-68a1-4165-9393-3e20b9e8d4c0

**Needed — this is a customer-visible change on every stack.**

These pages are shown to *end users of data apps*, which includes our customers'
own customers on apps that are shared externally. Nothing here changes behaviour,
URLs, auth, or status codes, so there is no action for anyone to take — but the
loading and error screens will look different from one deploy to the next.

Suggested: a short changelog entry noting that data app loading and error pages
have been redesigned for clarity, framed as a fix to the reported "loading screen
looks like an error" feedback. No advance notice or migration guidance required.

One thing worth saying out loud to anyone who asks: the pages are **not yet
themed per data app**. See "Justification" below.

## Impact analysis

**Blast radius:** every data app request that does not reach a running upstream —
so every cold start, every wake-from-sleep, every auth flow, and every error.
High visibility, low risk.

**Go changes:** no control flow, routing or status codes change. Two files:

- `error.go` gains `copyForStatus()`, a pure status → (title, guidance,
  retryable) mapping, and embeds its result in `errorPageData`. No branch in
  `WriteError` changes; the status the client receives is decided exactly as
  before.
- `renderer.go` calls the same mapping for the plain-text branch instead of the
  old fixed sentence.
- `proxy_test.go` — four string assertions, as before.

**What is untouched:**

- The 503 contract is identical: same status, same `Retry-After`, same 1s client
  poll, same `<noscript>` meta refresh.
- `suspended.go` and its plain-text response are untouched, including the
  Streamlit health-check workaround for AJDA-1935.
- `renderPlainText` keys on template *filenames*, and all six filenames are
  unchanged, so the plain-text fallback still resolves.
- Login POST target, password field `name`/`id`, `CsrfField` placement and the
  submit button `name` are all unchanged. The reveal button is `type="button"`,
  so it never submits the form, and it adds no field to the POST body.
- The 503 poll's interval, target and reload condition are untouched; the copy
  escalation only reads the clock and swaps text inside the existing callback.
- Provider links keep `href` as the first attribute — `proxy_test.go` asserts on
  that literal substring.
- All page `<title>` values are unchanged.
- `{{.Status}}`, `{{.StatusText}}`, `{{.ExceptionID}}` and `{{.Details}}` are all
  still rendered on the error page — the first two now in the reference block
  rather than the heading.

**Failure mode to be aware of:** `loadTemplates()` runs at service startup, so a
template parse error would be a boot failure rather than a cosmetic bug. This is
covered by `proxy_test.go`, which constructs the proxy and exercises these pages.

**Test coverage:** `copyForStatus` is covered by a table-driven unit test, and three
assertions now run against *rendered* bodies in cases that already exercise these
pages — the 403 explains itself in words, the 403 does **not** offer "Try again",
and the login form carries the reveal toggle and the autocomplete hint. Before
those, the templates were presentation no test looked at: deleting `{{ .Title }}`
or the reveal button passed CI green, because `loadTemplates` catches a parse
error at boot but not an empty heading.

**Verified locally:** `task tests-unit-appsproxy` — 243 tests, 0 failures.
`task lint` — no findings in the changed files, and `gofmt` is clean across
`internal/pkg/service/appsproxy/`. The password reveal was checked
in a browser against the rendered template: the input `type` flips, the
accessible name swaps, and the live region announces both states.

**Binary size:** −576 KB.

## Change type

- [x] Visual / UX change (customer-facing)
- [x] Asset removal
- [ ] Behaviour change
- [ ] API change
- [ ] Schema / data migration
- [ ] Configuration change

## Justification

Customer feedback, via UT-4805 (follow-up to PROF-161): the data app launch
screen visually reads as a 404 error page rather than an in-progress state.

The root cause was not the illustration quality but the vocabulary — six pages
sharing two images, with the loading page and the error page built from the same
layout, the same palette, the same 200×145 illustration slot and the same
metaphor. The only thing distinguishing "your app is launching" from "something
exploded" was whether the rocket was in one piece.

The fix is therefore a state system rather than a new picture. Each state is
distinguishable at a glance without reading, and the motion rules do real work:
a page that is still working never stops moving, and a page that has failed never
moves at all — so a frozen page now reliably means "stopped or failed", which is
exactly the signal that was missing.

**Why the state wash went.** Reviewers read the failed page's background as
cheap-looking, and the construction was the reason. It was a `position:fixed` 60vh
radial gradient at 13–16%, with no relationship to anything on the page: its
falloff landed at a different point relative to the content on every viewport
height, and it tinted the illustration's white plates on the way past. Diluting
`--danger` that far computes to `#FCE7E8` over white — a faded highlighter rather
than an alert — and to a muddy maroon over any dark `--app-bg`, which made it the
rule least likely to survive the theming work below. The four percentages were
also tuned so all four states read as *equally* present, which is the opposite of
ranking severity.

The gradient itself is back, as the strip's spill: anchored to the same top edge,
dead by a fixed 220px so it never reaches the illustration, at roughly half the
old intensity, and only on the two states that have a strip above it. A gradient
with nothing above it is unanchored decoration, which is what went wrong the first
time. Severity is now ranked structurally — two states carry page-level chrome,
two carry none.

**Deliberately out of scope — per-app theming.** The stylesheet is now built on
four custom properties that map 1:1 onto the values a data app already exposes in
its Streamlit config (`primaryColor`, `backgroundColor`, `secondaryBackgroundColor`,
`textColor`), and the defaults reproduce today's Keboola palette exactly — so
nothing changes visually until those values are actually injected. Plumbing them
through (`sandboxes-service` → `api.AppConfig` → `pagewriter.AppData`) is a
follow-up; the injection point is documented at the top of `assets/styles.css`.

Until that lands, a data app with a custom or dark Streamlit theme still gets the
Keboola palette on these pages — unchanged from today, but worth knowing.

**Accessibility:** `prefers-reduced-motion` is honoured on all six pages. Note
that reduced motion does *not* mean "no motion" on the working pages — a
completely frozen loading page is precisely what reads as an error, so those keep
an opacity-only pulse, and the progress sweep becomes a full-width bar pulsing in
place rather than travelling.

The starting page's lede is `aria-live="polite"`, so the escalation is spoken and
not only shown: a screen reader user waiting on a slow cold start is the one who
most needs telling that it is still going.

Illustrations are decorative and hidden from assistive tech with `aria-hidden`,
since each page's heading already states the state in words — a labelled image
would announce the same fact twice.

The error page's details use a native `<details>` element instead of the previous
CSS checkbox hack. Its status code and status text are separated by a real space
in the markup rather than a margin: a margin looks correct but leaves the text
content reading "502Bad Gateway" to a screen reader or a copy-paste.

The password reveal and the copy button are real `<button type="button">`s, so both
are reachable and operable by keyboard. The reveal's accessible name states the
*action* and swaps between "Show password" and "Hide password"; the resulting
*state* is announced separately by a visually-hidden `role="status"` region, rather
than relying on the reader to infer it from the label change. The copy button
announces its result the same way, and its tick reverts after two seconds — a
permanent tick stops reading as confirmation and starts reading as the button's
normal state. `autocomplete="current-password"` is set on the password field so
password managers recognise it.

**Copy review wanted.** The plain-language error titles and guidance in
`copyForStatus()` are the whole point of that change, and they are the part most
worth arguing about. They are deliberately written for someone who did not
deploy the app and may not be a Keboola user at all.

## Deliberately not in this PR

**Naming the signed-in account on a 403.** The commonest cause of "You don't have
access to this app" is being signed in to the wrong account — a personal Google
login rather than a work one — which the user cannot see and so cannot fix.

This was implemented and then backed out, because it cannot be done from
apps-proxy alone. The 403 that matters is raised inside `OAuthCallback`
(`AppError: "Invalid session: unauthorized"`), where the session is a local
variable: it is not on the oauth2-proxy request scope, and no session cookie was
saved, so there is nothing to read back. `ErrorPageOpts` carries only `Status`,
`RedirectURL`, `RequestID`, `AppError` and `Messages`, so the fork has to gain an
identity field first. Sequenced as: PR to `keboola/go-oauth2-proxy` → `go.mod`
bump → the page change. Nothing speculative was left behind here.

**An escape hatch out of the dead ends.** All six pages are terminal — no link
goes anywhere. Tracked in DES-83, because the right destination depends on
whether the reader can act on the failure (a project member) or only report it
(someone the app was shared with), and `AppConfig` has no owner or contact field
to point the second group at.

## Deployment

Merge & automatic deploy. Templates and assets are `go:embed`-ed, so this ships
with the normal apps-proxy release — no migration, no config, no coordination.

## Rollback plan

Revert of this PR. Self-contained: templates, one stylesheet, one deleted asset
and four test strings. No state, no schema, nothing persisted, no other service
depends on this markup.

## Post release support plan

None expected. If the pages render incorrectly on a specific stack the failure
would be visual and immediate rather than gradual, and revert is a single PR.

Worth a quick look after deploy at the loading page of any data app on a stack —
cold-start one from a sleeping state to see the working page, and hit a stopped
app to see the at-rest page.

---


Matovidlo (commented):
🥹 lovely rocket is gone.
Can you check suggested changes from copilot whether are relevant?

evviku:
Yes — checked all of them, and they were right on both counts.

**Five were the same accessibility point:** the state illustrations were exposed to
assistive tech via `role="img"` + `aria-label`, so a screen reader announced the
illustration and then the heading that says the same thing. Every mark is now
`aria-hidden="true" focusable="false"`, since each page states its state in words.

**The sixth was a real rendering bug** (it's in the collapsed "Suppressed comments"
block, so it's easy to miss): `<span class="code">{{ .Status }}</span>{{ .StatusText }}`
had no separator, so the heading read `503Service Unavailable` — `http.StatusText`
carries no leading space. Fixed, and then superseded: the code is no longer in the
heading at all. It sits in the reference block as `{{ .Status }} {{ .StatusText }}`
with a real space in the markup rather than a margin, so it survives copy-paste and
a screen reader.

Since your review the PR has also picked up a round of design and review feedback,
so the diff is quite different from the 08-04 version:

- the error page explains the failure instead of leading with `502 Bad Gateway`,
  and offers **Try again** only where a retry can actually succeed
- password reveal on the login form, and the exception ID is copyable
- the starting page escalates its wording at 15s and 45s, so a long cold start
  stops reading as stuck
- both cards lost their duplicated heading (`Basic Authentication` was the auth
  mechanism's config name, not something an end user recognises)
- the per-state background wash is gone — it read as cheap, and diluting a warm
  red to 13% computes to `#FCE7E8` on white and a muddy maroon on any dark
  background, so it was also the rule least likely to survive per-app theming.
  A 3px top strip replaces it: a sweeping segment while working, a static red
  edge on failure, nothing on the two states where nothing is in progress.
- three assertions now run against rendered bodies, because the templates were
  presentation no test looked at

Two things deliberately left out, both written up in the description: naming the
signed-in account on a 403 (needs an identity field on `ErrorPageOpts` in the
oauth2-proxy fork first — the session is a local variable at that denial, so
there's nothing to read from apps-proxy), and an escape hatch out of the dead-end
pages (DES-83).

`task tests-unit-appsproxy` is green at 243 tests. Ready for another look.

Matovidlo (approved):
LGTM

MiroCillik (approved):

PR #2645, opened 2026-08-05
feat(PSGO-37): apply API RowsSortOrder when loading remote config rows

Matovidlo:
## Release Notes
- When loading remote project state, config rows are now ordered according to the Storage API's `rowsSortOrder` field (if present), instead of always keeping whatever order the API's response happened to return

None. This only affects the in-memory/on-manifest row order the CLI operates with; it doesn't change any user-facing command behavior beyond making row order match what's set via the UI/API.

## Impact analysis
- `internal/pkg/state/remote/manager.go`: rows are reordered right after being loaded from the API, before anything else touches them. Rows not listed in `rowsSortOrder` keep their original relative position, appended at the end - so if the field is absent or only partially populated, behavior is unchanged from today.
- Orthogonal to the manifest's own `sortBy=id`/`sortBy=path` local file-ordering (`internal/pkg/state/manifest/records.go`), which continues to control on-disk layout unchanged - this only affects which order rows are considered "current" from the remote source.
- No breaking changes. Verified via new unit tests plus the existing `internal/pkg/state`/`internal/pkg/state/remote` suites (including the real-API `TestLoadRemoteState*` tests) - all pass unchanged.

## Change type
Feature - closes a gap identified during the PSGO-37 investigation (Linear issue's own TODO: "We should evaluate [rowsSortOrder] and it should be working")

## Justification
`Config.RowsSortOrder` has existed on keboola-sdk-go's `Config` struct since March 2025, but keboola-as-code never read it - rows were always kept in API-response order, with no way to honor an explicit reorder set via the UI/API. This was flagged as a follow-up in the PSGO-37 investigation into go-utils' OrderedMap coercion bug (separate PR: #2643) - implementing it here as its own scoped feature rather than bundling it with the bug fix.

## Deployment
Merge & automatic deploy.

## Rollback plan
Revert of this PR.

## Post release support plan
None.


Matovidlo:
@claude review

test/cli/cli_test.go
```diff
@@ -19,6 +19,14 @@ func TestCliE2E(t *testing.T) {
...
 		"push/config-data-gateway-push-and-push-dry-run": true,
 		"push/empty-data-gateway":                        true,
+		// The test project's Storage API now rejects creating classic "keboola.orchestrator"
+		// configs (HTTP 403 orchestrator.creationDisabled: "Please use Conditional Flows
+		// instead"). Unrelated to PSGO-37 - these fixtures need migrating to Conditional Flows.
+		"pull/orchestrator":         true,
+		"pull/orchestrator-partial": true,
+		"push/orchestrator":         true,
+		"push/orchestrator-partial": true,
+		"template-create/complex":   true,
```
pepamartinec, line 29: I've disable the flows migration feature on all `keboola-as-code` CI projects yesterday, try to drop this and run CI again.

pepamartinec (changes requested):
## The change has no observable effect

`applyRowsSortOrder` is applied to `config.Rows` at `manager.go:136`. That slice is a local variable in `LoadAll`; it is never stored. `loadObject(config.Config)` puts the inner `*model.Config` into the registry — `ConfigState.Remote` is `*Config`, not `ConfigWithRows` (`objectstate.go:219`). So the only thing the reordering feeds is the *sequence* of `loadObject(row)` calls at `manager.go:187`.

That sequence doesn't survive:

* `Registry.All()` re-sorts by `Manifest().SortKey(sortBy)` on every call (`registry/registry.go:88-91`)
* `Records.All()` / `SortRecords()` does the same before the manifest is serialised (`manifest/records.go:139-145`)
* `rename.NewPlan` feeds the paths generator from `b.All()` — already sorted

### Verified, not assumed

I ran the same load twice — identical input except `rowsSortOrder` — and diffed the serialised `.keboola/manifest.json`:

| naming template | sortBy | row names | manifest |
|---|---|---|---|
| with IDs | `id` | unique | **identical** |
| with IDs | `path` | unique | **identical** |
| without IDs (`sync init` default) | `id` | unique | **identical** |
| without IDs | `id` | **duplicate** | **differs** |

In the identical cases the `"rows"` array serialises as `1, 2, 3` even with `rowsSortOrder = [3,2,1]`.

The one case that differs is not the feature — it's the order-sensitive `-001` suffix from `naming.ensureUniquePath`, reached via the `changes.Loaded()` loop at `manager.go:313-320`:

```
 no rowsSortOrder                              rowsSortOrder=[3,2,1]
   "id": "1", "path": "rows/same-name"       →   "path": "rows/same-name-001"
   "id": "3", "path": "rows/same-name-001"   →   "path": "rows/same-name"
```

Worth noting `kbc sync init` writes `naming.TemplateWithoutIds()` (`sync/init/dialog.go:33`), so that is the default naming, not an exotic setup. It still needs two rows in one config whose names normalise identically — but the effect there is a silent path-assignment flip on fresh pulls, which is the opposite of the intent.

So these two lines in the description are not accurate as written:

> config rows are now ordered according to the Storage API's `rowsSortOrder` field ... instead of always keeping whatever order the API's response happened to return

The resulting state and manifest are identical either way.

> Verified via new unit tests plus the existing `internal/pkg/state`/`internal/pkg/state/remote` suites (including the real-API `TestLoadRemoteState*` tests) — all pass unchanged.

For a change of this kind, "all pass unchanged" is the signature of no effect rather than evidence of correctness — see below.

## Why nothing caught it

`grep -rln rowsSortOrder internal/pkg/fixtures test/` returns nothing. No fixture in the repo sets the field, so no test can distinguish the two code paths.

The assertion that *should* have caught this already exists — `load_remote_test.go:43`:

```go
assert.Equal(t, complexRemoteExpectedConfigsRows(envs), state.ConfigRows())
```

That is an ordered comparison of the rows in the result of `LoadAll`. It's blind here only because `complex.json` never sets `rowsSortOrder`, so both branches receive identical input. **Add `"rowsSortOrder"` to that fixture and update `complexRemoteExpectedConfigsRows` to the order you expect — that one change settles the question with no further discussion. If it goes green without touching the expectations, the code does nothing.**

Also worth stating: the four new unit tests exercise `applyRowsSortOrder` in isolation and pass whether or not it is wired to anything. They don't cover the claim in the description.

## Where the sort would actually matter

`internal/pkg/service/cli/cmd/template/create/dialog.go:104-109` makes its own `ListConfigsAndRowsFrom` call and builds rows via `model.NewConfigWithRows` — it never goes through `LoadAll`, so it doesn't get this reorder. And that path genuinely reads `Rows` in slice order: `templateIdsDialog.defaultValue()` renders the `### Row ...` lines in slice order, and `makeUniqueID` assigns template IDs in that order (so which of two same-named rows gets the `-001` suffix depends on it).

Three ways forward:

1. Move the reorder into `model.NewConfigWithRows` (`model/object.go:164`) — it already receives `*keboola.ConfigWithRows`, which carries `RowsSortOrder`. Then `template create` honours it and there's a user-visible behaviour to test.
2. Persist the order — a manifest field or a new `sortBy` mode — so it survives `SortRecords()`. Then the manifest diff is the test.
3. Drop the code and correct the description.

Any of these is fine; the current shape isn't, because it claims (1)/(2) and delivers neither.

## E2E skips

Covered in my inline comment — flows migration is now disabled on the `keboola-as-code` CI projects, so please drop the five skips and re-run rather than reworking the fixtures.

## Smaller points

**`applyRowsSortOrder` can return fewer rows than it received.** `byID` collapses duplicate IDs (last wins) and the fallback loop keys `used` by ID, so two rows sharing an ID whose ID appears in `sortOrder` would drop one row — and a dropped row never reaches `loadObject`, so it's absent from state entirely. Unique row IDs make this theoretical, but the failure mode is object loss rather than bad ordering. A `len(sorted) == len(rows)` guard is cheap.

**Comment reads backwards.** `// Order rows as returned by the API (RowsSortOrder), if set.` — "as returned by the API" describes the order being replaced. Suggest `// Apply the explicit row order from the API (rowsSortOrder), if set.`

**Test file package.** `rows_sort_order_test.go` is `package remote` while the neighbouring `manager_test.go` is `package remote_test`. Legal, but inconsistent; the four near-identical tests would also read better as one table, per CLAUDE.md.

**`rowsSortOrder` is read-only here** — nothing writes it back on push and it isn't stored in the manifest, so reorders don't round-trip. Worth saying explicitly in the description.

**Adjacent, pre-existing, not this PR:** `mapper/corefiles/remote_save.go:24` asserts `configState.RemoteState().(*model.ConfigWithRows)`, but `ConfigState.Remote` is always `*model.Config` — that assertion can never succeed, so the `rows` changed-field is never added. Same area; may deserve its own look.

---

*What I ran:* `go test ./internal/pkg/state/... ./internal/pkg/naming/...` (all pass except `internal/pkg/state`, which needs a real test project — environment limitation, not a defect), `go vet ./internal/pkg/state/remote/...` clean. The two-run manifest diff was a throwaway probe built on the existing `newTestRemoteUOW` harness in `manager_test.go`; happy to paste it if useful, though the `complex.json` fixture change above is the better home for it.

pepamartinec:
LOL, Claude se rozepsal :smile: Pointa - vypada to podle nej, ze ta uprava nebue mit efekt. Chtelo by to test, ktery zvaliduje, ze koncovy vystup je serazeny, jak se ocekava - tam by melo byt jasne videt, jestli se to spravne aplikuje nebo ne.

Matovidlo:
Closing per @pepamartinec's review: verified the reordering has no observable effect (Registry.All() / Records.SortRecords() unconditionally re-sort loaded objects by Manifest().SortKey(), discarding the insertion order this PR was trying to set), and template create's dialog.go — the one path that reads Rows in slice order — never went through LoadAll in the first place, so it wasn't fixed by this either. Dropped the code (see the last commit) per option 3 from his review.

The remaining diff was just the E2E orchestrator-creation-disabled test skips, which are already carried in #2643 — no net change to land once that merges, so closing here rather than keeping a now-empty PR open.

PR #2647, opened 2026-08-10
fix(stream): stop SlicePipeline.Close leaking unopened pipelines

Matovidlo:
## Release Notes
- Fixes a stream-service memory leak where `SlicePipeline.Close` could return without cancelling the async open-retry goroutine or notifying its parent, if it was called while the pipeline was still opening.
- Adds an explicit `closed` flag to `SlicePipeline` instead of overloading the nil-ness of the pipeline reference.
- Restores `Close`'s blocking idempotence for concurrent callers, which the leak fix had inadvertently dropped.

None.

## Impact analysis
- `internal/pkg/service/stream/storage/level/local/diskwriter/network/router/pipeline/pipeline_slice.go` and `internal/pkg/service/stream/storage/level/local/encoding/manager.go`.
- Root cause: `Close()` used `p.pipeline == nil` to mean both "already closed" and "still opening" (the async `tryOpen` retry loop hadn't set `p.pipeline` yet). In the latter case `Close()` returned immediately, skipping `cancel()`/`wg.Wait()`/`onClose()`. If the retry later succeeded, the resulting `encoding.Pipeline` stayed registered in `encoding.Manager.pipelines` forever with nothing left to close it — an unbounded heap leak, matching an observed `heap_objects` growth pattern (504K → 14.3M over 3.7 days) and elevated GC CPU (heap_alloc staying above next_gc). The leak window requires opens to be *failing* at the moment of close (`tryOpen` holds the lock across `OpenPipeline`, so `Close` only meets a "still opening" pipeline while the goroutine is sitting in the backoff retry loop) — it's rotation-driven conditional on open failures (e.g. writer nodes flapping or a deploy), not purely a function of rotation frequency independent of traffic/health.
- Fix: track `closed` explicitly under the existing `p.lock`, always run cancel→wait→(close underlying pipeline if opened)→onClose exactly once. `tryOpen` also checks `closed` before dialing, so a doomed open attempt after close doesn't allocate a pipeline just to discard it (less GC churn).
- `Close` can legitimately be invoked twice concurrently (once by `SinkPipeline` on rotation, once via the `closeFunc` callback the encoding pipeline calls on its own connection failure). The `closed` flag makes this explicit, and the losing caller now blocks on a `closeDone` channel until the winner's close actually finishes, instead of returning immediately — restoring the blocking-idempotence the old nil-check based `Close` had, which `router.go`'s `closeSyncer.Notify` (and the storage coordinator's `slicerotation` gate on it) relies on.
- `tryOpen` now returns whether it actually opened a pipeline, computed under the same lock as its `closed` check, and the retry goroutine gates `ready.NotifyReady()` on that instead of inferring it from `p.ctx.Err()` afterwards — the latter is set by `Close` before it even attempts the lock, so it could misreport a pipeline that genuinely opened as abandoned if `Close` raced in (e.g. via the encoding pipeline's own `closeFunc`) while `tryOpen` was still mid-`OpenPipeline`.
- `encoding/manager.go`: corrected a comment on the already-closing sentinel handling to reflect that `pipeline.Close` returns immediately without waiting for the in-flight close, rather than "closed either way".
- Added regression tests in `pipeline_slice_test.go` covering close-while-still-opening, concurrent-close blocking idempotence, and `tryOpen`'s already-closed short-circuit.
- No API changes, no migrations. Fully backwards-compatible.

## Change type
Bug fix — Fixes a heap/resource leak and a close-blocking-idempotence regression in the stream service's storage router.

## Justification
Investigation into elevated GC CPU (~4.5 GC cycles/sec) and monotonically growing `heap_objects` on the stream service source nodes traced back to `SlicePipeline.Close` returning early during the race between slice rotation (as fast as every 30s per sink, per the default `Trigger.Interval`) and the async pipeline-open retry loop, while opens were failing. Confirmed by reading `encoding.Manager.OpenPipeline`/`removePipeline`: entries are only removed via the pipeline's own `Close`, which never ran for pipelines abandoned mid-open. The leak requires concurrent open failures, not rotation frequency alone — flagged in review and reflected above and in the Post release support plan.

## Deployment
Merge & automatic deploy.

## Rollback plan
Revert of this PR.

## Post release support plan
Monitor stream source-node `heap_objects` / GC CPU trend after rollout; growth should flatten instead of increasing monotonically with sink count and slice-rotation frequency.

### DIFF
--- a/internal/pkg/service/stream/storage/level/local/diskwriter/network/router/pipeline/pipeline_slice.go
+++ b/internal/pkg/service/stream/storage/level/local/diskwriter/network/router/pipeline/pipeline_slice.go
@@ -38,6 +38,7 @@ type SlicePipeline struct {
 	wg     sync.WaitGroup
 
 	lock     sync.RWMutex
+	closed   bool
 	pipeline encoding.Pipeline
 }
 
@@ -134,37 +135,51 @@ func (p *SlicePipeline) WriteRecord(c recordctx.Context) (pipelinePkg.WriteResul
 }
 
 func (p *SlicePipeline) Close(ctx context.Context, cause string) {
-	p.lock.Lock()
-	defer p.lock.Unlock()
+	// Cancel the open-retry loop right away, without waiting for the lock, so an in-flight
+	// OpenPipeline attempt (if it respects ctx) can abort instead of running to completion.
+	p.cancel(errors.New("slice pipeline closed"))
 
-	// Stop if the pipeline is not opened
-	if p.pipeline == nil {
+	p.lock.Lock()
+	// Close may be called twice: explicitly by SinkPipeline and via the closeFunc callback
+	// the underlying encoding pipeline invokes on its own connection failure.
+	if p.closed {
+		p.lock.Unlock()
 		return
 	}
+	p.closed = true
+	pipeline := p.pipeline
+	p.pipeline = nil
+	p.lock.Unlock()
 
 	p.logger.Debugf(ctx, "closing slice pipeline: %s", cause)
 
-	// Cancel open loop, if running
-	p.cancel(errors.New("slice pipeline closed"))
+	// Wait for the open-retry goroutine to stop, whether it managed to open a pipeline or not.
 	p.wg.Wait()
 
-	// Close underlying encoding pipeline
-	ctx, cancel := context.WithTimeoutCause(context.WithoutCancel(ctx), 5*time.Minute, errors.New("slice pipeline close timeout"))
-	defer cancel()
-	if err := p.pipeline.Close(ctx); err != nil {
-		p.logger.Errorf(ctx, "cannot close slice pipeline: %s", err)
-	} else {
-		p.logger.Infof(ctx, "closed slice pipeline: %s", cause)
+	// Close the underlying encoding pipeline, if it was opened.
+	if pipeline != nil {
+		closeCtx, cancel := context.WithTimeoutCause(context.WithoutCancel(ctx), 5*time.Minute, errors.New("slice pipeline close timeout"))
+		defer cancel()
+		if err := pipeline.Close(closeCtx); err != nil {
+			p.logger.Errorf(closeCtx, "cannot close slice pipeline: %s", err)
+		} else {
+			p.logger.Infof(closeCtx, "closed slice pipeline: %s", cause)
+		}
 	}
-	p.pipeline = nil
 
-	// Notify parent SinkPipeline
+	// Notify parent SinkPipeline, even if the pipeline never finished opening.
 	p.onClose(ctx, cause)
 }
 
 func (p *SlicePipeline) tryOpen() error {
 	p.lock.Lock()
 	defer p.lock.Unlock()
+
+	// Already closed, e.g. by the encoding pipeline's own closeFunc callback, don't open a
+	// pipeline that would just have to be closed again the moment Close's wg.Wait returns.
+	if p.closed {
+		return nil
+	}
 
 	ctx := p.ctx
 
--- a/internal/pkg/service/stream/storage/level/local/diskwriter/network/router/pipeline/pipeline_slice.go
+++ b/internal/pkg/service/stream/storage/level/local/diskwriter/network/router/pipeline/pipeline_slice.go
@@ -37,9 +37,10 @@ type SlicePipeline struct {
 	cancel context.CancelCauseFunc
 	wg     sync.WaitGroup
 
-	lock     sync.RWMutex
-	closed   bool
-	pipeline encoding.Pipeline
+	lock      sync.RWMutex
+	closed    bool
+	closeDone chan struct{}
+	pipeline  encoding.Pipeline
 }
 
 // SliceData is part of the Slice model that is needed to create a SlicePipeline.
@@ -69,6 +70,7 @@ func NewSlicePipeline(
 		encoding:    encoding,
 		slice:       slice,
 		onClose:     onClose,
+		closeDone:   make(chan struct{}),
 	}
 
 	ctx = ctxattr.ContextWith(ctx, slice.SliceKey.Telemetry()...)
@@ -81,7 +83,8 @@ func NewSlicePipeline(
 		defer p.wg.Done()
 		for {
 			// Try open pipeline
-			if err := p.tryOpen(); err != nil {
+			opened, err := p.tryOpen()
+			if err != nil {
 				// Wait before retry
 				delay := b.NextBackOff()
 				p.logger.Warnf(p.ctx, "%s, waiting %s", err, delay)
@@ -93,8 +96,14 @@ func NewSlicePipeline(
 				}
 			}
 
-			// Pipeline is opened, close goroutine
-			ready.NotifyReady()
+			// tryOpen returns opened=false both when it found the slice pipeline already closed
+			// and when Close raced in while it was opening (see tryOpen's own closed check,
+			// evaluated under the same lock as the pipeline assignment) - only notify readiness
+			// if it actually opened a pipeline, or SinkPipeline.UpdateSlicePipelines could be
+			// unblocked for a slice pipeline that never actually opened.
+			if opened {
+				ready.NotifyReady()
+			}
 			return
 		}
 	}()
@@ -141,15 +150,20 @@ func (p *SlicePipeline) Close(ctx context.Context, cause string) {
 
 	p.lock.Lock()
 	// Close may be called twice: explicitly by SinkPipeline and via the closeFunc callback
-	// the underlying encoding pipeline invokes on its own connection failure.
+	// the underlying encoding pipeline invokes on its own connection failure. The loser waits
+	// on closeDone instead of returning early, so callers that rely on Close blocking until the
+	// pipeline is actually closed (e.g. router.go's closeSyncer.Notify, which gates the storage
+	// coordinator on all slice pipelines being closed) see accurate close-sync semantics.
 	if p.closed {
 		p.lock.Unlock()
+		<-p.closeDone
 		return
 	}
 	p.closed = true
 	pipeline := p.pipeline
 	p.pipeline = nil
 	p.lock.Unlock()
+	defer close(p.closeDone)
 
 	p.logger.Debugf(ctx, "closing slice pipeline: %s", cause)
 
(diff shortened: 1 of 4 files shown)


Matovidlo:
@claude review

internal/pkg/service/stream/storage/level/local/diskwriter/network/router/pipeline/pipeline_slice.go
```diff
@@ -134,38 +135,52 @@ func (p *SlicePipeline) WriteRecord(c recordctx.Context) (pipelinePkg.WriteResul
...
 
-	// Stop if the pipeline is not opened
-	if p.pipeline == nil {
+	p.lock.Lock()
+	// Close may be called twice: explicitly by SinkPipeline and via the closeFunc callback
+	// the underlying encoding pipeline invokes on its own connection failure.
+	if p.closed {
+		p.lock.Unlock()
 		return
 	}
```
pepamartinec, line 161: `Close` lost its blocking idempotence, and I think this is what commit 2 is really about.

Old `Close` held `p.lock` across `wg.Wait()` -> `pipeline.Close()` -> `onClose()`, so a second concurrent `Close` blocked for the whole duration and only returned once the close had actually finished. This one returns immediately.

That breaks close-sync. `router.go` calls `closeSyncer.Notify(ctx, changes.Header.Revision)` right after `onSlicesModification` -> `UpdateSlicePipelines` -> `slicePipeline.Close(...)`, commented "pipelines have been closed", and the coordinator gates on it at `slicerotation/operator.go:416` before reading statistics and moving the slice to upload. If the rotation-side `Close` is the loser, the source node reports the revision while chunks are still being flushed, and the coordinator proceeds with an unfinished slice.

It also explains the CI error commit 2 silences. `encoding.Pipeline.Close` has exactly two callers: `SlicePipeline.Close` and `Manager.close`. Shutdown callbacks run sequentially LIFO, and the encoding Manager is built before the router, so the router's callback (which closes and waits) runs first. The sweep can only meet a still-registered, mid-close pipeline if some `SlicePipeline.Close` returned before its encoding pipeline was closed - the new loser path. Commit 2 fixes the symptom.

Realistic trigger: a disk-writer node terminating concurrently with a rotation, i.e. every stream deploy.

Suggested fix - keep the close outside the lock, make the loser wait:

```go
// closeDone created in NewSlicePipeline
p.lock.Lock()
if p.closed {
    p.lock.Unlock()
    <-p.closeDone
    return
}
p.closed = true
pipeline := p.pipeline
p.pipeline = nil
p.lock.Unlock()
defer close(p.closeDone)
```

No self-deadlock risk: `p.Close` is the `closeFunc`, but every `onServerTermination` call site (`rpc/client.go:158,164,178`) is inside the dedicated `KeepAliveStream` goroutine, never on the winner's stack, and nothing on the winner's path waits for that goroutine. The cost is that the loser can block up to the 5-minute close timeout while holding `updateLock` - that was the old behaviour, and `WriteRecord`'s `RLock` is no longer held for the duration, which is the win worth keeping.

Worth re-running `TestKeboolaBridgeWorkflow` without commit 2 afterwards.

internal/pkg/service/stream/storage/level/local/diskwriter/network/router/pipeline/pipeline_slice.go
```diff
@@ -134,38 +135,52 @@ func (p *SlicePipeline) WriteRecord(c recordctx.Context) (pipelinePkg.WriteResul
...
+	// Close the underlying encoding pipeline, if it was opened.
+	if pipeline != nil {
+		closeCtx, cancel := context.WithTimeoutCause(context.WithoutCancel(ctx), 5*time.Minute, errors.New("slice pipeline close timeout"))
+		defer cancel()
+		if err := pipeline.Close(closeCtx); err != nil {
+			p.logger.Errorf(closeCtx, "cannot close slice pipeline: %s", err)
+		} else {
+			p.logger.Infof(closeCtx, "closed slice pipeline: %s", cause)
+		}
 	}
```
pepamartinec, line 182: Logging gap on exactly the path this PR is about: an unopened pipeline logs `"closing slice pipeline: %s"` at Debug but never the matching Info `"closed slice pipeline: %s"`, because that lives inside this `if pipeline != nil`. The abandoned-open case is what needs monitoring after rollout and it will be invisible at Info level.

internal/pkg/service/stream/storage/level/local/diskwriter/network/router/pipeline/pipeline_slice.go
```diff
@@ -134,38 +135,52 @@ func (p *SlicePipeline) WriteRecord(c recordctx.Context) (pipelinePkg.WriteResul
...
 
 func (p *SlicePipeline) tryOpen() error {
 	p.lock.Lock()
 	defer p.lock.Unlock()
 
+	// Already closed, e.g. by the encoding pipeline's own closeFunc callback, don't open a
+	// pipeline that would just have to be closed again the moment Close's wg.Wait returns.
+	if p.closed {
+		return nil
+	}
```
pepamartinec, line 203: `tryOpen` overloads its `nil` return: `nil` now means both "opened" and "abandoned because closed", and the retry loop can't tell them apart, so it calls `ready.NotifyReady()` for a pipeline that never opened.

Same line Copilot flagged, but its reachability claim doesn't hold - the `readyNotifier` is created per `UpdateSlicePipelines` call and handed only to pipelines created in that call, which aren't in the `Collection` until `Swap`, after the wait. Nothing can `Close` them during the window (`SinkPipeline.Close` would block on the `updateLock` the caller already holds), and after `Swap` the notifier has no reader. So it's a latent trap, not a live bug.

Its suggested fix (return a non-nil error) works but sends the goroutine through a spurious `Warnf` and a backoff round-trip before exiting via `<-p.ctx.Done()`. Cheaper: check `p.ctx.Err() == nil` before `NotifyReady()`.

internal/pkg/service/stream/storage/level/local/encoding/manager.go
```diff
@@ -160,7 +160,10 @@ func (m *Manager) close(ctx context.Context) error {
 	wg := &sync.WaitGroup{}
 	for _, w := range m.Pipelines() {
 		wg.Go(func() {
-			if err := w.Close(ctx); err != nil {
+			// The pipeline may have already been closed concurrently by its owning
+			// SlicePipeline (e.g. a slice rotation racing with this shutdown sweep).
+			// That's an expected outcome, not a failure - the pipeline is closed either way.
+			if err := w.Close(ctx); err != nil && !errors.Is(err, errPipelineAlreadyClosed) {
```
pepamartinec: The comment isn't accurate: `pipeline.Close` returns the sentinel *immediately*, without waiting for the in-flight close to finish. So this isn't "the pipeline is closed either way" - it's "someone else is closing it and we are not waiting for them". Keeping the sentinel is fine; please make the comment say that.

pepamartinec (changes requested):
Two things that aren't line-specific.

**No tests.** The `pipeline` package has zero `*_test.go` files and this PR adds none, for a concurrency fix in a leak path. Both regression tests are cheap and need no etcd:
- open fails -> `Close` during the backoff -> assert `encoding.Manager.Pipelines()` is empty and `onClose` fired;
- a second concurrent `Close` does not return before the first completes.

**The PR description overstates the trigger.** "an unbounded heap leak that grows with sink count and slice-rotation frequency (independent of traffic volume)" - `tryOpen` holds the lock across `OpenPipeline`, so the leak window also requires opens to be *failing* (`Close` landing while the goroutine sits in the backoff `select`). It's rotation-driven conditional on open failures, i.e. writer nodes flapping. This matters for the post-release plan: if `heap_objects` keeps climbing with healthy writer nodes, this wasn't the only cause.

internal/pkg/service/stream/storage/level/local/diskwriter/network/router/pipeline/pipeline_slice.go
Matovidlo: **Fixed** ✅ in commit bcc138ed93fcf8c16846a4a10b6324711327eada

Went with your suggested approach: added a `closeDone` channel, set as `p.closed = true` is set and closed via `defer close(p.closeDone)` right after the winner releases `p.lock`. The losing caller now does `p.lock.Unlock(); <-p.closeDone; return` instead of returning immediately, so it blocks until the winner's full teardown (wg.Wait -> pipeline.Close -> onClose) has actually finished. Added `TestSlicePipeline_CloseIsBlockingIdempotent`, which reproduces this by holding the winning Close inside onClose and asserting the losing Close does not return in that window - confirmed it fails against the pre-fix code (early-return version) and passes against the fix.

internal/pkg/service/stream/storage/level/local/diskwriter/network/router/pipeline/pipeline_slice.go
Matovidlo: **Fixed** ✅ in commit bcc138ed93fcf8c16846a4a10b6324711327eada

Moved the Info log outside the `if pipeline != nil` block: it now fires whenever the close completed without error, including the abandoned-open case (pipeline was nil), and stays suppressed only when `pipeline.Close` itself returned an error (already covered by the Errorf right above it, so no double-logging).

internal/pkg/service/stream/storage/level/local/diskwriter/network/router/pipeline/pipeline_slice.go
Matovidlo: **Alternative approach** 💡

Used your cheaper suggestion first (check before NotifyReady instead of returning an error from tryOpen), but a correctness pass on the diff found it was still an imprecise proxy: `p.ctx.Err()` is set by `Close` *before* `Close` even attempts `p.lock`, so it can go non-nil while `tryOpen` is still mid-`OpenPipeline` for a call that ends up genuinely succeeding (e.g. `Close` invoked concurrently via the encoding pipeline's own `closeFunc`) - the check would then wrongly skip `NotifyReady()` for a pipeline that did open. Fixed by having `tryOpen` itself return `opened bool`, computed under the same lock as the `closed` check (`false` at the early-return, `true` once `p.pipeline` is assigned), so the retry goroutine no longer infers anything from state a concurrent `Close` can independently mutate. Added `TestSlicePipeline_TryOpenSkipsAlreadyClosed` pinning the `opened=false`/no-side-effect contract for the closed-early-return path.

internal/pkg/service/stream/storage/level/local/encoding/manager.go
Matovidlo: **Fixed** ✅ in commit e7842f47514a7b8c79162353abbbfbee78b5ff1d

Reworded: the sweep now says the pipeline may already be closing (initiated by its owning SlicePipeline), and that `pipeline.Close` returns the sentinel immediately without waiting for that in-flight close to finish - so this is deferring to whoever's already closing it, not confirming it's closed either way.

Matovidlo:
Addressing the two non-line items from the review:

**No tests.** Added `pipeline_slice_test.go` with three tests, matching the two you asked for plus one more that came out of a follow-up correctness pass on the fix itself:
- `TestSlicePipeline_CloseIsBlockingIdempotent` — a second concurrent `Close` blocks until the first (winning) call fully finishes.
- `TestSlicePipeline_CloseWhileStillOpening` — `Close` during the open-retry backoff waits out the goroutine, leaves no pipeline reference, and still fires `onClose`.
- `TestSlicePipeline_TryOpenSkipsAlreadyClosed` — covers the `tryOpen` fix from the [ready.NotifyReady() thread](https://github.com/keboola/keboola-as-code/pull/2647#discussion_r3775614137).

**PR description.** Corrected the Impact analysis/Justification to state the leak requires concurrent open failures (not rotation frequency alone), and updated the Post release support plan note accordingly.

Commits: bcc138ed9 (Close idempotence + tryOpen + tests), e7842f475 (manager.go comment).

pepamartinec (approved):

PR #2650, opened 2026-08-21
fix(stream): bound chunk write retries and abandon unprocessable chunks on giveup

Matovidlo:
## Release Notes
- `stream-http-source` (and any other encoding pipeline) now gives up on a chunk write that has been failing continuously for 5 minutes instead of retrying forever, and closes the affected slice pipeline
- Fixes an unbounded goroutine/memory leak: a slice pipeline whose writes to a disk-writer node kept failing persistently previously never closed, leaking its retry goroutine, buffers, and `encoding.Manager` registration indefinitely
- Addresses review feedback: a discarded/abandoned chunk could previously be reported back to the client as successfully written; the pipeline could stay marked "ready" (and keep receiving new records) past give-up; and a slowly-draining (but still making progress) pipeline could be killed by the same 5-minute window

None.

## Impact analysis
- `internal/pkg/service/stream/storage/level/local/encoding`: `processChunks` retry loop now bounded to 5 minutes of continuous failure (`encoding/config.Config.MaxChunkRetryDuration`, configurable, default 5m), then closes the pipeline via the existing `closeFunc` callback (called asynchronously to avoid deadlocking `pipeline.Close`'s `chunksWg.Wait`)
- `chunk.Writer`: new `Abandon()` method discards unprocessable completed chunks and unblocks `WaitAllProcessedCh`
- `pipeline`: a new sticky `gaveUp` flag makes `Flush` fail fast and report an error once give-up has happened, instead of a later `Flush`/`Sync` call (e.g. from the syncer during `Close`) either hanging on its own real-time 30s timeout or - worse - being woken by `Abandon`'s notifier close and reporting a discarded chunk as successfully written. This also means `Close` no longer hangs ~60s post-give-up, so it no longer risks serializing `SinkPipeline.UpdateSlicePipelines` behind a stuck slice's close
- `pipeline.ready` is now set to `false` unconditionally at give-up (previously gated on `FailedChunksThreshold`), so the balancer stops routing new records to the pipeline immediately instead of after the async close completes
- The give-up window now tracks whether a round made any progress (fewer completed chunks after than before), not only whether it was fully clean - a link draining most-but-not-all chunks each round no longer gets killed and its buffer discarded; the retry backoff is reset in that case too
- One new config field: `encoding/config.Config.MaxChunkRetryDuration` (`duration.Duration`, default 5m, validated 1s-1h) - same pattern as the existing `FailedChunksThreshold`, backwards-compatible with a default, no other config/API/storage schema changes
- Confirmed via Datadog metrics on `kbc-eu-central-1`: one `stream-http-source` pod showed unbounded linear goroutine growth (27k -> 115k+ over 3 days) and RSS growth (262MB -> 900MB+, trending straight to an OOMKill) with no plateau, while its sibling pod stayed flat - this fix bounds that leak

## Change type
Bug fix — stream-http-source OOM caused by an unbounded goroutine/memory leak in the encoding pipeline's chunk-write retry loop

## Justification
`stream-http-source` pods on `kbc-eu-central-1` were trending toward OOMKill from unbounded goroutine growth. Root-caused to `processChunks` retrying a failing network write forever with no ceiling - a pipeline whose writes to a disk-writer node kept failing persistently never closed, so every affected slice rotation permanently leaked a goroutine, its buffers, and its `encoding.Manager` registration. This bounds the retry to a configurable duration (default 5 minutes) and makes the pipeline give up and close cleanly, without silently reporting discarded data as written or leaving the pipeline routable in the meantime.

## Deployment
Merge & automatic deploy.

## Rollback plan
Revert of this PR.

## Post release support plan
Watch `container.memory.rss` and `runtime.go.num_goroutine` (grouped by `pod_name`/`host`) for `stream-http-source` on `kbc-eu-central-1` post-deploy to confirm the leak is gone - growth should flatten instead of climbing linearly.

### DIFF
--- a/internal/pkg/service/stream/storage/level/local/encoding/chunk/writer.go
+++ b/internal/pkg/service/stream/storage/level/local/encoding/chunk/writer.go
@@ -202,6 +202,23 @@ func (w *Writer) ProcessCompletedChunks(fn func(chunk *Chunk) error) error {
 	return nil
 }
 
+// Abandon discards all completed, not yet processed chunks and unblocks WaitAllProcessedCh.
+// It is used when the processor (see ProcessCompletedChunks) gives up and will never process
+// them, so a caller waiting for them to be processed (see Flush) isn't left blocked forever.
+func (w *Writer) Abandon() {
+	w.lock.Lock()
+	defer w.lock.Unlock()
+
+	if len(w.completedChunks) == 0 {
+		return
+	}
+
+	w.freeChunks(w.completedChunks...)
+	w.completedChunks = nil
+	close(w.allProcessedNotifier)
+	w.allProcessedNotifier = make(chan struct{})
+}
+
 // freeChunks after it is no longer used.
 func (w *Writer) freeChunks(all ...*Chunk) {
 	for _, v := range all {
--- a/internal/pkg/service/stream/storage/level/local/encoding/pipeline.go
+++ b/internal/pkg/service/stream/storage/level/local/encoding/pipeline.go
@@ -466,6 +466,7 @@ func (p *pipeline) Close(ctx context.Context) error {
 
 func (p *pipeline) processChunks(ctx context.Context, clk clockwork.Clock, encodingCfg encoding.Config) {
 	b := newChunkBackoff()
+	var failingSince time.Time
 	for {
 		// The channel is unblocked if there is an unprocessed chunk,
 		// or all chunks have been processed and the writer is closed.
@@ -520,12 +521,28 @@ func (p *pipeline) processChunks(ctx context.Context, clk clockwork.Clock, encod
 				p.readyLock.Unlock()
 			}
 
+			// Give up after writes have been failing continuously for too long, instead of
+			// retrying forever - a permanently unreachable volume would otherwise leave this
+			// goroutine, and the pipeline it belongs to, running until the process restarts.
+			if failingSince.IsZero() {
+				failingSince = clk.Now()
+			} else if clk.Since(failingSince) > maxChunkRetryDuration {
+				p.logger.Errorf(ctx, "chunks write failed for over %s, giving up: %s", maxChunkRetryDuration, err)
+				// Nobody else will ever process these chunks, don't leave a future Flush call
+				// (e.g. from the syncer, during Close) waiting forever for them.
+				p.chunks.Abandon()
+				go p.closeFunc(ctx, "chunk write retries exhausted")
+				return
+			}
+
 			// Wait before retry
 			delay := b.NextBackOff()
 			p.logger.Warnf(ctx, "chunks write failed: %s, waiting %s, chunks count = %d", err, delay, cnt)
 			<-clk.After(delay)
 			continue
 		}
+
+		failingSince = time.Time{}
 
 		// All chunks have been written, mark the pipeline ready
 		p.readyLock.Lock()
--- a/internal/pkg/service/stream/storage/level/local/encoding/pipeline_test.go
+++ b/internal/pkg/service/stream/storage/level/local/encoding/pipeline_test.go
@@ -10,12 +10,14 @@ import (
 	"testing"
 	"time"
 
+	"github.com/c2h5oh/datasize"
 	"github.com/jonboulle/clockwork"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 
 	"github.com/keboola/keboola-as-code/internal/pkg/log"
 	commonDeps "github.com/keboola/keboola-as-code/internal/pkg/service/common/dependencies"
+	"github.com/keboola/keboola-as-code/internal/pkg/service/common/duration"
 	"github.com/keboola/keboola-as-code/internal/pkg/service/stream/dependencies"
 	"github.com/keboola/keboola-as-code/internal/pkg/service/stream/mapping/recordctx"
 	"github.com/keboola/keboola-as-code/internal/pkg/service/stream/storage/level/local/diskwriter/network/connection"
@@ -657,6 +659,60 @@ foo3
 `)
 }
 
+func TestEncodingPipeline_ChunkRetryGivesUpAndCloses(t *testing.T) {
+	t.Parallel()
+
+	tc := newEncodingTestCase(t)
+	tc.Slice.Encoding.Sync.Mode = writesync.ModeDisk
+	tc.Slice.Encoding.Sync.Wait = false
+	tc.Slice.Encoding.MaxChunkSize = 64 * datasize.KB // minimum allowed, keeps the test payload small
+	// Max allowed check interval, so advancing the clock by minutes below doesn't replay hundreds
+	// of thousands of periodic-sync ticks (each real-time-scheduled) through the fake clock.
+	tc.Slice.Encoding.Sync.CheckInterval = duration.From(30 * time.Second)
+
+	closed := make(chan string, 1)
+	tc.CloseFunc = func(ctx context.Context, cause string) {
+		closed <- cause
+	}
+
+	w, err := tc.OpenPipeline()
+	require.NoError(t, err)
+
+	// All writes fail permanently, e.g. the volume is unreachable.
+	tc.Output.WriteError = errors.New("some error")
+
+	// A record that exactly fills one chunk completes it immediately via the size-based split in
+	// chunk.Writer.Write, without needing an explicit Flush/Sync trigger (which has its own,
+	// real-time-based 30s timeout, unrelated to what this test is verifying).
+	body := strings.Repeat("x", 64*1024-1)
+	_, err = w.WriteRecord(tc.TestRecord(body))
+	require.NoError(t, err)
+	tc.ExpectWritesCount(t, 1)
+
+	// Wait for the first failed attempt, it starts the retry-duration clock.
+	assert.EventuallyWithT(t, func(c *assert.CollectT) {
+		tc.Logger.AssertJSONMessages(c, `{"level":"warn","message":"chunks write failed: some error, waiting %s, chunks count = 1"}`)
+	}, 5*time.Second, 100*time.Millisecond)
+
+	// Move well past the retry bound, the currently pending backoff wait fires,
+	// the next attempt fails again and the pipeline should give up instead of retrying forever.
+	tc.Clock.Advance(10 * time.Minute)
+
+	select {
+	case cause := <-closed:
+		assert.NotEmpty(t, cause)
+	case <-time.After(5 * time.Second):
+		t.Fatal("timeout waiting for closeFunc to be called after exhausting chunk retries")
+	}
+
+	// Deliberately not calling w.Close() here: the abandoned chunk is never drained by anyone
+	// once processChunks has given up on it, so Close (via the syncer's Stop -> forced sync)
+	// would block for up to its own real-time (non-fake-clock) 30s Flush timeout - twice, once
+	// for the pending auto-triggered sync woken by the Clock.Advance above, once for Stop's own
+	// forced one. That's an existing, pre-existing latency (Syncer.Stop ignores the caller's ctx
+	// for the underlying sync goroutine), not something this test needs to exercise.
+}
+
 // encodingTestCase is a helper to open encoding pipeline in tests.
 type encodingTestCase struct {
 	*writerSyncHelper
@@ -670,6 +726,7 @@ type encodingTestCase struct {
 	Events            *events.Events[encoding.Pipeline]
 	Manager           *encoding.Manager
 	Slice             *model.Slice
+	CloseFunc         func(ctx context.Context, cause string)
 }
 
 type writerSyncHelper struct {
@@ -707,6 +764,7 @@ func newEncodingTestCase(t *testing.T) *encodingTestCase {
 		Events:            events.New[encoding.Pipeline](),
 		Manager:           d.EncodingManager(),
 		Slice:             slice,
+		CloseFunc:         func(ctx context.Context, cause string) {},
 	}
 	return tc
 }
@@ -725,7 +783,7 @@ func (tc *encodingTestCase) OpenPipeline() (encoding.Pipeline, error) {
 		tc.Slice.Encoding,
 		tc.Slice.LocalStorage,
 		tc.Slice.Encoding.Compression.Type != compression.TypeNone,
-		func(ctx context.Context, cause string) {},
+		tc.CloseFunc,
 		tc.Output,
 	)
 	if err != nil {
--- a/internal/pkg/service/stream/storage/level/local/encoding/retry.go
+++ b/internal/pkg/service/stream/storage/level/local/encoding/retry.go
@@ -6,6 +6,12 @@ import (
 	"github.com/cenkalti/backoff/v5"
 )
 
+// maxChunkRetryDuration bounds how long processChunks keeps retrying writes to a broken
+// network output before giving up and closing the pipeline. Without a bound, a pipeline
+// stuck on a permanently unreachable volume retries forever, leaking its goroutine and
+// buffers instead of being recreated.
+const maxChunkRetryDuration = 5 * time.Minute
+
 func newChunkBackoff() *backoff.ExponentialBackOff {
 	b := backoff.NewExponentialBackOff()
 	b.RandomizationFactor = 0.1


Matovidlo:
@claude review

internal/pkg/service/stream/storage/level/local/encoding/chunk/writer.go
```diff
@@ -202,6 +202,23 @@ func (w *Writer) ProcessCompletedChunks(fn func(chunk *Chunk) error) error {
...
+	w.lock.Lock()
+	defer w.lock.Unlock()
+
+	if len(w.completedChunks) == 0 {
+		return
+	}
+
+	w.freeChunks(w.completedChunks...)
+	w.completedChunks = nil
+	close(w.allProcessedNotifier)
```
pepamartinec, line 218: :red_circle: **BLOCKER - discarded records can be reported to the client as written.**

`pipeline.Flush` parks on exactly this channel, and its only non-error branch is `return nil`:

```go
// pipeline.go:404-409
select {
case <-p.chunks.WaitAllProcessedCh():
    return nil
case <-ctx.Done():
    return errors.PrefixErrorf(...)
}
```

So a `Flush` already parked at the moment of give-up is woken by this `close()` and returns **success** for chunks that were just handed back to the pool. With `Sync.Mode = cache` the syncer's sync fn *is* `pipeline.Flush` (`writesync/sync.go:90`), and with `Sync.Wait = true` (the default) `WriteRecord` waits on that notifier - so the source acks records it discarded. `ModeDisk` only escapes because `network.Sync` fails afterwards; that is luck, not design.

I reproduced it with a throwaway test in this package (no etcd needed) - it passes against this branch today:

```go
w := chunk.NewWriter(log.NewNopLogger(), 1024)
_, err := w.Write(bytes.Repeat([]byte("x"), 3*1024))
require.NoError(t, err)
require.Equal(t, 3, w.CompletedChunks())

ch := w.WaitAllProcessedCh() // what a parked Flush holds
select {
case <-ch:
    t.Fatal("waiter should be blocked, 3 chunks are unprocessed")
default:
}

w.Abandon()

select {
case <-ch:
    // closed -> pipeline.Flush takes its "return nil" branch
default:
    t.Fatal("expected Abandon to unblock the parked waiter")
}
assert.Equal(t, 0, w.CompletedChunks()) // all 3 dropped, not just one
w.Abandon()                             // must not panic
```

Suggested fix: make give-up sticky - a `gaveUp` flag on the pipeline (or on the writer) that `Flush` checks *before* waiting, so it returns an error instead of relying on the notifier. That also fixes the ~60s `Close` noted on the test file.

Separately: `Abandon` is the only new exported method in this PR and has no direct test, even though `chunk/writer_test.go` exists. The snippet above is that test - folding it in covers both points.

pepamartinec (commented):
Review of the bounded-retry change. The core idea is right and the leak diagnosis is well evidenced - a pipeline that retries a broken write forever and never closes is exactly the shape of the goroutine/RSS growth described.

**Verified rather than assumed:**
- `TestEncodingPipeline_ChunkRetryGivesUpAndCloses` passes `-race -count=5` with no flake (~1.2s per run).
- Removing the give-up block makes it fail with `timeout waiting for closeFunc to be called after exhausting chunk retries`, so the test genuinely pins the new behaviour.
- The test's own comments check out: `minBytes=64kB` (`encoding/config/config.go:15`) and `maxDuration=30s` (`writesync/config.go:42`) are the real bounds, and `dummyEncoder` writes `body + \n` = 65536 B = exactly `MaxChunkSize`, so "exactly fills one chunk" is accurate.
- No goleak in this package, so the deliberate missing `w.Close()` trips nothing.
- No close/reopen thrash: on give-up `SlicePipeline.Close` runs `unregister`, so the slice leaves the collection and only comes back on an etcd slice change.

**2 blockers** (marked :red_circle: BLOCKER inline): discarded records can be reported to the client as successfully written, and `p.ready` can still be `true` at give-up. The first one I reproduced with a throwaway test in the `chunk` package; the second is a one-line fix.

The other 4 comments are non-blocking. On the backoff never being `Reset()` - Copilot already flagged that, and it pairs naturally with the `failingSince` comment below.

internal/pkg/service/stream/storage/level/local/encoding/pipeline.go
```diff
@@ -520,13 +521,29 @@ func (p *pipeline) processChunks(ctx context.Context, clk clockwork.Clock, encod
...
+			// Give up after writes have been failing continuously for too long, instead of
+			// retrying forever - a permanently unreachable volume would otherwise leave this
+			// goroutine, and the pipeline it belongs to, running until the process restarts.
+			if failingSince.IsZero() {
+				failingSince = clk.Now()
+			} else if clk.Since(failingSince) > maxChunkRetryDuration {
+				p.logger.Errorf(ctx, "chunks write failed for over %s, giving up: %s", maxChunkRetryDuration, err)
+				// Nobody else will ever process these chunks, don't leave a future Flush call
+				// (e.g. from the syncer, during Close) waiting forever for them.
+				p.chunks.Abandon()
```
pepamartinec, line 533: :red_circle: **BLOCKER - `p.ready` can still be `true` when we give up.** One-line fix.

`p.ready = false` is gated on `cnt >= encodingCfg.FailedChunksThreshold` (line 518, default `3`). Give up with 1 or 2 queued chunks and `p.ready` stays `true`.

`IsReady()` does not save us either - its only other check is `p.network.IsReady()`, which is a gRPC connectivity probe returning true for `Idle` **or** `Ready` (`rpc/client.go:205-210`). That is exactly the state where writes keep erroring while the connection still looks fine.

So between give-up and the async `closeFunc` finishing (~60s on this path, see the comment on the test file) the balancer keeps routing records into a pipeline nobody will ever drain.

```go
// before p.chunks.Abandon()
p.readyLock.Lock()
p.ready = false
p.readyLock.Unlock()
```

internal/pkg/service/stream/storage/level/local/encoding/pipeline.go
```diff
@@ -520,13 +521,29 @@ func (p *pipeline) processChunks(ctx context.Context, clk clockwork.Clock, encod
...
+			}
+
 			// Wait before retry
 			delay := b.NextBackOff()
 			p.logger.Warnf(ctx, "chunks write failed: %s, waiting %s, chunks count = %d", err, delay, cnt)
 			<-clk.After(delay)
 			continue
 		}
 
+		failingSince = time.Time{}
```
pepamartinec, line 545: **Non-blocking - the 5-minute window measures "time since the last fully clean round", not "time without progress".**

`ProcessCompletedChunks` returns an error as soon as *any* chunk fails, after having successfully written the ones before it (`processedIndex++` at `chunk/writer.go:197`, then early return on error). `failingSince` is only cleared here, on a completely error-free round.

So a link that is genuinely draining - writes 9 of 10 chunks, errors on the tail of every round - is killed at 5 minutes and its buffer discarded, even though it was making progress the whole time.

Cheap fix: capture `p.chunks.CompletedChunks()` before the call and clear `failingSince` when the count went down, not only when `err == nil`. This is also the natural place to `b.Reset()` (see Copilot's comment) so the retry delay and the give-up window stop disagreeing after a transient blip.

Worth confirming: was the `kbc-eu-central-1` volume fully dead, or degraded? Degraded is the more common shape, and it is the one this distinction decides.

internal/pkg/service/stream/storage/level/local/encoding/pipeline.go
```diff
@@ -520,13 +521,29 @@ func (p *pipeline) processChunks(ctx context.Context, clk clockwork.Clock, encod
 				p.readyLock.Unlock()
 			}
 
+			// Give up after writes have been failing continuously for too long, instead of
+			// retrying forever - a permanently unreachable volume would otherwise leave this
+			// goroutine, and the pipeline it belongs to, running until the process restarts.
+			if failingSince.IsZero() {
+				failingSince = clk.Now()
+			} else if clk.Since(failingSince) > maxChunkRetryDuration {
+				p.logger.Errorf(ctx, "chunks write failed for over %s, giving up: %s", maxChunkRetryDuration, err)
```
pepamartinec, line 530: **Non-blocking - the data loss is invisible.**

`Abandon()` drops buffered records that were already accepted (and, with `Sync.Wait = false`, already acked to the HTTP client), but this log line carries only the write error. There is no chunk count, no byte total, no metric.

For the incident this PR is chasing, the line an operator needs post-deploy is *"dropped N chunks / M bytes for slice X"*. The post-release plan watches `runtime.go.num_goroutine` and `container.memory.rss`; it should also be able to watch how much data this is throwing away. At minimum log the counts here; a counter would be better.

Related question: after give-up the slice pipeline closes and unregisters. Does anything downstream mark that slice as incomplete, or is it treated as finished with a hole in it?

internal/pkg/service/stream/storage/level/local/encoding/pipeline_test.go
```diff
@@ -657,6 +659,60 @@ foo3
...
+	tc.Clock.Advance(10 * time.Minute)
+
+	select {
+	case cause := <-closed:
+		assert.NotEmpty(t, cause)
+	case <-time.After(5 * time.Second):
+		t.Fatal("timeout waiting for closeFunc to be called after exhausting chunk retries")
+	}
+
+	// Deliberately not calling w.Close() here: the abandoned chunk is never drained by anyone
```
pepamartinec, line 708: **Non-blocking, but I would push back on "pre-existing".**

It was pre-existing while the drainer ran forever. This PR is what makes the drainer *exit*, so it now owns the hang. And `Abandon()` does not actually prevent it: a later `Flush` calls `chain.Flush` first, which completes the active chunk into `completedChunks` and installs a **fresh** `allProcessedNotifier` - with `processChunks` already returned, nothing drains it, so that `Flush` blocks its full 30s. `Abandon` only helps the one waiter parked at the instant of give-up (which is the blocker on `chunk/writer.go`).

Why this is more than a comment nit: `SinkPipeline.UpdateSlicePipelines` closes replaced pipelines **serially, inside `p.updateLock.Lock()`** (`pipeline_sink.go:109` and `148-156`). With N slices on a broken volume, the sink's slice-update path is blocked for ~N x 60s and no new slices open for that sink meanwhile. Note also that `pipeline.Close`'s `chunksWg.Wait()` is a plain `WaitGroup` wait - the 5-minute `closeCtx` that `SlicePipeline.Close` wraps around it does **not** interrupt it. Pre-PR that meant the lock was wedged forever, so this PR is still strictly better; this does not block.

The `gaveUp` flag suggested on `chunk/writer.go` fixes this and the blocker in one change. If you would rather not, please drop the "waiting forever" claim from `Abandon`'s doc comment and state plainly that `Close` takes ~60s on this path.

internal/pkg/service/stream/storage/level/local/encoding/retry.go
```diff
@@ -6,6 +6,12 @@ import (
 	"github.com/cenkalti/backoff/v5"
 )
 
+// maxChunkRetryDuration bounds how long processChunks keeps retrying writes to a broken
+// network output before giving up and closing the pipeline. Without a bound, a pipeline
+// stuck on a permanently unreachable volume retries forever, leaking its goroutine and
+// buffers instead of being recreated.
+const maxChunkRetryDuration = 5 * time.Minute
```
pepamartinec, line 13: **Non-blocking, minor.** Its sibling knob `FailedChunksThreshold` is config (`encoding/config/config.go:16`, `validate:"required,min=1,max=100"`). A limit that trades buffered data for bounded memory seems like it belongs next to it rather than as a hard-coded const - if 5 minutes turns out to be too aggressive on a slow stack, changing it currently needs a release.

internal/pkg/service/stream/storage/level/local/encoding/chunk/writer.go
Matovidlo: **Fixed** ✅ in commit 84799da71e1aee1d03dc527cecc81b8408468dd3

Added a sticky `pipeline.gaveUp` bool, set together with `ready=false` at give-up. `Flush` now checks it both up front (before touching `chain.Flush`/the notifier at all) and again right after being woken from `WaitAllProcessedCh` - so a `Flush` call already parked at the moment of give-up (your exact repro) gets woken by `Abandon`'s close, re-checks the flag, and returns an error instead of trusting the notifier alone. Also folded your repro snippet into a new direct `TestWriter_Abandon` in `chunk/writer_test.go` - there wasn't a unit test for `Abandon` before.

internal/pkg/service/stream/storage/level/local/encoding/pipeline.go
Matovidlo: **Fixed** ✅ in commit 84799da71e1aee1d03dc527cecc81b8408468dd3

`p.ready = false` (and `gaveUp = true`) is now set unconditionally at give-up, before `Abandon()`/the async `closeFunc`, instead of being gated on `FailedChunksThreshold`. The regression test now asserts `IsReady()` is `false` immediately after give-up.

internal/pkg/service/stream/storage/level/local/encoding/pipeline.go
Matovidlo: **Fixed** ✅ in commit 84799da71e1aee1d03dc527cecc81b8408468dd3

Capture the completed-chunk count before `ProcessCompletedChunks` and compare it to what's left after; if it went down (the round made progress), `failingSince` clears and `b.Reset()` runs, same as a fully clean round. A link draining 9/10 chunks per round now survives instead of getting killed at 5 minutes. This also covers Copilot's backoff-reset comment below.

On the volume question: checked with the reporter - the disk volume itself was up/healthy throughout. The repeated write failures weren't caused by a dead volume; the actual trigger was something else in the path to the disk-writer node that hasn't been pinned to one precise mechanism yet. I'll correct the PR description's "unreachable disk-writer volume" phrasing so it doesn't overclaim - the leak and this fix don't depend on which failure mode caused the writes to keep failing.

internal/pkg/service/stream/storage/level/local/encoding/pipeline.go
Matovidlo: **Fixed** ✅ in commit 84799da71e1aee1d03dc527cecc81b8408468dd3

The give-up log line now includes the abandoned chunk count: `"chunks write failed for over %s, giving up, abandoning %d chunk(s): %s"`. Didn't add a metric on top of it - flagging that as a possible follow-up rather than doing it here.

On the downstream question: traced it - no, nothing marks the slice incomplete today. `slicerotation/operator.go`'s `closeSlice` only waits for source nodes to stop *using* the slice (`closeSyncer.WaitForRevision`); it doesn't check what was actually written, and there's no `ExpectedRecordsCount`/checksum anywhere under `storage/` to reconcile against. A slice this PR gives up on proceeds through Closing -> Uploading -> Uploaded -> Imported exactly like a complete one. That's a real gap, but reconciling record counts/checksums across the storage pipeline is a separate, much larger effort than this PR - proposing it as a follow-up ticket rather than blocking here.

internal/pkg/service/stream/storage/level/local/encoding/pipeline_test.go
Matovidlo: **Fixed** ✅ in commit 84799da71e1aee1d03dc527cecc81b8408468dd3

Fair pushback, and the `gaveUp` fix above resolves this too: `Flush` now fails fast (checks `gaveUp` before doing any work) instead of blocking on its own 30s real-time timeout, so `Close` no longer hangs ~60s post-give-up, and `SinkPipeline.UpdateSlicePipelines` isn't serialized behind it either. Replaced the old comment explaining why the test didn't call `Close()` with an actual assertion: the test now calls `Close()` and asserts it returns an error promptly (bounded by a 5s real-time safety timeout on the test side, not a production delay).

internal/pkg/service/stream/storage/level/local/encoding/retry.go
Matovidlo: **Fixed** ✅ in commit 84799da71e1aee1d03dc527cecc81b8408468dd3

`maxChunkRetryDuration` is now `encoding/config.Config.MaxChunkRetryDuration` (`duration.Duration`, default 5m, `validate:"required,minDuration=1s,maxDuration=1h"`), threaded through the same way `FailedChunksThreshold` already was. The package-level const in retry.go is gone.

internal/pkg/service/stream/storage/level/local/encoding/pipeline.go
```diff
@@ -520,13 +521,29 @@ func (p *pipeline) processChunks(ctx context.Context, clk clockwork.Clock, encod
...
 			p.logger.Warnf(ctx, "chunks write failed: %s, waiting %s, chunks count = %d", err, delay, cnt)
 			<-clk.After(delay)
 			continue
 		}
 
+		failingSince = time.Time{}
+
 		// All chunks have been written, mark the pipeline ready
 		p.readyLock.Lock()
 		p.ready = true
```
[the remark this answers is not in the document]
Matovidlo: **Fixed** ✅ in commit 84799da71e1aee1d03dc527cecc81b8408468dd3

Fixed as part of the progress-tracking change (see reply to Pepa's comment on the give-up window, just above) - `b.Reset()` now runs on a fully clean round and on any round that made partial progress, not just implicitly never.

internal/pkg/service/stream/storage/level/local/encoding/config/config.go
```diff
@@ -14,6 +17,7 @@ type Config struct {
 	InputBuffer           datasize.ByteSize  `json:"inputBuffer" configKey:"inputBuffer" configUsage:"Max size of the buffer before compression, if compression is enabled. 0 = disabled" validate:"maxBytes=16MB"`
 	MaxChunkSize          datasize.ByteSize  `json:"maxChunkSize" configKey:"maxChunkSize" configUsage:"Max size of a chunk sent over the network to a disk writer node." validate:"required,minBytes=64kB,maxBytes=1MB"`
 	FailedChunksThreshold int                `json:"failedChunksThreshold" configKey:"failedChunksThreshold" configUsage:"If the defined number of chunks cannot be sent, the pipeline is marked as not ready." validate:"required,min=1,max=100"`
+	MaxChunkRetryDuration duration.Duration  `json:"maxChunkRetryDuration" configKey:"maxChunkRetryDuration" configUsage:"If chunks cannot be sent for longer than this, the pipeline gives up and closes." validate:"required,minDuration=1s,maxDuration=1h"`
```
pepamartinec, line 20: :red_circle: **BLOCKER (new) - slices already in etcd decode this as `0`, which means "give up after ~100ms".**

Asking for this knob was my round-1 comment, so this one is on me. The hard-coded const had no zero-value hazard; a stored config field does.

`encoding.Config` is not only runtime config - it is stored verbatim in etcd inside every file and slice record (`model.File.Encoding`, `model.Slice.Encoding`). This PR's own `initial-etcd-kvs.txt` fixtures show it.

Chain, traced end to end:

1. Records written by the version deployed today have no `maxChunkRetryDuration` key.
2. `Serde.Decode` does call `validate`, but that closure skips structs and pointers (`common/dependencies/etcdclient.go:33-38`), so the missing `required` field is not caught on read.
3. The field decodes to `0`.
4. New slices inherit it: `s.Encoding = file.Encoding` (`repository/slice/slice_new.go:37`).
5. `router.go:125` copies it into `SliceData.Encoding`, which is what reaches `processChunks`.
6. `clk.Since(failingSince) > encodingCfg.MaxChunkRetryDuration.Duration()` with `0` is true on the **second consecutive failure** - roughly 100ms, the first backoff interval - instead of after 5 minutes.

Replicated: unmarshalling a stored pre-PR config JSON into `encoding.Config` gives `MaxChunkRetryDuration == 0` while `FailedChunksThreshold` correctly decodes as `3`.

Why this blocks rather than being a wrinkle: the zero value of this field is the **most aggressive setting it can express**, and it costs buffered data. Every file open at deploy time keeps that config until it rotates - default import interval is 1 minute, but `target.import.trigger.interval` is configurable up to 24h, so a sink with a long interval keeps it far longer. During that window a single transient blip on the write path kills the pipeline and discards its buffer, on the very stack this PR exists to stabilise.

Note CI structurally cannot catch this: the E2E `initial-etcd-kvs.txt` fixtures were updated to *include* the new key, so the suite only ever seeds the new shape. The 13 green checks say nothing here.

Smallest fix - floor the value at the point of use instead of trusting the decoded record:

```go
// retry.go
const defaultChunkRetryDuration = 5 * time.Minute

// processChunks, before the loop
maxRetry := encodingCfg.MaxChunkRetryDuration.Duration()
if maxRetry <= 0 {
    maxRetry = defaultChunkRetryDuration
}
```

internal/pkg/service/stream/storage/level/local/encoding/pipeline.go
```diff
@@ -520,13 +541,46 @@ func (p *pipeline) processChunks(ctx context.Context, clk clockwork.Clock, encod
 				p.readyLock.Unlock()
 			}
 
+			// A round that wrote at least one chunk before failing on the rest is progress,
+			// not a stall - treat it the same as a fully clean round below, instead of letting
+			// a link that is slowly draining get killed by the give-up window.
+			switch {
+			case cnt < countBefore:
```
pepamartinec: **Non-blocking - this progress check is defeated by concurrent chunk arrival, which is the case it exists for.**

`ProcessCompletedChunks` runs its callback with the writer lock **released** (`chunk/writer.go:187-189`), so the encoder keeps completing new chunks during the round. The deferred cleanup trims the processed prefix but keeps the new arrivals, so real progress can still leave `cnt >= countBefore`.

Reproduced in the `chunk` package - 3 chunks queued, 2 written successfully, 5 completed concurrently during the round:

```
countBefore=3 cnt=6, progress detected = false
```

That is exactly the high-throughput shape this check is meant to protect: with `Sync.Wait = false` records keep flowing during an outage, so new chunks arrive on every round and `cnt` rarely drops below `countBefore`.

Fix without touching the `chunk.Writer` API - count successes in the callback `processChunks` already owns:

```go
written := 0
err := p.chunks.ProcessCompletedChunks(func(chunk *chunk.Chunk) error {
    // ... existing body, unchanged ...
    written++      // only reached after a successful network.Write
    return nil
})
...
case written > 0:
    failingSince = time.Time{}
    b.Reset()
```

Still strictly better than the previous never-reset behaviour, so this does not block.

internal/pkg/service/stream/storage/level/local/encoding/pipeline.go
```diff
@@ -404,12 +413,22 @@ func (p *pipeline) Flush(ctx context.Context) error {
 	// We have chunks, what next?
 	select {
 	case <-p.chunks.WaitAllProcessedCh():
+		// The notifier is also closed by Abandon, when chunks are discarded rather than written.
+		if p.hasGivenUp() {
```
pepamartinec, line 418: **Non-blocking - this is the guard that fixes the original blocker, and it is the one guard nothing drives.**

Deleting just this check leaves every test green (I ran it). The early `hasGivenUp()` at line 401 covers all *future* `Flush` calls, so the test is satisfied either way.

But this check is for the actual round-1 race: a `Flush` that already entered, passed line 401, completed its chunk and parked on `WaitAllProcessedCh` - and then give-up fires and `Abandon` closes that notifier under it. `TestWriter_Abandon` documents the hazard in a comment ("its caller must check for itself that chunks were discarded rather than written, see pipeline.gaveUp") but nothing exercises it.

Both guards are load-bearing in different situations, so neither should go. Deterministically parking a `Flush` at the instant of give-up is real work - your call whether it is worth a test. Flagging it so it stays a decision rather than becoming an oversight later.

internal/pkg/service/stream/storage/level/local/encoding/pipeline.go
```diff
@@ -520,13 +541,46 @@ func (p *pipeline) processChunks(ctx context.Context, clk clockwork.Clock, encod
...
+			case cnt < countBefore:
+				failingSince = time.Time{}
+				b.Reset()
+			case failingSince.IsZero():
+				failingSince = clk.Now()
+			case clk.Since(failingSince) > encodingCfg.MaxChunkRetryDuration.Duration():
+				// Give up after writes have been failing continuously for too long, instead of
+				// retrying forever - a permanently broken network output would otherwise leave
+				// this goroutine, and the pipeline it belongs to, running until process restart.
+				p.logger.Errorf(ctx, "chunks write failed for over %s, giving up, abandoning %d chunk(s): %s", encodingCfg.MaxChunkRetryDuration, cnt, err)
```
pepamartinec: **Nits, none blocking.**

- The chunk count here is a real improvement over the previous round, but it is still chunks only - no byte total and no metric. Round-1 item 4 is half closed: good enough for triage from logs, not enough to graph how much data a give-up threw away.
- `pipeline.Close` now returns an error on this path, so `SlicePipeline.Close` logs `cannot close slice pipeline: ...` at Error level on every give-up and skips its `closed slice pipeline` info line. That is correct - it *is* a failure - just flagging the new error-log volume so it does not surprise anyone on the dashboards.
- `gaveUp` shares `readyLock` (`pipeline.go:101`). It works and avoids a second lock, but the lock name no longer describes everything it guards; `stateLock` would read better.
- `minDuration=1s` lets an operator configure near-instant give-up with data loss attached. Consistent with `FailedChunksThreshold` having `min=1`, so probably fine as an expert knob - just noting it is now reachable through sink settings.

pepamartinec (commented):
Review round 2, on `45290d67d`. All 6 points from the previous round are addressed, and I checked each fix rather than taking it on trust. **One new blocker**, which my own round-1 request for a config knob is what introduced - see the comment on `encoding/config/config.go`.

**Round-1 status**

| # | Finding | Fix | Verified |
|---|---|---|---|
| 1 | BLOCKER: discarded records reported as written | `pipeline.gaveUp`, checked in `Flush` before waiting **and** after the notifier fires | yes, ablation below |
| 2 | BLOCKER: `p.ready` still true at give-up | `p.ready = false` set unconditionally alongside `gaveUp` | yes, ablation below |
| 3 | progress does not reset the window | `case cnt < countBefore` + `b.Reset()` | partly, see the comment on `pipeline.go:548` |
| 4 | data loss invisible | log now reports `abandoning %d chunk(s)` | partly - count only, no bytes, no metric |
| 5 | `Close` ~60s under the sink update lock | `Flush` fails fast once `gaveUp` | yes |
| 6 | hard-coded const | `MaxChunkRetryDuration` config field | yes, and it introduced the new blocker |

Also picked up Copilot's `b.Reset()` on the success path. Thanks for the `TestWriter_Abandon` unit test.

**What I ran**
- `go test -race -count=3` over `encoding/...`, `storage/level/local/config` and `stream/config`: all ok.
- CI on `45290d67d`: 13 checks green.
- Ablations against `TestEncodingPipeline_ChunkRetryGivesUpAndCloses`:
  - drop the post-wait `hasGivenUp()` only -> still passes
  - drop the early `hasGivenUp()` only -> still passes
  - drop **both** -> FAILS, `An error is expected but got nil` - `Close` silently succeeds for a discarded chunk, which is exactly the round-1 blocker. The test genuinely pins it.
  - drop `p.ready = false` -> FAILS on `assert.False(t, w.IsReady())`. `dummyOutput.IsReady()` always returns true, so that assertion really tests `p.ready` and not the network probe.

The two non-blocking comments below are on the *fix* for round-1 item 3 and on a coverage gap, not on anything that was previously wrong.

internal/pkg/service/stream/storage/level/local/encoding/config/config.go
pepamartinec: @Matovidlo Claude tohle podal celkem dramatciky, ale mozna to mas nejak pojisteny?

Matovidlo: Asi spravne to odchytil necham ho kuknut do kodu. Vpodstate sa jedna o to, ze ked mas vsade zrazu stare zaznamy tak to serializujes do `0` a tym padom to hned "failne".

Matovidlo: **Fixed** ✅ in commit 30bf6cd6ec62a3d1af07a016b903c4c598a244fa

Fixed. `processChunks` now floors the decoded value to a 5m default at the point of use whenever it's <= 0 (`retry.go: chunkRetryDuration`), instead of trusting the stored config field.

Also traced this further, since it changes the picture a bit: it's not only old records. `configpatch.ApplyKVs` - used both for the settings-patch bind and for the lazy Config merge on `File.Open` - never calls `Validator.Validate`, and the etcd serde's validate closure only fires for non-struct/non-pointer values (so it's skipped for every persisted domain struct, `File`/`Slice` included). That means a client-submitted explicit `0` would be persisted unchecked too, not just a pre-existing record missing the field - this is a separate, pre-existing gap in the etcd client's validate wiring (predates this PR per git blame), not something introduced here. Flooring at the use site protects against both causes regardless, but the validate closure itself probably deserves its own issue since it silently disables validation for every persisted struct, not just this field - happy to file that separately if useful.

Added `TestChunkRetryDuration` (unit test for the floor) and `TestEncodingPipeline_ChunkRetryZeroConfigFloorsToDefault` (pipeline-level regression: a decoded zero doesn't give up in seconds, still gives up at the floored default).

internal/pkg/service/stream/storage/level/local/encoding/pipeline.go
Matovidlo: **Fixed** ✅ in commit 30bf6cd6ec62a3d1af07a016b903c4c598a244fa

Fixed. Replaced the before/after `CompletedChunks()` comparison with a `written` counter incremented in the `ProcessCompletedChunks` callback only after a successful `network.Write`, and keyed `failingSince`/backoff reset off `written > 0`. Concurrent chunk arrival during the round no longer masks a stalled round, since it only counts chunks this round actually wrote, not queue depth before/after.

internal/pkg/service/stream/storage/level/local/encoding/pipeline.go
Matovidlo: **Intentional** ℹ️

Agreed both guards are load-bearing, leaving them as-is. Not adding a dedicated test for deterministically parking a `Flush` at the give-up instant right now - noting it here as a conscious call rather than an oversight, per your framing. Happy to revisit if it turns into a real footgun.

internal/pkg/service/stream/storage/level/local/encoding/pipeline.go
Matovidlo: **Fixed** ✅ in commit 60063e6914c6c9641e3788c3db03b804c88c9622

Fixed the two actionable nits: give-up log now includes the abandoned byte total alongside the chunk count (30bf6cd6e, via new `chunk.Writer.CompletedBytes()`), and `readyLock` renamed to `stateLock` since it also guards `gaveUp` (60063e69). Left the error-log-volume note and the `minDuration=1s` reachability as informational, no action taken there.

pepamartinec (approved):

PR #2655, opened 2026-09-08
feat(apps-proxy): AJDA-3219 track end-user sessions in data apps

sykora-ji:
[AJDA-3219](https://linear.app/keboola/issue/AJDA-3219/track-user-sessions-in-data-apps)

## Description

Adds end-user session tracking to apps-proxy. A signed cookie identifies the
session; append-only events are emitted to a Keboola Stream HTTP source, which
lands them in a Storage table.

**New package `dataapps/sessions`**, built on the same pattern as
`dataapps/notify` (SyncMap + per-key throttle + WaitGroup on shutdown):

- `sessions.go` — manager, middleware, deadline computation, event building
- `cookie.go` — `_kbc_dasid`, `<uuidv7>.<deadline>.<hmac>`, per-app signing key
  derived from app id + `cookieSecretSalt`
- `store.go` — in-memory state with time-based eviction
- `writer.go` — Stream sender, bounded queue, worker pool
- `event.go` — the event, whose fields are the target table's columns

**Three integration points**, all in code that already existed:

| Where | What |
|---|---|
| `apphandler/manager.go` | middleware wraps the upstream, between auth and the app |
| `upstream/upstream.go` | HTTP and per-frame websocket activity, session end on websocket close |
| `apphandler/apphandler.go` | session end on `/_proxy/sign_out`, which never reaches the upstream |

`authproxy/manager.go` stamps the admitting provider into the request context —
oauth2-proxy does not pass that on and one app can offer several providers.

**Event model rather than one row per session**, because Stream is append-only:
a session cannot be one mutable row. Sessions are reconstructed downstream by
grouping on `sessionId`. `requests` / `ws_frames` are deltas, not running
totals, so they sum correctly across replicas.

Three event types, and only one of them ends anything: `session_start` when the
cookie is minted, `heartbeat` for any flush (throttle interval, websocket close,
graceful shutdown), and `session_end` **only** for a sign-out, which is terminal
— at most one per session id, because signing out clears the cookie. Every other
way a visit stops emits nothing, so reporting closes those sessions with
`idle_timeout_seconds`, which is stamped on every row because the setting is
per-stack.

**Cookie validity slides** — every request buys another full `idleTimeout`
(30 min). A websocket handshake is the exception and gets a deadline covering
the whole connection (`upstream.wsTimeout` + 10 min grace), because after the
upgrade there is no response left to carry a `Set-Cookie` and Streamlit holds
one connection for hours; a 30 minute cookie would expire mid-connection and
split one visit into several sessions. Deadlines never move backwards and are
capped at `maxSessionLength` (12 h) from the session start. The deadline is
signed inside the cookie rather than left to `Max-Age`, which a client may
ignore.

**`frameworkpoll` extracted into its own package** so "the user did something"
has one definition. Streamlit's `/_stcore/health` and `/_stcore/host-config`
fire on every websocket reconnect independently of the user; they were already
excluded from the auto-suspend notification and are now excluded from session
activity too. It cannot live in the `upstream` package, which depends on
`sessions` in the other direction.

**Identity** is `provider_user_id`: the OIDC subject claim for a provider that
issues an ID token, and the account login for GitHub, which issues none. Either
way it names a person without describing them — no e-mail and no display name is
recorded. A shared-password or unauthenticated app yields a session id and
nothing else.

This does touch the authentication path, in two ways worth calling out:

- the OIDC provider now asks the IdP for the `sub` claim (`AdditionalClaims`),
  and oauth2-proxy injects it as a **new** `X-Kbc-User-Id` header on every
  authenticated request that reaches a data app — alongside the existing
  `X-Kbc-User-Name` / `-Email` / `-Roles`, which stay because data apps already
  read them, `keboola_streamlit` among them;
- for GitHub that header carries the account login instead, taken from the
  session's user field.

The two differ in how long they hold. A subject claim is stable for the life of
the account; a GitHub login is not — its owner can change it, and a released one
can be taken over by another account, so a renamed account counts as a new user
from then on.

`scripts/stream-sessions-setup.sh` provisions the Stream source and table sink
and writes the ingest URL to a mode-600 state file. It is idempotent and
resumable, and refuses to run against a sink whose column mapping no longer
matches the event — Stream would otherwise write empty values into the changed
columns without reporting anything. Verified end to end against a mock Stream
API and against the canary stack, including that refusal and the migration
procedure in the docs. `docs/apps-proxy/sessions.md` documents the data model, the
reconstruction query, cookie sizing, multi-replica behaviour and the delivery
limitations.

## Release Notes

- **Justification**
    - Nothing in the platform could answer who used a data app, when, or for
      how long. `lastRequestTimestamp` is a single overwritten scalar used for
      auto-suspend, and `AppRun` records pod lifetime for billing — neither
      says anything about people. The only existing option was an opt-in
      `keboola_streamlit.create_event()` call that an app author had to write
      themselves, which does not exist at all for PythonJS or custom images.
    - Data apps now record how many people open them and for how long, into a
      Storage table that can be queried like any other data. *Who* they were is
      recorded only where the app's own authentication provider says so — an app
      behind a shared password, or with no authentication at all, contributes
      sessions but no identity.
    - No customer communication needed. The feature is off unless a stack
      configures a Stream URL, and the data lands in a Keboola-internal
      project, not in customer projects. Should this later write into customer
      projects, that would need its own communication and a privacy review —
      the identity recorded belongs to an app's end users, who may be external
      to Keboola.
- **Impact Analysis**
    - With no `sessions.streamUrl` configured the manager is disabled: the
      middleware returns the next handler unchanged, no goroutine and no
      writer are started, and every exported method is a no-op. The only
      residue is one extra handler indirection and a context value from
      `WithAuthProvider`, applied unconditionally.
    - When enabled, the middleware sits in the request path of every data app
      on the stack. It adds a cookie read, an HMAC verification and, on most
      requests, one `Set-Cookie` header. Events are sent fire-and-forget from a
      bounded queue, so a slow or unreachable Stream drops events rather than
      adding request latency.
    - Single-tenant stacks are unaffected: Stream is deployed on 12 of the 20
      stacks that run apps-proxy, and the eight without it are all
      single-tenant clouds. They keep `streamUrl` empty.
    - The chart change that passes the env var lives in kbc-stacks and is
      gated the same way, so no stack renders the variable unless it sets a
      value.
- **Deployment Plan**
    - Continuous deployment. Enabling is a per-stack configuration change, not
      a code change, and is separate from this PR: canary first, to confirm
      events reach the table end to end, then stack by stack.
    - Expect identity-less rows for a while after a stack is enabled. The
      subject claim is written into the oauth2-proxy session cookie at login or
      refresh, so anyone already signed in keeps a cookie without it and
      `provider_user_id` stays empty until they authenticate again. It
      self-heals, and it would be easy to misread as a bug in the new code.
      `auth_provider_type IN ('oidc','gitlab','jumpcloud') AND provider_user_id = ''`
      is the anomaly worth watching; the same blank for `'password'` or for an
      empty provider type is by design.
- **Rollback Plan**
    - Revert and redeploy, or simply clear `sessions.streamUrl` to switch the
      feature off without a deployment. Not a one-way door — nothing outside
      the target Storage table holds state, and no existing data or schema is
      touched.
    - No Support involvement needed while the data stays in an internal
      project. Both ways this can fail are silent from the outside — a full
      queue drops events, and a failed send is not retried — so three metrics
      exist to make them visible, and they are what to watch on the first
      enabled stack:
      `keboola.go.appsproxy.sessions.events.sent` (split by `is_success` and
      `error_type`), `...events.dropped` (anything above zero means sessions
      are being undercounted) and `...sessions.tracked` (bounded by eviction,
      so a line that only climbs is a leak). The matching log lines carry the
      reason but are rate-limited to the first occurrence and every thousandth,
      so they are for diagnosis after a metric has already fired.


MiroCillik (commented):
Thorough review of the session tracking work. The design is well documented and `docs/apps-proxy/sessions.md` §6 already acknowledges most of the obvious failure modes, so I've excluded anything the docs call out.

One finding worth acting on before merge (the lost `sign_out` signal on multi-replica deployments); the rest are low severity.

Verified as correct and not worth re-litigating: middleware ordering (oauth2-proxy → `WithAuthProvider` → sessions → upstream) puts `X-Kbc-User-*` and the provider id in scope before `begin` runs; `Sessions.Validate()` *is* wired in (`configmap.doVisitStruct` calls `OnValue` per nested struct); `onCloseConn` is reachable because `httputil.ReverseProxy.handleUpgradeResponse` closes `backConn` via `defer close(backConnCloseCh)`, and `res.Request.Context()` does carry the session since the outbound request is `req.Clone(ctx)`; cookie signing rejects the interesting forgeries (re-stated deadline, wrong app, missing deadline segment — a UUID contains no `.`, so `strings.Cut` catches it); lock ordering is consistently store→entry with no inversion; `sanitize` does strip the secret-bearing URL out of `*url.Error`; and the `Event` JSON tags match the sink column mapping in `scripts/stream-sessions-setup.sh:247-262` exactly.

internal/pkg/service/appsproxy/dataapps/sessions/sessions.go
```diff
@@ -0,0 +1,481 @@
...
+		return
+	}
+	s, ok := FromContext(ctx)
+	if !ok {
+		return
+	}
+
+	// No entry means the session was already ended (or never seen on this
+	// replica): nothing to report, and this makes a repeated End a no-op.
+	item, found := m.store.take(s.ID)
```
MiroCillik: **medium — `sign_out` `session_end` is silently lost on multi-replica deployments.**

`End` returns early when `store.take` finds no cached entry, and `EndRequest` (sign-out) goes through it. The proxy runs 2 replicas with no session affinity, so a `/_proxy/sign_out` request frequently lands on a replica that never cached that session id: the cookie is cleared (`clearCookie` already ran on line 469) but no `session_end` row is ever written. The session then closes via the 30 min idle fallback instead.

That contradicts the docs, which promise `end_reason = 'sign_out'` is the one *final* signal — in practice it goes missing for roughly half of sign-outs.

The `!found` early return only needs to protect idempotence of the repeated ws-close path. The sign-out path could build the event with zero deltas when there is no cached entry.

internal/pkg/service/appsproxy/dataapps/sessions/sessions.go
```diff
@@ -0,0 +1,481 @@
...
+	// The mint time is encoded in the UUIDv7, so it survives a proxy restart,
+	// is identical across replicas, and cannot be moved by the client.
+	state, found := readCookie(req, key, now, m.cfg.MaxSessionLength)
+	if !found {
+		sessionID, err := newSessionID()
+		if err != nil {
+			m.logger.Errorf(ctx, "cannot generate session id: %s", err.Error())
+			return nil
+		}
+		state = cookieState{sessionID: sessionID, startedAt: now}
```
MiroCillik: **low — `session_start` differs between the first row and all later rows of the same session.**

On a new session `startedAt` is `m.clock.Now()`, but on every subsequent request it is decoded from the UUIDv7 (`sessionStartFromID`, ms-truncated, and minted a moment *after* `now`). So `buildEvent` (line 404) writes a different `sessionStart` on the `session_start` row than on all later rows.

This breaks the documented schema invariant at `docs/apps-proxy/sessions.md:108` — "Decoded from `session_id`, identical on every row of a session". Harmless for the `MIN(session_start)` query in the docs, but it splits any downstream `GROUP BY session_id, session_start` into two groups.

Deriving `startedAt` from the freshly minted id (rather than from `now`) fixes it.

internal/pkg/service/appsproxy/dataapps/sessions/sessions.go
```diff
@@ -0,0 +1,481 @@
...
+		// session_start whenever a session moved between replicas. A missing
+		// cookie, by contrast, means the id was generated a moment ago and no
+		// replica can have seen it.
+		m.emit(ctx, s, item, EventSessionStart, "", now)
+	case s.userEmail != "" || s.userName != "":
+		// A session can start on a public path and authenticate later. Flush a
+		// heartbeat as soon as identity appears instead of waiting out the
+		// heartbeat interval.
+		item.lock.Lock()
+		pending := !item.identitySent
```
MiroCillik: **low — check-then-act on `identitySent` can emit a duplicate identity heartbeat.**

The flag is read here under `item.lock`, the lock is released, and only then does `emit` re-acquire it and set the flag inside `buildEvent`.

Two requests arriving concurrently right after login (e.g. the page load plus its first XHR) can both observe `pending == true`, so each emits an identity heartbeat — producing a duplicate row, the second with `requests`/`wsFrames` zeroed.

Doing the test-and-set in one critical section removes the window.

internal/pkg/service/appsproxy/dataapps/sessions/sessions.go
```diff
@@ -0,0 +1,481 @@
...
+			}
+		}
+	}()
+
+	d.Process().OnShutdown(func(ctx context.Context) {
+		m.logger.Info(ctx, "waiting for pending session events")
+		m.writer.close(ctx)
+	})
+
+	// The websocket timeout is not settable through configuration, so this can
```
MiroCillik: **low — this comment isn't accurate, and the invariant is only a warning.**

`Upstream.WsTimeout` carries `configKey:"wsTimeout"` (`config/config.go:92`, with no `validate` tag at all), and `Sessions.MaxSessionLength` is settable too. Both sides of the inequality are operator-controlled, yet the mismatch only produces a `Warnf`.

An operator setting `APPS_PROXY_UPSTREAM_WS_TIMEOUT=24h` (or `MAX_SESSION_LENGTH=1h`) starts the proxy normally, and every long websocket visit is then split into several session ids — inflating session counts in the reported data.

This cross-struct invariant belongs in `Config.Validate()` alongside the `Sessions.Validate()` checks; failing that, the comment should be corrected.

internal/pkg/service/appsproxy/proxy/apphandler/apphandler.go
```diff
@@ -37,6 +38,10 @@ type appHandler struct {
 
 type ruleIndex int
 
+// signOutPath is the oauth2-proxy sign-out endpoint, mounted under the proxy
+// prefix configured in oauthproxy.proxyConfig.
+const signOutPath = config.InternalPrefix + "/sign_out"
```
MiroCillik: **low — duplicates the already-exported `selector.SignOutPath`.**

This re-derives a value that `authproxy/selector/selector.go:28` already exports with an identical value. If the selector's path ever changes, this copy silently stops matching — no compile error and no test failure, and `session_end` emission plus cookie clearing on sign-out just quietly stop working.

Suggest referencing `selector.SignOutPath` instead.

internal/pkg/service/appsproxy/config/config_test.go
```diff
@@ -104,3 +104,141 @@ func TestConfig_RequiresStorageAPIURL(t *testing.T) {
...
+// TestSessionsConfig_Envs pins the ENV names down: they are what the kbc-stacks
+// chart sets, and they are documented in docs/apps-proxy/sessions.md.
+func TestSessionsConfig_Envs(t *testing.T) {
+	t.Parallel()
+
+	cfg := validConfig(t)
+
+	envs := env.Empty()
+	envs.Set("APPS_PROXY_SESSIONS_STREAM_URL", "https://stream-in.keboola.local/stream/123/sessions/secret")
+	envs.Set("APPS_PROXY_SESSIONS_MAX_SESSION_LENGTH", "6h")
```
MiroCillik: **low — this ENV is set but never asserted, so the test can't fail.**

`TestSessionsConfig_Envs` exists to "pin the ENV names down", sets `APPS_PROXY_SESSIONS_MAX_SESSION_LENGTH=6h`, and then never asserts `cfg.Sessions.MaxSessionLength`. Since `validConfig(t)` starts from `config.New()` (12h) and 12h satisfies validation, a wrong or renamed config key for the one field the docs advertise at `docs/apps-proxy/sessions.md:151` would leave the test green.

```go
assert.Equal(t, 6*time.Hour, cfg.Sessions.MaxSessionLength)
```

internal/pkg/service/appsproxy/dataapps/sessions/writer.go
```diff
@@ -0,0 +1,192 @@
...
+	queueSize   int
+	workers     int
+	sendTimeout time.Duration
+}
+
+func newWriter(logger log.Logger, cfg writerConfig) *writer {
+	w := &writer{
+		logger: logger,
+		url:    cfg.url,
+		client: &http.Client{Timeout: cfg.sendTimeout},
```
MiroCillik: **low — no `Transport`, so idle connections are capped at 2 for 4 workers.**

An `http.Client` with no `Transport` uses the process-wide `http.DefaultTransport`, whose `MaxIdleConnsPerHost` is 2, while `Workers` defaults to 4. Under sustained load half the sends must open a fresh TLS connection to Stream, and that handshake is charged against the 5s `sendTimeout` — making event drops more likely exactly when volume is highest.

A dedicated `Transport` with `MaxIdleConnsPerHost >= Workers` would also stop this path sharing connection state with the rest of the proxy.

sykora-ji:
Thanks — this was a genuinely useful pass. Six of the seven are fixed in `bd7904e98`; one had a premise I could not reproduce.

**`sign_out` lost on multi-replica** — correct, and it contradicted the docs exactly as you say. The early return in `End` is what makes a repeated websocket close idempotent, so I kept it there and gave the sign-out path its own emit with zero deltas when the replica holds no entry. A repeated sign-out stays quiet because `clearCookie` already ran, so the second request carries no cookie at all. Covered by `TestManager_SignOutIsRecordedOnAReplicaThatNeverSawTheSession`, which uses two managers over one stream to stand in for the two replicas.

**`session_start` differing between the first row and the rest** — correct, and it is visible in the canary data: single-event sessions carry microseconds (`10:58:46.885683`), multi-event ones milliseconds (`10:38:36.757`). Both paths now decode the id. `TestManager_SessionStartIsIdenticalOnEveryRow` pins it.

**check-then-act on `identitySent`** — correct, the test and the set now happen in one critical section.

**duplicated `signOutPath`** — correct, uses `selector.SignOutPath`.

**ENV set but never asserted** — correct, and my own doing when renaming `cookieTtl` to `maxSessionLength`. The assertion is back. Amusingly the new cross-struct check below rejected the `6h` that test was using, which is the enforcement demonstrating itself.

**`http.DefaultTransport` idle cap** — correct. The writer now clones the default transport (keeping proxy support and dial timeouts) and raises `MaxIdleConnsPerHost` to the worker count.

---

**On the invariant being "operator-controlled on both sides"** — I could not reproduce that half. `Upstream` carries `configKey:"-"` on the `Config` field, and the parent's exclusion also detaches its children from ENV binding. I probed it directly:

```
APPS_PROXY_UPSTREAM_WS_TIMEOUT=24h  -> WsTimeout=6h0m0s
APPS_PROXY_WS_TIMEOUT=24h           -> WsTimeout=6h0m0s
```

So `wsTimeout` can only move in code, and the comment was accurate as written. Your second point stands though: `maxSessionLength` *is* settable, and it is the side that gets set by mistake — a warning was too weak for it. It is now enforced in `Config.Validate()` (only when tracking is enabled, so stacks without Stream are unaffected), the warning is gone, and `TestConfig_ValidateWebsocketFitsInSession` also proves the root `Validate` hook is reached at all.

Docs updated for the enforced invariant. Race tests across `appsproxy` and `golangci-lint` with the CI config are both clean.

internal/pkg/service/appsproxy/dataapps/sessions/sessions.go
MiroCillik: Thanks — and you're right, my premise here was wrong. I checked `configmap`: `mapAndFilterField` returns `ok=false` for `configKey:"-"`, and `doVisit` does `continue` on that (visit.go:223), so the whole `Upstream` subtree is skipped and `WsTimeout`'s own tag never binds. Only `maxSessionLength` is operator-settable, exactly as your comment says.

One thing the new check does lose, though: it's strictly weaker than the warning it replaced, in the band the grace exists for.

`deadline()` grants a websocket handshake `m.wsTimeout + websocketGrace` (6h10m by default), and that's then capped at `MaxSessionLength`. The old warning compared against `wsTimeout + websocketGrace`; `Config.Validate` now compares against `wsTimeout` alone. So `MAX_SESSION_LENGTH=6h5m` passes validation, yet a handshake asking for a 6h10m cookie window gets capped to 6h5m — the reconnect after the drop finds no valid cookie and mints a new session, which is the splitting the check exists to prevent.

`c.Sessions.MaxSessionLength <= c.Upstream.WsTimeout + websocketGrace` would close it (needs `websocketGrace` visible to the config package, or the constant moved).

internal/pkg/service/appsproxy/dataapps/sessions/sessions.go
MiroCillik: Confirmed fixed. `EndRequest` now takes from the store and falls back to `&entry{}`, and the early `readCookie` return keeps a cookie-less request (a bot hammering `/_proxy/sign_out`) from emitting anything. `buildEvent` only reads plain counters off `item`, so the zero entry is safe — the row just carries zero deltas, which is the right thing for a session this replica never saw.

internal/pkg/service/appsproxy/dataapps/sessions/sessions.go
MiroCillik: Confirmed fixed — deriving `startedAt` from the minted id makes the `session_start` row agree with every later row. The `err != nil` fallback to `now` reintroduces the mismatch in principle, but only if `sessionStartFromID` fails on an id `newSessionID` just produced, so that is fine.

internal/pkg/service/appsproxy/dataapps/sessions/sessions.go
MiroCillik: Confirmed fixed — the test and the set are in one critical section now, and hoisting the event out to enqueue after the unlock keeps the lock off the writer. Dropping the `emit` helper in favour of the explicit lock/build/unlock at both call sites reads well.

internal/pkg/service/appsproxy/proxy/apphandler/apphandler.go
MiroCillik: Confirmed fixed — now referencing `selector.SignOutPath`, so the constant cannot drift out of sync.

internal/pkg/service/appsproxy/config/config_test.go
MiroCillik: Confirmed fixed — the assert is in, and changing the value to `8h` keeps it distinct from the `12h` default so a broken config key actually fails the test.

internal/pkg/service/appsproxy/dataapps/sessions/writer.go
MiroCillik: Confirmed fixed — and cloning `http.DefaultTransport` rather than building one from scratch is the right call, it keeps the proxy support and dial timeouts.

MiroCillik:
Note on the red `E2E: CLI` jobs — I looked into them, they are not caused by this PR.

All three platforms fail only on `TestCliE2E/remote-table-upload/table-exist-same-csv-columns-diff-order`, and the assertion is purely about column ordering in the resulting project state:

```
@@ -45,0 +46,3 @@   (- expected, + actual)
-            "Region", "Name", "Id"
+            "Id", "Name", "Region"
```

The table's reported column order after uploading a CSV whose header is in a different order no longer matches what the fixture pins. That is Storage-side behaviour in `remote table upload` — this branch touches only `appsproxy/`, `docs/` and `scripts/`, and the fixture has not changed since 3e522d8c1 (2026-06-09).

It is not flake either: it reproduced identically on all three platforms across both runs (34218555057 and 34224647154). The reason no other PR shows it is that no other branch has run `E2E: CLI` since 2026-09-02, and main's last CI run was 2026-08-13 — so there is simply no recent baseline. A run of `E2E: CLI` on main would confirm it fails there too.

So it needs a separate fix (fixture updated for the new column-order behaviour, or a Storage regression raised) rather than anything on this branch.

MiroCillik (approved):
Approving. All seven findings from my earlier pass are fixed, and I verified each in the code rather than taking the summary on trust — including the two-manager sign-out test and the `Config.Validate` hook actually being reached.

One thing, not blocking.

**The `websocketGrace` band is still open** (`config/config.go:155`). The check reads `MaxSessionLength <= c.Upstream.WsTimeout`, but `deadline()` grants a websocket handshake `wsTimeout + websocketGrace`, so the validation is narrower than the invariant its own comment describes:

```go
if c.Sessions.StreamURL != "" && c.Sessions.MaxSessionLength <= c.Upstream.WsTimeout + websocketGrace {
```

I want to be clear that this is minor, having initially framed it as bigger than it is. The unprotected band is only `(6h, 6h10m]`, an operator would have to choose something like `6h5m` rather than a round number, and even then the cookie is still valid at the reconnect that follows the drop — the split needs the reconnect to be delayed more than 5 minutes on top of that. The consequence is analytics-only: one visit counted as two sessions. Worth folding into a follow-up rather than holding the PR.

internal/pkg/service/appsproxy/dataapps/sessions/sessions.go
```diff
@@ -0,0 +1,534 @@
...
+				return
+			case <-ticker.Chan():
+				if n := m.store.evictBefore(m.clock.Now().Add(-m.cfg.IdleTimeout)); n > 0 {
+					m.logger.Debugf(ctx, "evicted %d idle sessions, %d remaining", n, m.store.len())
+				}
+			}
+		}
+	}()
+
+	d.Process().OnShutdown(func(ctx context.Context) {
```
pepamartinec, line 116: **Nothing flushes the store on shutdown.** `writer.close` drains the outgoing *queue*, but every live session's `requests`/`wsFrames` accumulated since its last heartbeat were never queued — so every deploy discards up to `heartbeatInterval` (5 min) of counters per session.

The outer ordering is already right, which is worth saying: `OnShutdown` callbacks run LIFO (`servicectx.go:102`) and the HTTP server registers later in `StartServer`, so the server drains its requests before this runs. So the fix fits inside this same callback — flush, then `writer.close(ctx)`.

The blocker is that `store.entry` carries no identity (only `lock, lastSeen, nextHeartbeatAfter, requests, wsFrames, identitySent`), so no event can be built from an entry alone. **The same missing piece blocks three separate things** — this flush, the websocket-close flush, and any sweep-time event. One `*Session` snapshot on the entry plus one "drain this entry as an event" helper covers all three; that is the structural fix and the rest falls out of it.

One detail if you do it: stamp `event_time` from `item.lastSeen`, not `clock.Now()`. `buildEvent(s, item, typ, reason, now)` already takes the time as a parameter, so it is a one-argument change — and using `now` would date the activity at shutdown time, inflating every session's duration by up to the heartbeat interval.

To be explicit about what *not* to add: **please don't emit `session_end` here.** A restart does not end the visit — the cookie survives and Streamlit reconnects — so that would split one visit into N sessions per N deploys.

internal/pkg/service/appsproxy/dataapps/sessions/sessions.go
```diff
@@ -0,0 +1,534 @@
...
+	// with every visitor for the lifetime of the process.
+	ticker := m.clock.NewTicker(sweepInterval)
+	go func() {
+		defer ticker.Stop()
+		for {
+			select {
+			case <-ctx.Done():
+				return
+			case <-ticker.Chan():
+				if n := m.store.evictBefore(m.clock.Now().Add(-m.cfg.IdleTimeout)); n > 0 {
```
pepamartinec: **The eviction cutoff is sized for HTTP but also governs websocket sessions.** This evicts at `now - IdleTimeout` (30 min), while a websocket-issued cookie is valid `wsTimeout + websocketGrace` ≈ 6 h 10 min. And `lastSeen` only moves on a *data* frame — I checked `wsactivity/frame.go:165`, ping/pong are control frames and deliberately excluded.

So a Streamlit tab left open while the user is at lunch loses its entry at 30 minutes **while the connection is still live**. Then `End` → `store.take` finds nothing → returns silently: no `session_end`, and the pending frame counts are gone. The docs attribute a lost `session_end` to restarts and hard disconnects, but not to this.

It does self-heal — the next data frame recreates the entry and the same session id continues. But it creates a wrinkle worth documenting either way: because the cookie stays valid for hours, one `session_id` can span a 35-minute silence, and `GROUP BY session_id` with `MIN`/`MAX(event_time)` will report that as a single continuous session including the dead gap.

`max(IdleTimeout, wsTimeout + websocketGrace)` would fix it. At ~200 bytes per entry, 6 h retention is cheap. Worth noting this becomes load-bearing if a sweep ever emits events: a live-but-quiet connection would get a *spurious* end.

internal/pkg/service/appsproxy/dataapps/sessions/writer.go
```diff
@@ -0,0 +1,202 @@
...
+type writer struct {
+	logger    log.Logger
+	url       string
+	client    *http.Client
+	queue     chan Event
+	done      chan struct{}
+	closeOnce sync.Once
+	wg        sync.WaitGroup
+	dropped   atomic.Uint64
+	failed    atomic.Uint64
```
pepamartinec, line 42: **No metrics anywhere in the package** — `git grep -E 'Meter|Counter|Gauge|Histogram'` over `dataapps/sessions/` returns nothing.

`dropped` and `failed` only ever reach a log line (the 1st and every 1000th), and `store.len()` appears in one Debug log, and only when eviction actually removed something. So today you cannot alert when events start dropping, graph the store size to spot a leak, or see the send error rate.

That matters more here than usual because **the only failure mode is silent**: the queue drops on overflow and sends are fire-and-forget with no retries (correctly so — a retried heartbeat would double-count its delta). A log line cannot carry an alert.

CLAUDE.md asks for "metrics for critical paths". Three would do it: events sent by status, events dropped, store size. The Post-Release Support Plan currently says to watch the logs on the first enabled stack — metrics are what would make that possible.

internal/pkg/service/appsproxy/dataapps/sessions/event.go
```diff
@@ -0,0 +1,93 @@
...
+	// session continues — so one session id can carry several of these. Only a
+	// sign-out is final, and it also clears the cookie.
+	EventSessionEnd EventType = "session_end"
+)
+
+// EndReason explains why the session ended. Only set on EventSessionEnd.
+type EndReason string
+
+const (
+	EndReasonWebsocketClose EndReason = "ws_close"
```
pepamartinec: **`session_end{ws_close}` does not end the session, and the name is doing real damage.** Agreed with @Pepa-M that this should go.

Your own docstring makes the case better than I can: *"One session id can therefore carry several session_end rows; downstream takes the last event, not the first end."* When a type name needs three sentences of prose to explain that it does not mean what it says, the name is the problem.

And the payload says a `ws_close` row *is* a heartbeat. Same `buildEvent` call, same 9 constant dimensions, same delta drain, same throttle re-arm — two strings differ. "The websocket closed" also changes no downstream decision, because the user may reconnect in seconds and the idle window decides either way.

The schema is the contract, and this table is sold as queryable like any other data. Anyone writing ad-hoc SQL — BI, support, an LLM — will write `WHERE event_type = 'session_end'` to count finished sessions, over-count sessions and under-report durations. The correct semantics live only in a doc file they will not open.

**Agreed shape:**

| type | meaning |
|---|---|
| `session_start` | cookie minted |
| `heartbeat` | a flush — throttle, websocket close, shutdown |
| `session_end` | **sign-out only.** Terminal. |

That buys a real invariant: **at most one `session_end` per session id**, because sign-out clears the cookie so a second one carries none and `EndRequest` returns early. `end_reason` can then go — the type is the reason.

We also considered a `timeout` reason emitted from the sweep and **decided against it**: a marker present on only some sessions is worse than none. Reporting needs the idle-window fallback regardless, so sessions with the marker would be dated at `lastSeen` and those without at `MAX(event_time) + idle_window`, up to 30 minutes apart — and you cannot normalise, because "no marker" is indistinguishable between "still open" and "end was lost" (a replica that restarts with no further request for that session emits nothing, ever). One uniform rule beats a partly-populated one.

Cheap now while `streamUrl` is empty on every stack; a table migration once a stack is enabled.

internal/pkg/service/appsproxy/dataapps/sessions/store.go
```diff
@@ -0,0 +1,111 @@
...
+//
+// It is a cache, never a source of truth. The proxy runs several replicas and
+// restarts on deploy, so the same session can be tracked by more than one
+// entry over its lifetime, or by none. Everything that must survive lives in
+// the cookie (session id, and the start time encoded in it) or in the emitted
+// events. Consequences that downstream queries must tolerate:
+//
+//   - heartbeat deltas of one session may arrive from two replicas,
+//   - session_end may never arrive at all.
+type entry struct {
```
pepamartinec, line 11: **Comment density.** Measured across the five new non-test files: **301 comment lines vs 697 code lines, 43%** — `event.go` is at 179%, more comment than code. `sessions.go` is 534 lines, over the 500-line guidance in CLAUDE.md.

The sharper problem is not volume but duplication: `docs/apps-proxy/sessions.md` is 326 lines covering the same reasoning, so there are two copies of the same argument and they will drift the first time one changes. This docblock is an example — the two bullets are §6 almost word for word.

House style is "default to no comment; rationale, tradeoffs and cross-system mechanism go in the PR body or docs, never in the code". Candidates to move out, most already in the docs: the `writer` struct's retry rationale, `newWriter`'s `MaxIdleConnsPerHost` reasoning, this docblock, `store.take`'s reconnect-cycle explanation, `startsSession`'s three paragraphs, `frameworkpoll.Is`, and the `End`/`EndRequest` docblocks. That is ~70 lines out of `sessions.go`, which also puts it back under 500.

**The ones that must stay** are the ones a reader would otherwise break — shortened to a line or two each: the `case !found:` note that `session_start` is gated on the cookie and never on the cache (by far the most important comment in the package; someone *will* try to "optimise" it), `buildEvent`'s lock requirement, the identity-flush critical section, `enqueue`'s never-close-the-queue, `timeFormat`'s fixed precision, `keyPrefix`'s collision with the oauth2-proxy key space, `SameSite=Lax`, and `deadline()`'s never-move-backwards.

internal/pkg/service/appsproxy/config/config.go
```diff
@@ -31,10 +31,45 @@ type Config struct {
...
+	HeartbeatInterval time.Duration `configKey:"heartbeatInterval" configUsage:"Minimum interval between heartbeat events of one session." validate:"required,minDuration=10s"`
+	// IdleTimeout is how long a session survives without activity. It sizes the
+	// session cookie deadline on an ordinary request, evicts the proxy's
+	// in-memory state, and is the window downstream uses to close sessions
+	// whose session_end event never arrived.
+	//
+	// A websocket handshake is the exception: that cookie is issued to cover
+	// the connection's whole lifetime, because after the upgrade there is no
+	// response left to carry a Set-Cookie.
+	IdleTimeout time.Duration `configKey:"idleTimeout" configUsage:"Inactivity after which a session is considered ended." validate:"required,minDuration=1m"`
```
pepamartinec, line 67: **`idleTimeout` is load-bearing for reporting but is not recorded in the data.**

With no timeout event (see the `event.go` thread), every non-sign-out session's end is computed downstream as `MAX(event_time) + idleTimeout`. But this is per-stack settable — `APPS_PROXY_SESSIONS_IDLE_TIMEOUT`, and `config_test.go` even uses `15m` as its example — and nothing in the event rows records which value was in force.

So a stack that changes it silently invalidates every historical duration, and a query spanning stacks cannot tell them apart.

Either pin it as a documented non-varying constant, or stamp it into the events. I'd stamp it — the table already carries 9 constant dimensions per row, so one more is consistent with its design, and it makes the reporting query self-contained.

internal/pkg/service/appsproxy/proxy/apphandler/authproxy/oauthproxy/config.go
```diff
@@ -70,6 +70,9 @@ func proxyConfig(
 	v.Session = options.SessionOptions{Type: options.CookieSessionStoreType}
 	v.EmailDomains = []string{"*"}
 	v.InjectRequestHeaders = []options.Header{
+		// Kept: data apps already read these, keboola_streamlit among them.
```
pepamartinec: Nit: `// Kept:` sits above the `X-Kbc-User-Id` line, but that header is **new in this PR** — `git log -S 'X-Kbc-User-Id' -- <this file>` returns only `65dfad8df`. The comment describes the `X-Kbc-User-Name` / `-Email` / `-Roles` lines below it. Move it under the new line, or reword.

pepamartinec (commented):
Reviewed at `4c86b0db3` (re-verified after the last push — the GitHub-login change and the sink-mapping fix landed mid-review). Findings discussed with @Pepa-M; the design calls below are agreed with him, not just my opinion.

**This is strong work.** The cookie carries everything durable, `session_start` is gated on the cookie rather than the cache, and `docs/apps-proxy/sessions.md` is better than most design docs get. Most of what follows is about the in-memory store and the event vocabulary, not the core design.

### Claims I replicated rather than trusted

| Claim | Verified how | Result |
|---|---|---|
| `Set-Cookie` survives the 101 upgrade (sessions.md) | read `handleUpgradeResponse` in the Go source | correct — `copyHeader(rw.Header(), res.Header)` then `res.Header = rw.Header()` |
| `X-Kbc-User-Id` cannot be spoofed | `apphandler.go:152-158` | correct — all inbound `x-kbc-*` deleted before auth and before the middleware |
| Cookie is per-app | `AppConfig.CookieDomain` = `<app>.<publicURL host>` | correct, and per-app signing key on top |
| Eviction goroutine outlives startup | `entrypoint.runOrErr` ctx is cancelled only after `run` returns; `run` blocks on `WaitForShutdown()` | correct — the sweeper does not die early |
| `onCloseConn` always fires | `handleUpgradeResponse` closes `backConn` on req-ctx done or handler return; `sync.Once` guards | correct |
| `Config.Validate` is reached | `configmap.ValidateAndNormalize` visits the root value | correct, and `TestConfig_ValidateWebsocketFitsInSession` proves it |
| `userIDClaim` is right for every provider | GitLab embeds `*OIDCProvider` → `buildSessionFromClaims` → `AdditionalClaims["sub"]` populated; JumpCloud keeps `OIDCProvider` type; GitHub `getUser` sets `s.User = user.Login` | correct for all four |
| Sink mapping matches the `Event` JSON tags | compared field by field | correct |

### Already fixed while I was reading — nice

- `check_mapping` in the setup script is a better fix than I was going to ask for: it covers both the resume path and the 409-reuse path, and refuses rather than half-migrating.
- GitHub identity via the account login, plus the stability caveat in the docs and §5.2's migration procedure. That closes what I had down as the biggest gap.

### Build / test / lint at `4c86b0db3`

- `go build ./internal/pkg/service/appsproxy/...` — clean
- `go test -race ./internal/pkg/service/appsproxy/...` — **all 17 packages green**
- `golangci-lint run ./internal/pkg/service/appsproxy/...` — 4 issues, all in files this PR does not touch. No new lint debt.
- Tests are genuinely good: `manager_test.go` drives a real `httptest` Stream server through the real writer, and covers page-load bursts, sliding expiry, the websocket deadline floor, the absolute cap, sign-out on a replica with no state, and forged cookies. One gap: nothing asserts session events end to end through the real upstream, so the `GotConn → Activity` and `onClose → End` wires are covered by reading code only.

### Not inline-able

**The PR description is out of date, and it is the release-notes record.** It still says *"Identity is the email and name from the `X-Kbc-User-*` headers … both nullable"* — two renames ago. It also says *"No change to any authentication path"*, but `oauthproxy/config.go` now requests an extra claim and injects a **new** `X-Kbc-User-Id` header into every authenticated request reaching a customer data app. And *"Data apps now record who opened them"* needs the shared-password / no-auth carve-out spelled out.

### Considered and deliberately deferred (agreed with @Pepa-M)

- **Dev-mode previews are tracked untagged.** kai-preview reaches the tracked upstream via `serveKaiPreview`, but never passes `authproxy.NewHandlers`, so it gets no `WithAuthProvider` stamp — `authProviderId` comes out empty, identical to a public no-auth app. Raised, considered, left for later on purpose. If it is ever wanted, stamping a provider id like `kai-preview` on that path is the cheap fix, and it is cheap only while `streamUrl` is empty everywhere.

### Suggested order

Nothing here blocks the merge. I would hold **enabling it on a stack** until the store flush and the metrics land, because until then the two things that go wrong are both invisible.

internal/pkg/service/appsproxy/dataapps/auth/provider/oidc.go
```diff
@@ -33,6 +33,11 @@ func (v OIDC) ProxyProviderOptions() (proxyOptions.Provider, error) {
 				Default: []string{"select_account"},
 			},
 		},
+		// The subject claim identifies the user without naming them. It is
+		// stable across e-mail and display-name changes, which the email claim
+		// used as the user id is not. Only reachable when the provider issues
+		// an ID token, so GitHub — which does not — yields nothing.
+		AdditionalClaims: []string{"sub"},
```
pepamartinec, line 40: Worth calling out in the rollout plan rather than changing: **identity will be blank for everyone already signed in when this deploys.**

`AdditionalClaims` is populated in `ProviderData.buildSessionFromClaims` at login or refresh and stored inside the oauth2-proxy session cookie. Existing cookies carry no `sub`, so `X-Kbc-User-Id` is empty until each user re-authenticates. Self-healing, but expect a stretch of identity-less rows on the first enabled stack — and it would be easy to misread as a bug in the new code.

GitHub is unaffected, since `s.User` has always been populated for authz.

This is also a good argument for the diagnostic query being in the docs: `auth_provider_type IN ('oidc','gitlab','jumpcloud') AND provider_user_id = ''` is an anomaly worth watching, whereas the same blank for `'github'` (pre-login-change), `'password'`, or an empty provider type is by design.

internal/pkg/service/appsproxy/dataapps/sessions/cookie.go
```diff
@@ -0,0 +1,206 @@
...
+}
+
+// signingKey derives a per-app HMAC key. Per-app keys mean a cookie minted for
+// one app cannot be replayed as a valid session of another.
+//
+// The leading keyPrefix keeps this out of the oauth2-proxy cookie secret's key
+// space, which is SHA256(appID + "/" + providerID + "/" + salt): without it, an
+// app configuring a provider with the id "sessions" would derive the very same
+// key for both.
+func signingKey(appID api.AppID, salt string) []byte {
```
pepamartinec, line 63: Minor, defence in depth: an empty `salt` is not rejected here, whereas `oauthproxy.generateCookieSecret` errors on it.

With an empty salt the per-app key is just `SHA256("data-app-session/" + appID + "/")`, which anyone who knows the app id can compute and use to forge a session cookie. `CookieSecretSalt` is `validate:"required"`, so this needs an already-broken setup to reach — but this function is the one place that would silently accept it.

internal/pkg/service/appsproxy/dataapps/sessions/sessions.go
sykora-ji: Fixed in `30c44731b`.

The entry now carries a `*Session` snapshot, refreshed on **every** request rather than just the first — a session that starts anonymous and authenticates later would otherwise flush without its identity. That snapshot was the shared blocker you identified, and it is what any sweep-time event would need too; I have not added one.

Both of your conditions are in: the flush emits a `heartbeat`, never `session_end`, and it is stamped at `item.lastSeen` rather than `clock.Now()`.

Two things I checked rather than took on trust. The LIFO ordering holds — `servicectx.go:103` iterates callbacks in reverse — so the flush runs after the server has drained. And the test is not decorative: with `m.flushAll(ctx)` commented out it fails on `timed out waiting for a session event`, and it also asserts the row is a heartbeat with an empty end reason and dated at the activity, not at the shutdown (the clock advances 4 minutes in between).

internal/pkg/service/appsproxy/dataapps/sessions/sessions.go
sykora-ji: Fixed in `e90297958`, as `max(IdleTimeout, wsTimeout + websocketGrace)` behind a `retention()` helper.

I verified every step of the mechanism before accepting it: `wsTimeout` defaults to 6 h and `websocketGrace` is 10 min, and `wsactivity/frame.go:165` gates the callback on `opcode&controlBit == 0`, so ping and pong genuinely never move `lastSeen`.

Worth adding to your point: the comment on `End` listed exactly two reasons for a missing entry — already accounted for, or never seen on this replica — so it was actively asserting that this third case could not happen. Reworded.

The test is internal and deterministic rather than racing the sweeper goroutine: nothing in the repo synchronises on the fake clock, and a test that advanced time and hoped would have passed with the bug in place.

internal/pkg/service/appsproxy/dataapps/sessions/writer.go
sykora-ji: Added in `90b1ee3b8`, the three you asked for:

```
keboola.go.appsproxy.sessions.events.sent      by is_success, and error_type on failure
keboola.go.appsproxy.sessions.events.dropped   queue full
keboola.go.appsproxy.sessions.tracked          entries on this replica
```

Shape and naming follow `common/task`. Outcome is one counter split by attribute rather than two, so the failure rate is a ratio of the same series; `error_type` is there because a timeout and a 4xx mean different things — Stream being slow versus the mapping or the url being wrong. The gauge is observed on collection, since the store is a gauge by nature.

Two tests assert the instruments actually record rather than merely exist, and the writer's own tests now run with real instruments on a no-op meter instead of a `nil` the code would have to guard against.

The Post-Release Support Plan in the description names these now instead of pointing at log lines.

internal/pkg/service/appsproxy/dataapps/sessions/event.go
sykora-ji: Agreed, and done in `7a9da1b9b` in exactly the shape you and @Pepa-M settled on.

```
session_start  cookie minted
heartbeat      a flush — throttle, websocket close, graceful shutdown
session_end    the user signed out. terminal.
```

`End` is now `WebsocketClosed` and `EndRequest` is `SignOut`, so the call sites say which of the two they are instead of passing a reason, and `EndReason` is gone entirely. A test pins the invariant.

I checked your payload claim before accepting the argument, and it was exact: all four call sites go through the same `buildEvent`, which drains the same deltas and re-arms the same throttle, so a `ws_close` row really did differ from a heartbeat in two strings.

One consequence to flag: the canary sink still has the old schema, and my own `check_mapping` now refuses to run against it until the sink and table are recreated — which is the right order, but it means canary needs a `.5` before it collects anything again.

internal/pkg/service/appsproxy/dataapps/sessions/store.go
sykora-ji: Done in `fa260e757`, following your keep-list as written. 376/788 → 280/795, so 47% → 35%.

You were right that duplication was the real problem rather than volume, so the rationale moved into the doc and the comments kept only what a reader would otherwise break — shortened, not deleted. The `case !found:` note is still there and still says why it is gated on the cookie and never on the cache.

`sessions.go` was still 557 lines after the trim, so I split it instead of shaving further: `emit.go` takes recording activity and turning it into events, `sessions.go` keeps recognising a session and maintaining its cookie and cache. 360 and 210 lines, and the seam is a real one.

`event.go` stays high by ratio at 145% and I left it deliberately — it is a 16-field struct whose fields are the target table's columns, so a line per column is the point of the file rather than an excess. Say the word if you disagree.

Also fixed in passing: I had orphaned the `deadline` docblock onto `retention` when inserting that helper above it.

internal/pkg/service/appsproxy/config/config.go
sykora-ji: Stamped, in `7a9da1b9b` — your reasoning, and it follows directly from dropping the `ws_close` end, since nothing reports the ordinary end of a visit any more.

It is `idleTimeoutSeconds` on every row, named with the unit rather than `idle_timeout` so nobody has to guess. The reconstruction query in the docs is now:

```
ended_at = COALESCE(signed_out_at, last_activity_at + idle_timeout_seconds)
```

The dashboard reads it from the rows as well, and shows which windows are actually present in the data rather than assuming one.

internal/pkg/service/appsproxy/proxy/apphandler/authproxy/oauthproxy/config.go
sykora-ji: Right, and it was my comment. Moved in `480d219ec` — the new header gets its own line and `// Kept:` now sits above the three it actually describes.

internal/pkg/service/appsproxy/dataapps/auth/provider/oidc.go
sykora-ji: Good catch — I had not considered it, and you are right that it would read as a bug in the new code.

It is in the description's Deployment Plan now, with your diagnostic condition: `auth_provider_type IN ('oidc','gitlab','jumpcloud') AND provider_user_id = ''` is the anomaly worth watching, while the same blank for `'password'` or an empty provider type is by design.

I confirmed the mechanism rather than taking it as read: `AdditionalClaims` is populated only inside `buildSessionFromClaims`, called from `createSession(ctx, token, refresh)` and `CreateSessionFromToken`, and it lives in the session cookie as msgpack `ac`. So an existing cookie carries no `sub` and stays that way until the next login or refresh.

internal/pkg/service/appsproxy/dataapps/sessions/cookie.go
sykora-ji: Done in `480d219ec`, though not quite where you pointed.

The check is in `NewManager` rather than in `signingKey`: that is where the decision to run belongs, and it keeps an error return off the request path. With an empty salt tracking switches itself off and logs an error, on the grounds that tracking which trusts cookies it cannot verify is worse than no tracking.

Testing it needed its own dependency stub — the mocked service scope substitutes a salt when one is missing (`dependencies/mocked.go:83`), so it cannot express this case at all. Verified the test fails without the guard.

sykora-ji:
Thanks @pepamartinec — that was a genuinely useful review. All nine are addressed at `fa260e757`; replies are on the individual threads. Verifying each one first was worth it: every claim held, which is not what I expected going in.

| | | |
|---|---|---|
| store flush on shutdown | `30c44731b` | + the `*Session` snapshot you identified as the shared blocker |
| eviction cutoff vs the websocket window | `e90297958` | |
| `session_end` means the session ended | `7a9da1b9b` | `end_reason` gone, `idle_timeout_seconds` added |
| metrics | `90b1ee3b8` | |
| empty cookie salt, and the `// Kept:` nit | `480d219ec` | |
| comments, and `sessions.go` over 500 lines | `fa260e757` | split into `sessions.go` + `emit.go` |

**The description is fixed too.** You were right that it was the worse problem, since it is the release-notes record: it still described identity as the e-mail and name, and still claimed no authentication path changed while we now inject a new `X-Kbc-User-Id` header into every authenticated request reaching a customer app. Both gone, and it now carries the event vocabulary, your re-authentication caveat with its diagnostic query, and the metric names in the support plan.

**Two things I did differently from the suggestion**, both flagged on the threads: the salt check sits in `NewManager` rather than `signingKey`, to keep an error return off the request path; and `event.go` stays at 145% comments because it is a struct whose fields are the table's columns.

**Your suggested order is now unblocked.** The two invisible failure modes were the store flush and the missing metrics, and both have landed, so holding the per-stack enable on them no longer applies. One new thing does block canary specifically: the schema change means the existing sink and table have to be recreated, and my own `check_mapping` refuses to run against the old mapping until they are — so canary needs a `.5` and a fresh sink before it collects anything again. The kbc-stacks chart PR (keboola/kbc-stacks#24070) carries no value for any stack, so nothing is enabled by merging it.

Dev-mode previews being tracked untagged is left as agreed, and still cheap while `streamUrl` is empty everywhere.

internal/pkg/service/appsproxy/dataapps/sessions/emit.go
```diff
@@ -0,0 +1,210 @@
...
+	if !m.enabled {
+		return
+	}
+	s, ok := FromContext(ctx)
+	if !ok {
+		return
+	}
+
+	now := m.clock.Now()
+	item, _ := m.store.getOrInit(s.ID, now)
```
pepamartinec: New, minor, and a one-line fix — but it fails silently, which is the reason to bother.

`activity()` calls `getOrInit` but never sets `item.session`, and `getOrInit` creates entries as `&entry{lastSeen: now}` with a nil snapshot (`store.go:62`). `begin()` is normally what fills it in, and normally runs first — but an entry can be removed and then recreated by `activity()` alone, and `drain()` then skips it:

```go
if item.session == nil || (item.requests == 0 && item.wsFrames == 0) {
    return Event{}, false
}
```

I reproduced the path with a throwaway internal test rather than reasoning about it:

```go
item, _ := st.getOrInit("sess-1", now)
item.session = &Session{ID: "sess-1"}   // begin() ran

_, found := st.take("sess-1")           // WebsocketClosed on connection A
require.True(t, found)

item2, created := st.getOrInit("sess-1", now)  // connection B's next frame
require.True(t, created)
item2.wsFrames += 7                     // real user activity

assert.Nil(t, item2.session)
_, ok := (&Manager{}).drain(item2)
assert.False(t, ok)                     // 7 frames silently discarded
```

Passes. The realistic trigger is two overlapping websockets on one session id — a Streamlit reconnect where the new socket is up before the old one's close fires — and then a deploy before the next ordinary HTTP request re-runs `begin()`. Narrow, and it only loses frame counts, never a session.

`activity()` already has `s` in hand, so `item.session = s` alongside the delta increment closes the whole class.

Separately, `drain()` folds two different conditions into one silent `false`: "no snapshot" is a bug, "no deltas" is normal. Now that the metrics exist, splitting them — and counting the first — would mean this could never hide again.

pepamartinec (commented):
Re-reviewed at `80746d6`. **I replicated every claim rather than taking the replies at face value** — all seven commits do what they say.

| Claim | How I checked | Result |
|---|---|---|
| Flush runs before `writer.close`, LIFO holds | read the `OnShutdown` body — `flushAll(ctx)` then `writer.close(ctx)`; `servicectx.go` iterates callbacks in reverse | correct |
| Flush emits `heartbeat`, never an end | `drain` → `buildEvent(..., EventHeartbeat, ...)`; `SignOut` is the only `EventSessionEnd` call site in the package | correct |
| Stamped at `lastSeen`, not `now` | `m.buildEvent(item.session, item, EventHeartbeat, item.lastSeen)` | correct |
| `retention()` = `max(IdleTimeout, wsTimeout+grace)` | read the helper; sweeper now uses `-m.retention()` | correct |
| Three metrics, named as described | `metrics.go` — `…sessions.events.sent` / `.dropped` / `.tracked`, gauge observed on collection | correct |
| `EndReason` gone entirely | no occurrence left in `event.go`; `buildEvent` no longer takes a reason | correct |
| `idleTimeoutSeconds` on every row | `emit.go:130`, plus the `idle_timeout_seconds` sink column and the `COALESCE(...)` query in the docs | correct |
| Density 47% → 35% | re-measured with my own script: **280 comment / 795 code = 35%**, matching your number exactly | correct |
| `sessions.go` split | 360 lines + `emit.go` 210 — both well under 500 | correct |
| Salt guard | `NewManager` disables tracking and logs; `TestNewManager_EmptySaltDisablesTracking` fails without it | correct |

`go build` clean · `go test -race ./internal/pkg/service/appsproxy/...` **all 17 packages green** · `golangci-lint` 4 issues, all pre-existing in files this PR does not touch.

**On `event.go` staying at 145%** — agreed, leave it. A struct whose fields are the target table's columns is exactly the case where a line per field is the documentation, not padding. The `emit.go`/`sessions.go` seam is a real one too; the split reads better than a shave would have.

**Two things I am accepting rather than pushing on**, just so they are on the record:

- `evictBefore` still discards pending deltas silently — only the shutdown path flushes. With retention now at ~6 h 10 min the session is long finished by then and the loss is bounded by one heartbeat interval, so I would not add a sweep-time event. That is the sweep event we agreed against, and I still agree.
- The nil-snapshot case below is the one thing the `*Session` snapshot did not fully close.

One new inline comment. Everything else from the first pass is resolved as far as I can verify.

MiroCillik:
@pepamartinec @sykora-ji sry kluci, tohle se melo dnes nasadit a sbirat data a ne tady ladit minor issues a nits!

internal/pkg/service/appsproxy/dataapps/sessions/sessions.go
```diff
@@ -0,0 +1,360 @@
...
+				return
+			case <-ticker.Chan():
+				if n := m.store.evictBefore(m.clock.Now().Add(-m.retention())); n > 0 {
+					m.logger.Debugf(ctx, "evicted %d idle sessions, %d remaining", n, m.store.len())
+				}
+			}
+		}
+	}()
+
+	d.Process().OnShutdown(func(ctx context.Context) {
```
pepamartinec, line 116: **The shutdown budget overruns the pod's grace period, and the new flush is what gets sacrificed.** The kbc-stacks chart does not set `terminationGracePeriodSeconds`, so it is the Kubernetes default of 30 s.

Shutdown runs LIFO and sequentially:

| Stage | Budget | Source |
|---|---|---|
| 1. `srv.Shutdown` drains in-flight requests | **30 s** | `gracefulShutdownTimeout`, `server.go:34` |
| 2. `flushAll` | fast | this callback |
| 3. `writer.close` | **10 s** | `drainTimeout`, `writer.go:20` |
| **wanted** | **≈40 s** | |
| **available** | **30 s** | K8s default |

Stage 1's budget alone equals the whole grace period, and `Upstream.HTTPTimeout` is also 30 s (checked on `main` and here), so a request arriving just before SIGTERM can legitimately hold `srv.Shutdown` for its full 30 s. When that happens **the pod is SIGKILLed before stages 2 and 3 run at all**.

The sharp bit: `flushAll` exists to stop losing pending deltas on deploy, but it is last in the chain, so it is the first thing dropped — and precisely on the busy deploys where there is most to flush. `writer.go`'s own comment says the 10 s bound exists so a dead Stream is not "long enough for Kubernetes to SIGKILL the pod and turn a graceful shutdown into a hard one". The bound is right; it was just sized in isolation rather than against the pod's total budget.

To be fair on attribution: 30 s drain against a 30 s grace period was already borderline before this PR. This PR adds a second stage on top, which turns "exactly at the limit" into "over it".

Options:

1. Set `terminationGracePeriodSeconds` to ~60 s in the chart. Durable, but cross-repo, and this feature is meant to be enable-able without a code change.
2. Shrink the budgets so they sum under 30 s — e.g. server 20 s + drain 5 s. Self-contained here. Truncating the server drain at 20 s does cut a long in-flight request, but that request is being hard-killed at 30 s today, so it is strictly better.
3. One shared deadline, each stage taking from what remains. Cleanest, most code.

I would do 2 now and 1 as the durable fix, with a comment naming the constraint so whoever next tunes either number sees the coupling.

Cross-PR heads-up: there is a branch raising `Upstream.HTTPTimeout` to 10 min. If that lands, stage 1 will hit its 30 s bound far more often and this goes from a tail case to a routine one.

internal/pkg/service/appsproxy/dataapps/sessions/emit.go
pepamartinec, line 38: Minor and a one-line fix, but it fails silently, which is why it is worth bothering with.

`activity()` calls `getOrInit` but never sets `item.session`, and `getOrInit` creates entries as `&entry{lastSeen: now}` with a nil snapshot (`store.go:62`). `begin()` is what normally fills it in, and normally runs first — but an entry can be removed and then recreated by `activity()` alone, and `drain()` then skips it:

```go
if item.session == nil || (item.requests == 0 && item.wsFrames == 0) {
    return Event{}, false
}
```

Reproduced with a throwaway internal test rather than reasoned about:

```go
item, _ := st.getOrInit("sess-1", now)
item.session = &Session{ID: "sess-1"}          // begin() ran
_, found := st.take("sess-1")                  // WebsocketClosed, connection A
require.True(t, found)
item2, created := st.getOrInit("sess-1", now)  // connection B's next frame
require.True(t, created)
item2.wsFrames += 7                            // real user activity
assert.Nil(t, item2.session)
_, ok := (&Manager{}).drain(item2)
assert.False(t, ok)                            // 7 frames silently discarded
```

Passes. The realistic trigger is two overlapping websockets on one session id — a Streamlit reconnect where the new socket is up before the old one's close fires — then a deploy before the next ordinary HTTP request re-runs `begin()`. Narrow, and it only loses frame counts, never a session.

`activity()` already has `s` in hand, so `item.session = s` alongside the delta increment closes the whole class.

Separately, `drain()` folds two different conditions into one silent `false`: "no snapshot" is a bug, "no deltas" is normal. Now that the metrics exist, splitting them and counting the first would stop this hiding again.

internal/pkg/service/appsproxy/dataapps/sessions/store.go
```diff
@@ -0,0 +1,123 @@
...
+	for id, item := range s.items {
+		items = append(items, item)
+		delete(s.items, id)
+	}
+	return items
+}
+
+// evictBefore drops entries not seen since the given time and returns how many
+// were removed.
+func (s *store) evictBefore(cutoff time.Time) int {
```
pepamartinec, line 102: Scaling note, not a problem today. `evictBefore` holds the **write** lock while walking every entry, so it blocks every request for its whole duration. Measured:

| entries | full stall |
|---|---|
| 1 000 | 0.04 ms |
| 10 000 | 0.54 ms |
| 100 000 | **8.6 ms** |

Runs every `sweepInterval` (5 min). At 100k concurrent sessions that is an 8.6 ms stall on every in-flight request, once per sweep. Linear, so it only bites a stack with very many concurrent visitors.

Not asking for a change — the hot path is 41 ns and this is nowhere near mattering yet. But nothing in the code or docs says the sweep is O(n) under an exclusive lock, and the obvious future tuning knob (shortening `sweepInterval`) would make it worse rather than better, so it is worth a sentence. If it ever does matter: snapshot the keys under `RLock` and delete in batches, or shard the map.

internal/pkg/service/appsproxy/dataapps/frameworkpoll/frameworkpoll.go
```diff
@@ -0,0 +1,19 @@
+// Package frameworkpoll identifies data-app frontend background polls, as its
+// own package so that "the user did something" has exactly one definition:
+// auto-suspend and session tracking both need it, and their packages already
+// depend on each other in the other direction.
+package frameworkpoll
```
pepamartinec: Naming, plus the package's stated reason for existing does not hold.

The docblock says:

> …as its own package so that "the user did something" has exactly one definition: auto-suspend and session tracking both need it, and **their packages already depend on each other in the other direction**.

There is no mutual dependency. `upstream` imports `sessions` (`upstream.go:21`), but `sessions` imports nothing from `upstream` — I checked every import in `sessions.go`, `emit.go`, `cookie.go`, `store.go`, `writer.go`, `event.go` and `metrics.go`. The dependency is one-way, so there is no cycle to break: the helper could live in `sessions` and `upstream` would just call it.

On the name itself — `Is` on its own is a fine Go idiom (`errors.Is`, `os.IsNotExist`), so it is not wrong by itself. The problem is that the package name is carrying all the meaning and carrying it badly: "framework" names an abstraction that does not exist, since the contents are two `/_stcore/` paths, i.e. Streamlit and only Streamlit. Both call sites also use the predicate purely to *skip* something, so it always reads negated.

Options:

1. Move it into `sessions`. No new package, no cycle. Cost: an auto-suspend rule living in a package called `sessions` is a semantic oddity.
2. `dataapps/streamlit` + `IsBackgroundPoll(path)` — my preference. Honest about the scope, and `streamlit.IsBackgroundPoll(req.URL.Path)` reads as a sentence at both call sites. Generalise the day a second framework appears.
3. `dataapps/useractivity` + `IsBackgroundPoll(path)` if a second framework is expected soon.

Either way the function needs a real name — once the package is concrete, a bare `Is` says nothing. Note `upstream.go:383` also refers to `frameworkpoll.Is` in prose, so a rename touches that comment too.

pepamartinec:
@MiroCillik to, ze ty reportovany data nebyly konzistentni je minor issue?

pepamartinec (commented):
Re-reviewed at `80746d6`. Everything from the previous round is fixed, and I replicated each claim rather than taking the replies at face value — all seven commits do what they say.

| Claim | Verified how | Result |
|---|---|---|
| Flush runs before `writer.close`, LIFO holds | read the `OnShutdown` body; `servicectx.go` iterates callbacks in reverse | correct |
| Flush emits `heartbeat`, never an end | `drain` → `EventHeartbeat`; `SignOut` is the only `EventSessionEnd` call site in the package | correct |
| Stamped at `lastSeen`, not `now` | `buildEvent(item.session, item, EventHeartbeat, item.lastSeen)` | correct |
| `retention()` = `max(IdleTimeout, wsTimeout+grace)` | read the helper; the sweeper uses it | correct |
| Three metrics, as named | `metrics.go` — `sessions.events.sent` / `.dropped` / `.tracked` | correct |
| `EndReason` gone entirely | no occurrence left; `buildEvent` takes no reason | correct |
| `idleTimeoutSeconds` on every row | `emit.go:130`, the `idle_timeout_seconds` sink column, and the `COALESCE(...)` docs query | correct |
| Density 47% → 35% | re-measured independently: **280 comment / 795 code = 35%** | matches exactly |
| `sessions.go` split | 360 + `emit.go` 210, both well under 500 | correct |
| Salt guard | `NewManager` disables tracking and logs; the test fails without it | correct |

`go build` clean · `go test -race ./internal/pkg/service/appsproxy/...` **all 17 packages green** · `golangci-lint` 4 issues, all pre-existing in files this PR does not touch.

The new tests are the good kind — `WebsocketCloseIsAFlushNotAnEnd`, `AtMostOneSessionEndPerSession`, `RetentionCoversTheLongestCookie`, `ShutdownFlushesPendingActivity` and the two metrics tests each pin a specific behaviour we argued about, rather than restating the implementation. The PR description is current now too, including the identity-less-rows note in the Deployment Plan.

**Agreed and not pursuing:** `event.go` staying at 145% by ratio (a struct whose fields are the target table's columns is the case where a line per field *is* the documentation), and `evictBefore` discarding pending deltas silently (at ~6 h 10 min retention the session is long finished; that is the sweep-time event we decided against).

I also checked the cookie HMAC cost per request, lock contention in `activity()`, the goroutine count, and the blast radius of Stream being unreachable. All fine — happy to share the numbers if useful.

Five comments, most severe first.

internal/pkg/service/appsproxy/proxy/apphandler/authproxy/oauthproxy/config.go
```diff
@@ -70,6 +70,9 @@ func proxyConfig(
 	v.Session = options.SessionOptions{Type: options.CookieSessionStoreType}
 	v.EmailDomains = []string{"*"}
 	v.InjectRequestHeaders = []options.Header{
+		// New, and the only one session tracking records.
+		headerFromClaim("X-Kbc-User-Id", userIDClaim(proxyProvider.Type)),
```
pepamartinec, line 74: Tohle bych vypustil. Do appky autorizovany pres GH to zacne posilat `user` jako ID. Nekontroloval jsem, co presne v tom je, ale lepsi by bylo fixnout, aby GH byl plnohodnotny OIDC a poskytoval `sub`.

Proc bych to ted uplne vypustil:
* neni to potreba (nesouvisi s trackingem session)
* menit to pozdeji je breaking change

pepamartinec, line 74: Tohle bych vypustil. Do appky autorizovany pres GH to zacne posilat `user` jako ID. Nekontroloval jsem, co presne v tom je, ale lepsi by bylo fixnout, aby GH byl plnohodnotny OIDC a poskytoval `sub`.

Proc bych to ted uplne vypustil:
* neni to potreba (nesouvisi s trackingem session)
* menit to pozdeji je breaking change

pepamartinec (approved):
Dobry.. ted uz jsme nasel fakt je nity, ktery neni nutny resit. Funkcne 👍 

Jediny co, hned v zapeti by bylo dobry jeste zas dropnout `X-Kbc-User-Id` - tam se mi to moc nezda.

internal/pkg/service/appsproxy/dataapps/sessions/sessions.go
sykora-ji: Correct on both counts — the comment above the check was wrong (`WsTimeout` does carry `configKey:"wsTimeout"`, so it's operator-settable), and the check itself is a hard `Config.Validate()` error, not a `Warnf` — that part was already fixed by the time this was written. Not fixing the stale comment text right now; resolving as-is.

internal/pkg/service/appsproxy/proxy/apphandler/upstream/upstream.go
```diff
@@ -364,14 +380,15 @@ func (u *AppUpstream) trace() chain.Middleware {
 
 			// Trace connection events. Background polls emitted by data-app
 			// frontends independent of user interaction (see
-			// isFrameworkBackgroundPoll) are not considered activity and do
+			// frameworkpoll.Is) are not considered activity and do
 			// not bump lastRequestTimestamp.
```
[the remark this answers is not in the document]
sykora-ji: Checked: the parens actually match across the two comment lines ("(see" / "streamlit.IsBackgroundPoll)", after the rename) — reads correctly as one sentence. Resolving.

scripts/stream-sessions-setup.sh
```diff
@@ -0,0 +1,348 @@
...
+#   export KEBOOLA_TOKEN=<sapi-token-of-the-internal-project>
+#   bash scripts/stream-sessions-setup.sh
+#
+# Optional overrides:
+#   KEBOOLA_BRANCH_ID — branch id, or "default" (the default). Note this is not
+#                       the Storage API convention: Stream rejects "0".
+#   STREAM_API_HOST — defaults to stream.keboola.com
+#   SOURCE_NAME     — defaults to "Data App Sessions"
+#   TABLE_ID        — defaults to in.c-data-apps.sessions
+#   CLEANUP         — set to "true" to delete the source instead
```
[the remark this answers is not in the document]
sykora-ji: Correct — SINK_NAME, STATE_FILE and ALLOW_STATE_IN_REPO are all real overrides missing from that list. Not editing the docblock right now; resolving as a known gap.

PR #2657, opened 2026-09-14
feat(apps-proxy)!: stop leaking app name and raw user id in session events

sykora-ji:
[Linear: AJDA-3219](TODO: paste issue link)

## Description

Follow-up hardening of the Data App session tracking feature (AJDA-3219, already live on `kbc-eu-central-1` and `com-keboola-azure-north-europe`) that emits `session_start`/`heartbeat`/`session_end` events to a Keboola Stream HTTP source, landing in the `in.c-data-apps.sessions` Storage table.

Two fields in the emitted `Event` (`internal/pkg/service/appsproxy/dataapps/sessions/event.go`) carried more than they should:

- **`app_name`** — the data app's display name, redundant with `app_id` which already identifies it uniquely. It is now never populated. The `Event.AppName` field and the `app_name` Storage column are intentionally kept (not removed), so the sink mapping in `scripts/stream-sessions-setup.sh` and the table schema need no change — the column just reads empty going forward.
- **`provider_user_id`** — previously the raw OIDC subject claim, or GitHub account login, taken verbatim from the internal `X-Kbc-User-Id` header. It is now pseudonymized before ever being stored on a `Session` or sent anywhere:

  ```
  provider_user_id = hex(HMAC-SHA256(sessions.userIdHashKey, auth_provider_id + ":" + sub))
  ```

  The full HMAC digest is kept (not truncated like the session cookie signature in `cookie.go`'s `sign()`), since this is a one-way pseudonymization, not a tamper check. The key is a **new per-stack secret**, `sessions.userIdHashKey` (`config.go`), deliberately not a reuse of the existing `cookieSecretSalt` (used for signing the session cookie, a different purpose).

  If `sessions.streamUrl` is set but `sessions.userIdHashKey` is empty, the manager now disables tracking entirely (logged as an error, not fatal) — mirroring the existing guard for an empty `cookieSecretSalt`. The raw id must never leave the process unhashed, but a stack missing the new secret should not crash apps-proxy.

Docs (`docs/apps-proxy/sessions.md`) and tests updated accordingly, including a new case asserting `auth_provider_id` is actually mixed into the hash (so the same subject under two different providers cannot collide).

## Release Notes

- **Justification**
    - `app_name` had no identified downstream use beyond `app_id`, and the raw end-user subject claim / GitHub login being sent to Stream in the clear is more exposure than the feature needs — this closes both gaps before the feature reaches more stacks.
    - Non-technical: the anonymous usage-tracking events data apps already send now identify the app only by its ID (not its name), and identify the end user only by an irreversible hash instead of their actual account identifier.
    - No customer communication needed — this is an internal Keboola analytics pipeline (session tracking events), not customer-visible behavior.
- **Impact Analysis**
    - **Action required from SRE before/at deploy**: stacks that already have `sessions.streamUrl` configured — currently `kbc-eu-central-1` and `com-keboola-azure-north-europe` — need a new secret added to their apps-proxy secrets in kbc-stacks:
      - Key: `sessions.userIdHashKey`
      - Value: a random per-stack secret, e.g. generated with `openssl rand -hex 32`
      - Location: same encrypted `apps-proxy/secrets.yaml` as the existing `sessions.streamUrl`, under the `sessions:` block.
    - If this is deployed to those stacks **before** the secret is added, session tracking silently pauses (error-logged) until the secret is in place — apps-proxy itself keeps running fine, only this feature stops. No data is lost beyond not being tracked during the gap; no raw ids are ever sent unhashed.
    - Single-tenant impact: none beyond the two stacks above — stacks without `sessions.streamUrl` set are unaffected either way (tracking already off).
    - Not behind a separate feature flag; gated by the existing per-stack `sessions.streamUrl` config.
- **Deployment Plan**
    - Continuous deployment, same as the original session-tracking rollout (dev/testing wave, then production wave). **Coordinate with SRE to add `sessions.userIdHashKey` to `kbc-eu-central-1` and `com-keboola-azure-north-europe` apps-proxy secrets before or immediately alongside this deploy**, to avoid a tracking gap.
- **Rollback Plan**
    - Reverting/redeploying the previous version is sufficient; not a one-way door. The new secret can stay in kbc-stacks unused if rolled back.
    - Watch the `session tracking is disabled, user id hash key is empty` error log on the two affected stacks after deploy — its presence means the secret still needs adding. No other monitoring beyond the existing `keboola.go.appsproxy.sessions.*` metrics.


internal/pkg/service/appsproxy/config/config.go
```diff
@@ -46,6 +46,11 @@ type Sessions struct {
 	// e.g. https://stream-in.<suffix>/stream/<projectId>/<sourceId>/<secret>.
 	// Empty disables session tracking.
 	StreamURL string `configKey:"streamUrl" configUsage:"Full Stream HTTP source URL including the secret. Empty disables session tracking." validate:"omitempty,url" sensitive:"true"`
+	// UserIDHashKey is the per-stack HMAC-SHA256 key used to pseudonymize the
+	// end user id (auth_provider_id + ":" + the provider's subject claim/login)
+	// before it is sent to Stream. Tracking is disabled if StreamURL is set but
+	// this is empty, so the raw id is never sent unkeyed.
+	UserIDHashKey string `configKey:"userIdHashKey" configUsage:"Per-stack HMAC-SHA256 key used to pseudonymize the end user id before it is sent to Stream." validate:"omitempty" sensitive:"true"`
```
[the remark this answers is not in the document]
sykora-ji: Good catch, fixed in dcd80fdcb — added `APPS_PROXY_SESSIONS_USER_ID_HASH_KEY` to `TestSessionsConfig_Envs` and verified it actually fails when the `configKey` tag is wrong.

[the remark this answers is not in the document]
sykora-ji: Fixed in dcd80fdcb — added `APPS_PROXY_SESSIONS_USER_ID_HASH_KEY` to `TestSessionsConfig_Envs` and verified it fails when the `configKey` tag is broken.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

internal/pkg/service/appsproxy/config/config.go
```diff
@@ -46,6 +46,11 @@ type Sessions struct {
 	// e.g. https://stream-in.<suffix>/stream/<projectId>/<sourceId>/<secret>.
 	// Empty disables session tracking.
 	StreamURL string `configKey:"streamUrl" configUsage:"Full Stream HTTP source URL including the secret. Empty disables session tracking." validate:"omitempty,url" sensitive:"true"`
+	// UserIDHashKey is the per-stack HMAC-SHA256 key used to pseudonymize the
+	// end user id (auth_provider_id and the provider's subject claim/login)
+	// before it is sent to Stream. Tracking is disabled if StreamURL is set but
+	// this is empty, so the raw id is never sent unkeyed.
+	UserIDHashKey string `configKey:"userIdHashKey" configUsage:"Per-stack HMAC-SHA256 key used to pseudonymize the end user id before it is sent to Stream." validate:"omitempty" sensitive:"true"`
```
MiroCillik: **Medium — a weak or junk key passes validation silently.**

`validate:"omitempty"` is a no-op here: any non-empty string is accepted as the HMAC key, including a single character, and nothing logs a warning. The whole privacy guarantee of this PR rests on this value being unpredictable, yet only the empty string fails closed.

Suggest `validate:"omitempty,min=32"` (or a hex-format check) so a junk key fails the same way an empty one does. See also the paste hazard in `scripts/stream-sessions-setup.sh`.

scripts/stream-sessions-setup.sh
```diff
@@ -395,6 +399,16 @@ cat <<EOF
...
 
-  Leaving it unset keeps session tracking switched off, which is how stacks
-  without Stream stay unaffected.
+  Tracking additionally needs a hash key apps-proxy uses to pseudonymize the
+  end user id before it is sent to Stream — this script does not generate one
+  (re-running this script must not silently rotate an already-deployed key).
+  Generate it once and add it alongside the URL above, in the same encrypted
+  secrets:
+
+    APPS_PROXY_SESSIONS_USER_ID_HASH_KEY=\$(openssl rand -hex 32)
```
MiroCillik: **Medium — this line is a paste hazard that can silently produce a publicly-known key.**

The destination here is a SOPS-encrypted **YAML** file in `kbc-stacks`, not a shell — so `$(openssl rand -hex 32)` is not expanded. An operator who pastes the line verbatim gets the literal string `$(openssl rand -hex 32)` as the HMAC key: identical on every stack that does it (defeating "cannot be correlated across stacks") and publicly known (defeating pseudonymization entirely). Combined with the missing length validation in `config.go`, nothing in the config, logs, or metrics indicates anything is wrong.

Either print the generated value itself, or make it explicit that the operator must run the command and paste the *output*.

docs/apps-proxy/sessions.md
```diff
@@ -40,6 +40,15 @@ e-mail claim, and no display name is recorded. An OIDC issuer is free to use
 the e-mail address as its subject, so the value can still look like one — that
 is the issuer's choice, not something this asks for.
 
+That raw value never leaves the process. Before it is stored on the session or
```
MiroCillik, line 43: **Medium — rows already in `in.c-data-apps.sessions` are not addressed anywhere in this PR.**

Two concrete consequences:

1. The privacy fix is incomplete — every row written before this deploy still holds the raw subject claim / GitHub login and the app display name, so the data the commit message says is no longer leaked is still sitting in Storage.
2. `COUNT(DISTINCT provider_user_id)` over any window spanning the release counts each returning user twice, once as the raw id and once as the hash, with no way to tell from the row which era it came from.

The schema-evolution note further down covers *new columns going empty*, but not a column whose *values change meaning*. Either purge/rewrite the historical rows, or document the cut-over timestamp analysts have to split on.

internal/pkg/service/appsproxy/dataapps/sessions/cookie.go
```diff
@@ -78,6 +79,33 @@ func sign(payload string, key []byte) string {
 	return hex.EncodeToString(mac.Sum(nil)[:signatureLen])
 }
 
+// hashProviderUserID pseudonymizes the provider's user id before it is ever
+// stored on a Session or sent anywhere: HMAC-SHA256 of authProviderID and sub,
+// hex-encoded, with the full digest kept (unlike sign, this is not a tamper
+// check, so truncating it would only weaken it). Keyed per stack, so Stream
```
MiroCillik, line 85: **Low — "the hash cannot be reversed" overclaims** (the same wording is mirrored in `docs/apps-proxy/sessions.md`).

It cannot be reversed *without the key*, but the input space is small and enumerable — GitHub logins, corporate e-mail-shaped subject claims — so anyone holding the key recovers the original by dictionary in seconds. And the key lives in the same encrypted secrets bundle as the Stream URL, i.e. reachable by roughly the same people who can read the table.

"Pseudonymized, not anonymized" is the accurate claim. The current wording will lead someone to treat the column as safe to export or retain indefinitely.

internal/pkg/service/appsproxy/dataapps/sessions/manager_internal_test.go
```diff
@@ -56,6 +57,44 @@ func TestNewManager_EmptySaltDisablesTracking(t *testing.T) {
...
+func TestHashProviderUserID_NoBoundaryCollision(t *testing.T) {
+	t.Parallel()
+
+	// Neither input is guaranteed free of ':' — authProviderID is a
+	// config-time provider.ID, sub an OIDC-issuer-controlled subject claim —
+	// so a plain authProviderID + ":" + sub join would let these two distinct
+	// (provider, subject) pairs hash identically. Length-prefixing the first
+	// field must keep them apart.
+	key := []byte("key")
+	a := hashProviderUserID(key, "a", "b:c")
```
MiroCillik: **Low — this test does not test what its name says.**

`("a", "b:c")` and `("a:b", "c")` also differ under plain concatenation with no separator and no prefix (`"ab:c"` vs `"a:bc"`), so deleting the `binary.BigEndian.PutUint64` length prefix entirely leaves this test green — despite pinning that prefix being the entire reason commit `eae8b90` exists.

The pair that actually requires the length prefix is `("ab", "c")` vs `("a", "bc")` — both concatenate to `"abc"`. Same applies to the mirrored `hashUserID` helper in `manager_test.go`.

internal/pkg/service/appsproxy/dataapps/sessions/sessions.go
```diff
@@ -88,6 +89,15 @@ func NewManager(ctx context.Context, d dependencies) *Manager {
 		return m
 	}
 
+	// Without this key the end user id could only be sent raw or not at all.
+	// Tracking off beats tracking that leaks it.
+	if cfg.Sessions.UserIDHashKey == "" {
+		m.enabled = false
+		logger.Error(ctx, "session tracking is disabled, user id hash key is empty")
```
MiroCillik, line 113: **Low — fail-closed is right, but the failure is invisible outside pod logs.**

This is the first time the feature can go dark on a stack that was previously tracking. Deploying this image before the secret lands in `kbc-stacks` stops all session rows, and the only signal is this single `ERROR` at startup: the `unattributed` metric won't fire (nothing reaches the store path that increments it) and no other counter moves, so on any dashboard the failure looks identical to a quiet weekend.

Consider emitting a `sessions_enabled` gauge, or having the sessions dashboard alert on zero events, so the misconfiguration is visible without reading logs.

MiroCillik (commented):
Review of the session pseudonymization change.

The core change is sound: `hashProviderUserID`'s length-prefixed encoding is genuinely unambiguous, the guard in `NewManager` sits before `m.writer`/`m.metrics` are built (so no goroutine or writer leaks when tracking is refused), returning `""` for an empty `sub` preserves the `providerUserID != ""` identity-flush and `item.identitySent` semantics, and `Session.providerUserID`/`appName` have no readers outside the package (`FromContext` exposes only `ID`/`StartedAt`), so nothing downstream breaks. Sink mapping keeps `app_name` with `defaultValue: ""`, so the always-empty field is safe. No correctness bug found in the hash construction or the wiring.

Six comments below: two medium (key strength / operator paste hazard, and the untouched historical rows), four low.

internal/pkg/service/appsproxy/config/config.go
sykora-ji: Fixed in e6958d5 — added `validate:"omitempty,min=32"` plus a test (`TestSessionsConfig_RejectsShortUserIDHashKey`) asserting both the rejection and the boundary at exactly 32 chars. Also updated `configUsage` to state the minimum and a generation hint.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

scripts/stream-sessions-setup.sh
sykora-ji: Fixed in e6958d5 — the script now tells the operator to run `openssl rand -hex 32` themselves and paste the *output*, not a `KEY=$(command)` line, into the encrypted secrets. Docs updated with the same warning.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

docs/apps-proxy/sessions.md
sykora-ji: Documented in e6958d5 as a Known Limitation (§6): rows written before this ships keep the raw values, this diff neither purges nor backfills them, and analysts need the per-stack cut-over timestamp to split on for `COUNT(DISTINCT provider_user_id)`. I deliberately did not touch `in.c-data-apps.sessions` itself — purging or rewriting production Storage rows is a decision for the table's owner, not something this fix does automatically.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

internal/pkg/service/appsproxy/dataapps/sessions/cookie.go
sykora-ji: Fixed in e6958d5 — reworded to "pseudonymized, not anonymized" and spelled out why (small enumerable input space, key sits in the same secrets bundle as the table's ingest URL), instead of claiming irreversibility.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

internal/pkg/service/appsproxy/dataapps/sessions/manager_internal_test.go
sykora-ji: Confirmed and fixed in e6958d5 — you're right, `("a","b:c")` vs `("a:b","c")` stayed distinct even with the prefix removed entirely (`"ab:c"` vs `"a:bc"`). Replaced with `("ab","c")` vs `("a","bc")`, which do concatenate to the same `"abc"`; verified this version fails without the length prefix, where the old one didn't.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

internal/pkg/service/appsproxy/dataapps/sessions/sessions.go
sykora-ji: Fixed in e6958d5 — added a `keboola.go.appsproxy.sessions.enabled` gauge, registered unconditionally (unlike the rest of `*metrics`, which never exists at all when tracking is off) and read live off the manager, so it reflects whichever guard turned tracking off. Had to update `proxy/server_test.go`'s "no keboola.* metrics" assertion to expect it.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

internal/pkg/service/appsproxy/dataapps/sessions/sessions.go
```diff
@@ -74,6 +75,13 @@ func NewManager(ctx context.Context, d dependencies) *Manager {
 		store:     newStore(),
 	}
 
+	// Registered before any of the checks below can turn tracking off, and
+	// read through m rather than captured as a fixed value, so a stack that
+	// goes dark because a required secret is missing shows up as this gauge
+	// reading 0 rather than the rest of *metrics simply not existing — which
+	// on a dashboard looks identical to an ordinary quiet period.
+	registerEnabledGauge(d.Telemetry().Meter(), func() bool { return m.enabled })
```
MiroCillik: **Medium — this registration publishes `m` to the collector goroutine before `m.enabled` is done being written.**

The closure escapes to the meter's observable-gauge callback, which the SDK invokes from the metric reader's own goroutine. But `m.enabled` is then written at lines 94 and 102 (`m.enabled = false` on the salt and hash-key guards) with no synchronization between the two. That is an unsynchronized write/read pair across goroutines — a data race by the Go memory model, not merely a stale read.

The window is narrow (a collection would have to land between line 83 and the return), so `-race` in CI almost certainly won't flag it — which is what makes it worth catching by review rather than by test.

The comment above the call says the gauge is registered early deliberately, and the *intent* is right: the gauge has to exist on the paths that return early. But registration order doesn't need to precede the writes to achieve that — `defer` gives you both, since deferred calls run at return, after every `m.enabled = false`, and the meter's own internal locking then provides the happens-before edge:

```go
// Registered on every return path, including the ones below that turn
// tracking off, so a stack going dark shows up as this gauge reading 0
// rather than as the rest of *metrics simply not existing.
defer func() { registerEnabledGauge(d.Telemetry().Meter(), func() bool { return m.enabled }) }()
```

`sync/atomic.Bool` for the field would also work, but it would make every request-path read at lines 202/211 atomic for no reason — `enabled` is immutable once `NewManager` returns, and `defer` keeps it that way from the collector's point of view too.

MiroCillik (commented):
Re-reviewed `e6958d5`. All six are addressed, and addressed properly rather than papered over — in particular the boundary test now genuinely fails without the length prefix, the `min=32` comment is honest about what length validation can and cannot prove, and the Known Limitation spells out the double-counting rather than just noting the columns changed.

One new issue introduced by the `sessions_enabled` gauge — a data race. Details inline.

internal/pkg/service/appsproxy/dataapps/sessions/sessions.go
sykora-ji: Confirmed and fixed in bcae908 — applied exactly the suggested `defer` (registering the gauge on every return path, after whichever path's writes to `m.enabled` are done, so the collector goroutine can only observe the final value). `-race` still passes, as expected given the narrow window.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

MiroCillik: Verified `bcae908` — correct. The `defer` runs after every return path's writes to `m.enabled`, and since deferred calls complete before the caller resumes, the gauge is registered before the manager is reachable by anything else. Nothing else the closure reads is written afterwards, and `enabled` is never touched outside `NewManager`, so the collector goroutine can only observe the final value. Resolved from my side.

sykora-ji: Acknowledged, thanks for verifying. No further change needed here.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

MiroCillik (approved):

PR #2659, opened 2026-09-15
fix(apps-proxy): reject over-long Stream source and sink names up front

Halama:
Linear: [ST-4483](https://linear.app/keboola/issue/ST-4483/nastavit-session-tracking-pro-cs-a-slsp)

## Release Notes
`scripts/stream-sessions-setup.sh` now rejects a source or sink name longer than Stream's 40-character limit before it creates anything, instead of failing part-way through.

None.

## Impact analysis

Operator script only — no service code, nothing deployed, no runtime behaviour changes.

Hit while provisioning the remaining multi-tenant stacks by looping over the script with a `"<prefix> - <stack name>"` naming convention:

```
{"statusCode":400,"error":"stream.api.invalidLength",
 "message":"Length of body.name must be lesser or equal than 40 but got value
            \"Data App Sessions - com-keboola-azure-north-europe\" (len=50)"}
```

`Source.Name` and `Sink.Name` are both `validate:"required,min=1,max=40"` (`internal/pkg/service/stream/definition/{source,sink}_base.go`), but they are validated at different points in this script: the source name in step 1, the sink name only in step 3. An over-long **sink** name therefore fails *after* the source has been created and the state file already holds its ingest URL — leaving an orphan source, and a resume that walks straight back into the same error, since the script skips the steps it has already recorded.

That is the case this fixes. The over-long **source** name (what was actually hit) merely failed with a clumsy error; the sink case leaves real debris behind.

Both names are now checked before the first API call, so a rejected run creates nothing and writes no state file.

Refusing rather than truncating is deliberate: two stacks whose names differ only past the 40th character would collapse onto the same source, and because the script resolves a 409 by reusing the existing source by name, the second stack's run would silently adopt the first stack's source — sending two stacks into one table with no way to tell the rows apart.

Verified by running the script against a non-resolving host with a dummy token:

| Case | Result |
|---|---|
| `SOURCE_NAME` 50 chars (the real failure) | rejected before any API call |
| `SINK_NAME` 47 chars | rejected before any API call — previously created the source first |
| name exactly 40 chars | accepted, proceeds |
| name 41 chars | rejected |
| rejected runs | no state file written |

## Change type
Bug fix — operator tooling.

## Justification
Provisioning session tracking across the remaining stacks means running this script once per stack, where the limit is easy to cross. Failing fast with the limit and the actual length in the message costs one comparison; the alternative is an orphan source that has to be cleaned up by hand.

## Deployment
Merge & automatic deploy. The script is run manually by an operator, so nothing ships on merge.

## Rollback plan
Revert of this PR.

## Post release support plan
None.


Halama:
Closing — the fleet rollout this came out of is done, so the script is not expected to be run again soon and this is not worth carrying as an open PR.

For anyone who hits it later: Stream caps `Source.Name` and `Sink.Name` at 40 characters, and the script validates them at different points — the source name in step 1, the sink name only in step 3. An over-long sink name therefore fails after the source exists and the state file already holds its ingest URL, leaving an orphan source behind. Keep source and sink names at 40 characters or fewer.

The commit is preserved on `claude/stream-sessions-name-length` if it is ever wanted.

PR #2660, opened 2026-09-16
PAT-2113 Route to a Sandbox that owns its own hostname

pepamartinec:
## Release Notes
https://linear.app/keboola/issue/PAT-2113

A draft is a `Sandbox` CR that owns one hostname of its own, published by the operator at `.status.appsProxy.publicUrl` and carrying no app id. apps-proxy now watches Sandboxes, indexes them by that hostname, and checks the index before the existing `parseAppID` normalisation, which is byte-for-byte unchanged. A match supplies the app id from the Sandbox's own `spec.appId`, so the app config still loads.

Routing identity moves from `api.AppID` to `WorkloadRef{AppID, SandboxName}`, so the handler cache, upstream, E2B token, dev-mode check and wakeup follow the resolved workload. A Sandbox route starts no session and sends no `notify` — both are keyed by app id and would be attributed to production — and telemetry adds `proxy.sandbox.name` alongside the app id.

Key changes:
* An exact Sandbox match is a **priority, not a tie-break**. Nothing arbitrates: the platform allocates every published hostname and will not issue an App's to a Sandbox. Revisit before any workload with a free-form hostname.
* No domain validation on that path either — the index has one writer, fed only from `.status.appsProxy.publicUrl`, which only the operator can write.
* A Sandbox route is served, not redirected to the App's canonical host, and upstream redirects are rewritten to the Sandbox's own hostname.
* Caches keyed by workload are evicted when the workload is removed.

An arbitration step letting the App keep any hostname it answers on was built first and removed: it prevented no collision the platform can produce, and it refused every draft under a never-promoted App. Reasoning lives on `StateWatcher.ResolveWorkloadForHost`.

None.

## Impact analysis
No end-user impact for existing apps — every App route resolves as before. Drafts become reachable, except under OIDC, where login cannot complete until drafts get their own auth; it fails closed.

Needs the apps-proxy `Role` in `kbc-stacks` to grant `sandboxes`, or the informer is denied and drafts 404.

## Change type
New feature

## Justification
Drafts (PAT-2093) need a reachable URL of their own.

## Deployment
Merge & automatic deploy.

## Rollback plan
Revert of this PR.

## Post release support plan
None.


internal/pkg/service/appsproxy/dataapps/appconfig/middleware.go
```diff
@@ -16,41 +17,49 @@ type ctxKey string
...
-	AppID     api.AppID
+	Workload  k8sapp.WorkloadRef
 	AppConfig api.AppConfig
 	Err       error
 }
 
+// WorkloadResolver reports the workload that owns an exact hostname, if any.
+type WorkloadResolver interface {
+	ResolveWorkloadForHost(ctx context.Context, host string) (k8sapp.WorkloadRef, bool)
+}
```
Matovidlo, line 36: It seems weird that interface definition is in the middleware.

internal/pkg/service/appsproxy/dataapps/appconfig/middleware.go
```diff
@@ -78,6 +87,41 @@ func Middleware(configLoader Loader, host string) middleware.Middleware {
...
+// nothing arbitrates between a Sandbox and an App.
+func resolveWorkload(req *http.Request, resolver WorkloadResolver, host string) (k8sapp.WorkloadRef, bool) {
+	if ref, ok := resolver.ResolveWorkloadForHost(req.Context(), req.Host); ok {
+		return ref, true
+	}
+	if ref, ok := workloadForAppID(req, host); ok {
+		return ref, true
+	}
+	return k8sapp.WorkloadRef{}, false
+}
```
Matovidlo, line 113: This seems weird, why we return boolean ? Either the workload ref is empty of not which this function actually performs, therefore bool is redundant and cehck against empty structure can be done

internal/pkg/service/appsproxy/dataapps/appconfig/middleware.go
```diff
@@ -78,6 +87,41 @@ func Middleware(configLoader Loader, host string) middleware.Middleware {
...
+
+// workloadForAppID does not check that the App exists: an App is validated by
+// loading its config, not against the K8s cache.
+func workloadForAppID(req *http.Request, host string) (k8sapp.WorkloadRef, bool) {
```
Matovidlo, line 117: same here

internal/pkg/service/appsproxy/dataapps/k8sapp/appinfo.go
```diff
@@ -20,6 +27,20 @@ func AppGVR() schema.GroupVersionResource {
...
 }
 
+// SandboxGVR returns the GroupVersionResource for the Sandbox CRD.
+func SandboxGVR() schema.GroupVersionResource {
+	return schema.GroupVersionResource{Group: Group, Version: SandboxVersion, Resource: SandboxResource}
+}
+
+// NormalizeHost strips any port and lowercases the hostname, so index keys and
+// request hostnames are compared the same way.
+func NormalizeHost(host string) string {
```
Matovidlo, line 37: this is perferct utility function, either its already in utils or if not add it there

internal/pkg/service/appsproxy/dataapps/k8sapp/appinfo.go
```diff
@@ -35,7 +56,31 @@ const (
...
+	SandboxName string
+}
+
+// IsSandbox reports whether a Sandbox CR, rather than the App CR, owns the route.
+func (r WorkloadRef) IsSandbox() bool {
+	return r.SandboxName != ""
+}
+
+func (r WorkloadRef) String() string {
+	if r.SandboxName == "" {
```
Matovidlo, line 72: Either use `!IsSandbox()` or remove it

internal/pkg/service/appsproxy/dataapps/k8sapp/watcher.go
```diff
@@ -188,22 +221,139 @@ func (w *StateWatcher) WakeupApp(ctx context.Context, appID api.AppID) error {
...
+	parsed, ok := w.parseObject(ctx, "App", obj)
 	if !ok {
 		return
 	}
+
+	appID := api.AppID(parsed.appID)
+
+	w.routeLock.Lock()
+	w.apps.Store(appID, parsed.entry)
+	w.routeLock.Unlock()
```
Matovidlo, line 320: For me its confusing, we have sync Map and then bunch of regulars maps with mutexes. This map is locked during storing but `Store` is already sync so either
1. The apps map should not be sync.Map
2. The apps map should not be protected with mutex.
Make sure to have coherence in code

internal/pkg/service/appsproxy/dataapps/k8sapp/watcher.go
```diff
@@ -218,7 +368,16 @@ func (w *StateWatcher) handleUpsert(ctx context.Context, obj any) {
...
 		} else {
-			w.logger.Warnf(ctx, "App CRD %q (appID=%s) invalid upstream URL %q from appsProxy.upstreamUrl: %s", k8sName, appObj.Spec.AppID, rawURL, err)
+			w.logger.Warnf(ctx, "%s CRD %q (appID=%s) invalid upstream URL %q from appsProxy.upstreamUrl: %s", kind, k8sName, appObj.Spec.AppID, rawURL, err)
+		}
+	}
+
+	var host string
+	if rawURL := appObj.Status.AppsProxy.PublicURL; rawURL != "" {
+		if t, err := url.Parse(rawURL); err == nil {
+			host = NormalizeHost(t.Host)
```
Matovidlo, line 378: the NormalizedHost will be util

internal/pkg/service/appsproxy/dataapps/k8sapp/watcher.go
```diff
@@ -234,42 +393,108 @@ func (w *StateWatcher) handleUpsert(ctx context.Context, obj any) {
...
+		w.releaseHost(prev.host, e.k8sName)
+	}
+	if e.host != "" {
+		if owner, ok := w.sandboxHosts[e.host]; ok && owner != e.k8sName {
+			claimedFrom = owner
+		}
+		w.sandboxHosts[e.host] = e.k8sName
+	}
+	w.sandboxes[e.k8sName] = e
+	w.routeLock.Unlock()
```
Matovidlo, line 429: I would extract this to separate function so we can use defer Unlock call

internal/pkg/service/appsproxy/dataapps/k8sapp/watcher.go
```diff
@@ -234,42 +393,108 @@ func (w *StateWatcher) handleUpsert(ctx context.Context, obj any) {
...
+	}
+	k8sName := u.GetName()
+
+	w.routeLock.Lock()
+	e, found := w.sandboxes[k8sName]
+	if found {
+		w.releaseHost(e.host, k8sName)
+		delete(w.sandboxes, k8sName)
+	}
+	w.routeLock.Unlock()
```
Matovidlo, line 459: similar here

internal/pkg/service/appsproxy/dataapps/k8sapp/watcher.go
```diff
@@ -234,42 +393,108 @@ func (w *StateWatcher) handleUpsert(ctx context.Context, obj any) {
...
+	u, ok := objectFromDeleteEvent(obj)
+	if !ok {
+		return
+	}
+	k8sName := u.GetName()
+
+	w.routeLock.Lock()
+	e, found := w.sandboxes[k8sName]
+	if found {
+		w.releaseHost(e.host, k8sName)
```
Matovidlo, line 456: Are we sure we want to have recursion in here? Is the sandboxes map actually acyclic tree structure? So we do not have bidirectional references

internal/pkg/service/appsproxy/proxy/apphandler/manager.go
```diff
@@ -78,69 +80,105 @@ func NewManager(ctx context.Context, d dependencies) (*Manager, error) {
...
+		wrapper.cancel = nil
+	}
+	wrapper.handler = nil
 }
 
 func (m *Manager) HandlerFor(ctx context.Context, result appconfig.AppConfigResult) http.Handler {
-	wrapper := m.handlers.GetOrInit(result.AppID)
+	// The entry can be evicted between taking the pointer and taking its lock,
+	// in which case it is no longer the cache's and must not be built into.
+	for {
```
Matovidlo, line 127: Inifinite loop ? This looks not correct

Matovidlo (changes requested):

internal/pkg/service/appsproxy/dataapps/appconfig/middleware.go
pepamartinec: This is the Go convention: the interface is declared by the consumer, not next to the implementation. Defining it in `k8sapp` would make the watcher package depend on its caller, and the middleware would no longer state what it actually needs. `StateWatcher` satisfies it without importing anything from here.

internal/pkg/service/appsproxy/dataapps/appconfig/middleware.go
pepamartinec: Fixed in `aa2f9318c`.

internal/pkg/service/appsproxy/dataapps/appconfig/middleware.go
pepamartinec: Fixed in `aa2f9318c`.

internal/pkg/service/appsproxy/dataapps/k8sapp/appinfo.go
pepamartinec: Nope, this is no generic normalization. This is a normalization tailored for the needs of app routing.

You're right that there already is `NormalizeHost` in `utils` package, but that produces different format (trims trailing `/`, does not lowercase, does not strip port)

internal/pkg/service/appsproxy/dataapps/k8sapp/appinfo.go
pepamartinec: Fixed in `74aaa24a5`.

internal/pkg/service/appsproxy/dataapps/k8sapp/watcher.go
pepamartinec: Fixed in `8dc030b14` — `apps` is a plain map under the same lock as the Sandbox caches now, so there is one idiom.

I kept the lock rather than dropping it, because the lazy E2B token path reads an entry, fetches the secret, then writes it back, and `sync.Map` cannot make that atomic. The lookup sits in front of a proxied request, so an uncontended `RLock` is not on the critical path. Happy to split into two locks if you would rather `apps` did not share with the Sandbox maps — they no longer touch the same state.

internal/pkg/service/appsproxy/dataapps/k8sapp/watcher.go
pepamartinec: Same as the `NormalizeHost` thread above: it is routing-specific normalization, not a generic one, and it differs from the `utils` version.

internal/pkg/service/appsproxy/dataapps/k8sapp/watcher.go
pepamartinec: Fixed in `dd164dc29`.

internal/pkg/service/appsproxy/dataapps/k8sapp/watcher.go
pepamartinec: Fixed in `dd164dc29` — and `handleUpsert` had the same shape, so that one is extracted too. No manual `Unlock` calls remain.

internal/pkg/service/appsproxy/dataapps/k8sapp/watcher.go
pepamartinec: Could you elaborate on what you are seeing here? I do not think there is any recursion: `handleSandboxDelete` calls `releaseHost`, which deletes a single key from `sandboxHosts` and calls nothing further. `sandboxes` is a plain map from K8s object name to a cache entry — no tree, no parent or child references, nothing bidirectional.

`dd164dc29` extracts the locked part into its own function, which may make it easier to see. If you meant something else, point me at it.

internal/pkg/service/appsproxy/proxy/apphandler/manager.go
pepamartinec: Fixed in `262b88ca3` — the loop is gone.

It was not infinite: it only re-ran when the entry was evicted between taking the pointer and taking its lock, and the next `GetOrInit` returns a fresh entry. But retrying rebuilds a handler for a workload whose CR has just been deleted, so such a request is answered as not found instead.

internal/pkg/service/appsproxy/dataapps/k8sapp/watcher.go
```diff
@@ -35,14 +37,25 @@ type entry struct {
...
+	tokenLoadGroup      singleflight.Group  // coalesces concurrent lazy-load K8s API calls per secret
+
+	// routeLock guards the Sandbox cache and serialises writes to the App cache,
+	// which are read-modify-write on the lazy E2B token path.
+	routeLock    sync.RWMutex
+	sandboxes    map[string]entry  // Sandbox K8s object name → entry
+	sandboxHosts map[string]string // exact hostname → Sandbox K8s object name
+
+	removedLock sync.RWMutex
+	removed     []func(WorkloadRef)
```
Matovidlo: This is unconventional having for each state watcher object slice of function references. Why do we need to persist it  ?

internal/pkg/service/appsproxy/proxy/apphandler/manager.go
```diff
@@ -78,69 +80,102 @@ func NewManager(ctx context.Context, d dependencies) (*Manager, error) {
...
+		handlers: syncmap.New[k8sapp.WorkloadRef, appHandlerWrapper](func(k8sapp.WorkloadRef) *appHandlerWrapper {
 			return &appHandlerWrapper{lock: &sync.Mutex{}}
 		}),
 		clock:                d.Clock(),
 		storageTokenVerifier: verifier,
 		sessionsManager:      d.SessionsManager(),
-	}, nil
+	}
+
+	d.AppStateWatcher().OnWorkloadRemoved(m.evictWorkload)
```
Matovidlo: If I undertsand it correctly we have slice of onWorloadRemoved functions but we have only 1 function because we have it only once in constructor. Am I right ?

internal/pkg/service/appsproxy/dataapps/k8sapp/watcher.go
Matovidlo: Also what should I imagine udner StateWatched.removed(workload) ?  I removed from state watcher various worloads? I find the naming weird because we have no reference to WorkloadRef in any other struct members except of this one

internal/pkg/service/appsproxy/dataapps/k8sapp/appinfo.go
Matovidlo: did you check the utils?

internal/pkg/service/appsproxy/dataapps/k8sapp/watcher.go
```diff
@@ -188,22 +221,139 @@ func (w *StateWatcher) WakeupApp(ctx context.Context, appID api.AppID) error {
...
+
+// ResolveWorkloadForHost reports the workload that owns an exact hostname. A
+// Sandbox that publishes status.appsProxy.publicUrl owns it, and its own
+// spec.appId names the app config, so the hostname need not contain an app id.
+//
+// Nothing arbitrates between a Sandbox and an App: the platform allocates every
+// published hostname and does not issue one to a Sandbox that an App already
+// answers on. A workload kind with a free-form hostname breaks that assumption
+// and has to be reconsidered here before one exists.
+func (w *StateWatcher) ResolveWorkloadForHost(_ context.Context, host string) (WorkloadRef, bool) {
```
Matovidlo: The context is omited 🤔

internal/pkg/service/appsproxy/dataapps/k8sapp/appinfo.go
pepamartinec: ofc, did you? :slightly_smiling_face: 

https://github.com/keboola/keboola-as-code/blob/f43e7476d245cdec3442f4e129675d8f88abf045/internal/pkg/utils/strhelper/strhelper.go#L125

internal/pkg/service/appsproxy/dataapps/appconfig/middleware.go
Matovidlo: I agree that the convention is correctly here.
I think I do not understand the intent. We have interface, we have implementation of it it k8sapp and very likely in test so we can "pretend" that we have `ResolveWorkloadForHost` fake implementation that suits unit tests. We define it in middleware because we need to import it in the `middleware_test.go`.
I do not think that if we have another `WorkloadResolver` implementation that it would be located in middleware but in something like `workload_resolver.go` package. Can we add such package and under it implementation for k8sapp? Then it starts to make sense because we can create testing implementation or other implementation for workloadResolver this way we are ought to implement it here which is incorrect and you face refactor.

internal/pkg/service/appsproxy/dataapps/k8sapp/watcher.go
Matovidlo: oh sorry my bad, this is some helper function because the function definition is between 2 functions that both uses that and it looked like recursion

